PR #2, opened 2018-04-08
Updated YAML format to support different enumerated values for R/W

lochsh:
The previous YAML format didn't support separate enumerated values for read and write usage.  This is required e.g. here http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0553a/Cihcajhj.html

I updated the peripheral yaml files that used the old format and the README example.  There was also quite a funny typo in the README :hankey: :scream: :see_no_evil:


adamgreig:
Thanks, lgtm!

PR #5, opened 2018-04-23
Add cargo build support

tones111:
This change adds a cargo build script to allow dependent crates to specify stm32 as a Cargo.toml dependency.  Cargo can also be used to build within the repository with something like...
cargo build --features 'rt stm32f303'

It does depend on a svd2rust PR
https://github.com/japaric/svd2rust/pull/202

This support only allows one device feature to be specified (otherwise duplicate statics prevent compilation).  The build.rs script will panic if it detects too many devices.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,5 @@
 **/Cargo.lock
 .deps/
 */src/stm32*/
+src/lib.rs
 html/
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,0 +1,81 @@
+[package]
+name = "stm32"
+version = "0.6.0"
+authors = ["Adam Greig <adam@adamgreig.com>"]
+description = "Device support crates for STM32 devices"
+repository = "https://github.com/stm32-rs/stm32-rs"
+readme = "README.md"
+keywords = ["stm32", "svd2rust", "no_std", "embedded"]
+categories = ["embedded", "no-std"]
+license = "MIT/Apache-2.0"
+
+[lib]
+path = "src/lib.rs"
+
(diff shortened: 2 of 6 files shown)


tones111:
This PR has been updated to not generate internal modules since only one device feature is allowed at a time.  Clippy build.rs support has been added and assiciated fixes made.  The svd2rust dependency has now been internalized so the user no longer needs it in their path.

The only external dependency to build is the nightly version of rustfmt.

tones111:
With https://github.com/rust-embedded/svd2rust/pull/267 landing, I've brushed off some of the dust from this effort.  Hopefully this simplifies crate maintenance and makes downstream usage more consistent.

tones111:
Pinging @adamgreig for another look please.  Thanks.

tones111:
Pinging @adamgreig for feedback on this changeset please.

gkelly:
I think this is a useful change, especially for users at the bleeding edge who want to use devices not yet in releases (stm32f730 in my case) who then need to clone and follow build instructions and then reference the crate locally instead of just having `stm32f7 = {git ...` as a dependency.

adamgreig:
After a lot of thinking about this I've decided to not include it in stm32-rs. I don't think having end users generate the modified SVD, then generate the PAC, and then finally build it is the way to go. We've got the crates.io builds for most users, and the new [stm32-rs nightlies](https://github.com/stm32-rs/stm32-rs-nightlies) which should allow using the latest master as a git dependency, so I think most use cases are covered.

Thank you regardless for putting the PR together, and apologies that it's taken almost two years to decide!

PR #11, opened 2018-05-02
fix matching of LCKR fields in GPIO_v2

bergus:
`fnname` doesn't seem to support `([0-9]|1[0-5])` regex, so I couldn't think of any more elegant solution than copying the descriptions.


adamgreig:
Good spot, thank you!

Issue #14, opened 2018-05-16
Improve clarity


xd009642:
So I didn't twig that `stm32l4` and `stm32h7` were published from this repo from initially reading the `README.md`, potentially adding the `crates.io` badges for each device family to the `README.md` would make it easier to find the `crates.io` pages (since the search functionality is occasionally lacking).

If there's a consensus on this I was going to tackle some of the documentation and could slip the badges in with the PR :+1:

xd009642:
As the changes suggested are now merged in I'll close this issue.

PR #15, opened 2018-05-16
Added WWDG + readme changes

xd009642:
This PR is mainly to add the Window Watchdog peripheral to to project and document it, it's my first time contributing and I've just been going off the README and some of the other peripheral yaml files so there may be some mistakes, let me know and I'll sort them tomorrow!

Additionally added, crates.io badges (as per #14) and some new devices (unintentionally but they were part of the svd zip I downloaded and hopefully they're beneficial to the project).

### DIFF
--- a/peripherals/wwdg/wwdg.yaml
+++ b/peripherals/wwdg/wwdg.yaml
@@ -1,7 +1,7 @@
 # The most common features, found on Basic timers TIM6 and TIM7,
 # and found on all more advanced timers.
 
-"WWDG":
+WWDG:
   CR:
     _modify:
       WDGA: 
@@ -13,6 +13,13 @@
         bitOffset: 0
         bitWidth: 7
   CFR:
+    WDGTB:
+      Div1: [0, "Counter clock (PCLK1 div 4096) div 1"]
+      Div2: [1, "Counter clock (PCLK1 div 4096) div 2"]
+      Div4: [2, "Counter clock (PCLK1 div 4096) div 4"]
+      Div8: [3, "Counter clock (PCLK1 div 4096) div 8"]
+      _merge:
+        - "WDGTB*"
     _modify:
       EWI:
         description: "Used to enable early wakeup interrupt. This allows an interrupt to be serviced before device reset. Only cleared after a reset"
@@ -22,17 +29,6 @@
         description: "Window value to be compared to the downcounter."
         bitOffset: 0
         bitWidth: 7
-    _add:
-      WDGTB:
-        Div1: [00, "Counter clock (PCLK1 div 4096) div 1"]
-        Div2: [01, "Counter clock (PCLK1 div 4096) div 2"]
-        Div4: [10, "Counter clock (PCLK1 div 4096) div 4"]
-        Div8: [11, "Counter clock (PCLK1 div 4096) div 8"]
-      registers:
-        WDGTB0:
-          description: "First bit of the WDGTB register"
-        WDGTB1:
-          description: "Second bit of the WDGTB register"
   SR:
     EWIF:
       _read:
--- a/peripherals/wwdg/wwdg.yaml
+++ b/peripherals/wwdg/wwdg.yaml
@@ -5,36 +5,32 @@
   CR:
     _modify:
       WDGA: 
-        description: Activation bit for the watchdog. This bit is set by software and only cleared after a reset.
+        description: "Activation bit for the watchdog. This bit is set by software and only cleared after a reset."
         bitOffset: 7
         bitWidth: 1
-        access: read-set
       T:
-        description: These bits contain the value of the watchdog counter. A reset is produced when it rolls over from 0x40 to 0x3F (T6 us cleared).
+        description: "These bits contain the value of the watchdog counter. A reset is produced when it rolls over from 0x40 to 0x3F (T6 us cleared)."
         bitOffset: 0
         bitWidth: 7
-        access: read-write
   CFR:
     _modify:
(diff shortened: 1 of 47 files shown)


adamgreig:
Thank you for the PR!

Does this build successfully (running `make`) for you locally? The Travis CI failing probably indicates an error in the new WWDG peripheral (though the error message is unhelpful). I'll have some time to check into it more carefully tomorrow. Also, I don't see any new devices added  in these commits, maybe there's a missing file?

xd009642:
Oh sorry not new devices edited ones, my mistake. And I've started going over it and found a number of mistakes which I'm now addressing. Hopefully version 2 will be good :+1:

xd009642:
Part of it which I'm trying to crack is there's a 2 bit wide variable in the register and it shows up in the SVD as two fields `WDGTB0` and `WDGBT1` when it should be a single field. If you have any guidance on that it would be helpful

xd009642:
Hi @adamgreig could you have a look at my latest version of the yaml? I've made some progress but have gotten stuck. But it's a lot better than it was!

peripherals/wwdg/wwdg.yaml
```diff
@@ -0,0 +1,38 @@
...
+  CR:
+    _modify:
+      WDGA: 
+        description: "Activation bit for the watchdog. This bit is set by software and only cleared after a reset."
+        bitOffset: 7
+        bitWidth: 1
+      T:
+        description: "These bits contain the value of the watchdog counter. A reset is produced when it rolls over from 0x40 to 0x3F (T6 us cleared)."
+        bitOffset: 0
+        bitWidth: 7
```
adamgreig: It seems like the WDGA and T fields are already present in the SVD, so they don't need modifying (the SVD already has the description, bitOffset, and bitWidth). Or is it broken/missing in some SVD?

The thing to do instead is give either enumerated values (perhaps Disabled and Enabled for WDGA), or a range of permissible values (such as [0, 127] for T). For example:

```
CR:
  WDGA:
    Disabled: [0, "Watchdog disabled"]
    Enabled: [1, "Watchdog enabled"]
  T: [0, 127]
```

peripherals/wwdg/wwdg.yaml
```diff
@@ -0,0 +1,38 @@
...
+  CFR:
+    WDGTB:
+      Div1: [0, "Counter clock (PCLK1 div 4096) div 1"]
+      Div2: [1, "Counter clock (PCLK1 div 4096) div 2"]
+      Div4: [2, "Counter clock (PCLK1 div 4096) div 4"]
+      Div8: [3, "Counter clock (PCLK1 div 4096) div 8"]
+      _merge:
+        - "WDGTB*"
```
adamgreig: Merging is a good idea but the `_merge` key needs to belong to the `CFR`, one level up, not be part of the `WDGTB`. This should work ok if you just unindent it 2 lines (and perhaps put it before the WDGTB block for readability).

peripherals/wwdg/wwdg.yaml
```diff
@@ -0,0 +1,38 @@
...
+    _modify:
+      EWI:
+        description: "Used to enable early wakeup interrupt. This allows an interrupt to be serviced before device reset. Only cleared after a reset"
+        bitOffset: 9
+        bitWidth: 1
+      W:
+        description: "Window value to be compared to the downcounter."
+        bitOffset: 0
+        bitWidth: 7
```
adamgreig: As with the WDGA and T fields, I think EWI and W should already exist in the SVD files, so you just need to give them variants and/or ranges here. Since EWI is just "set" or "reset" it doesn't really need enumerated values, but W would benefit from being told it can take any value [0, 127] (so that the setter functions become safe).

peripherals/wwdg/wwdg.yaml
```diff
@@ -0,0 +1,38 @@
...
+  SR:
+    EWIF:
```
adamgreig: Annoyingly this field is called `EWI` on some of the devices you've added the WWDG peripheral to (I think in all the STM32F1 family), which is why the automatic build is failing. The best thing is probably to add a fix to the F1 device files, like:
```
WWDG:
  SR:
    _modify:
      EWI:
        name: EWIF
```
You could add a note that the field in the SVD is named incorrectly compared to the reference manual (RM0008).

xd009642:
Ah thanks, it's late here but I'll try and get through all these changes tomorrow and let you know if I have any questions!

adamgreig:
Thanks, this looks good now! The final thing is you can actually add the WWDG to almost _all_ the devices. There's a script to make this easier, try:
```
python3 scripts/matchperipherals.py peripherals/wwdg devices/*.yaml
```
to see what it can apply against cleanly, and then run the same script with --update to automatically update the device yaml files. It looks like everything except the STM32H family can take this peripheral as-is!

xd009642:
I just added the peripheral to all the devices after you approved the changes, and it looks to have failed. I'll investigate and hopefully fix it but I may not have time tonight

adamgreig:
I think the problem is what I mentioned in the last comment -- the STM32H family apparently has a different WWDG so you can't add it to that. Did you use the script I mentioned? It doesn't match against the STM32H.

xd009642:
Ah I didn't see that comment. My internets being a bit dodgy so I put a comment and it doesn't get posted till a minute or so later and new ones don't pop up until the connection comes back! I'll revert the change to the H7 and push

adamgreig:
All good, thanks!

PR #17, opened 2018-05-23
Missing patch for `stm32f446`

gbip:
Hi,
It seems that there is a missing patch for the microcontroller `stm32f446`.
Please tell me if there is any problem with this change ! :smile:


gbip:
The part that is interresting on [this](https://github.com/adamgreig/stm32-rs/blob/master/peripherals/rcc/rcc_v2.yaml#L17) patch is this one :
```yaml
CFGR:
_merge: ["SW[01]", "SWS[01]"]
```
It seems that the build is broken and I don't really now what to do...

gbip:
Ok, then I'll try to implement that !

PR #20, opened 2018-05-27
WIP: Flexible memory controller

lochsh:
I still need to go through the different STMs and find what's common and what's not etc.  But this is a good start :)


adamgreig:
Rebased and merged by https://github.com/stm32-rs/stm32-rs/commit/884edafcea7f42f6d9da63d3ad8a43e74cd8ccb2

PR #1, opened 2018-04-05
Update SVDs

IngoMaindorfer:
Integrated new SVD files from STM.
Use unzip's feature to convert to lower case because not all "rename" commands support regex.
Add prerequirement section in readme.md showing how to get needed rustfmt

PR #28, opened 2018-06-18
Add stm32l4x5 yaml file

c-nixon:
I noticed that there was an unused svd file for the stm32l4x5 devices. Was this device family deliberately left out? I have access to one and would like to try getting it running with rust.

If it was just an oversight this PR adds the missing .yaml file.

If the stm32l4x5s were deliberately left out then feel free to close this PR and let me know if there's anything I help with.


adamgreig:
Thanks! I think this SVD got added by #1 but no corresponding yaml. All looks good to me.

Issue #19, opened 2018-05-24
Use form to split device files into mutiple folder.


gbip:
I think it is a good idea to use form to split devices files (`stm32f429.rs`) into a folder with subfolder for each module.
I can implement it if you want to :smile:

gbip:
Hey @adamgreig can you take a look at this :smile:
I am willing to implement it if you think that it can make documentation generation easier !

adamgreig:
If you'd like to have a go I'm happy to review, I'm up for at least seeing what it looks like and happy to merge if it seems reasonable. If it makes documentation easier then even better!

I imagine we'd have a folder for each device, so like `stm32f4/src/stm32f405/` which contains the generated files, and then feature-gated modules in `lib.rs` as currently?

adamgreig:
The support for svd2rust 0.13.1 means each device already gets its own folder, so hopefully running Form over it should be even easier now. I'll have a play with it this week.

gbip:
I'll try to see how it plays out this week too.

gbip:
So since I am not familiar at all with Makefiles, I didn't do anything except running form on a few `mod.rs` with success (by hand).

adamgreig:
Closed by #29

Issue #25, opened 2018-06-11
new sed script issues


baldwin1974:
Hi there again, I think for mac os the sed command `sed -i ""` is the necessary format. In your newest Makefile from last night, the command `sed -i''` (in single quotes) generates errors. Sorry about giving bad information earlier. I do think using `-i.tmp` is better but it does generate more bloat, and using double-quotes works; there must be a space between -i and the double-quotes, too.

adamgreig:
I don't think `-i ''` can work, at least with GNU sed, as it interprets the '' as an empty command string and the actual command string as an invalid filename. Perhaps `--in-place=""` will work on both?

baldwin1974:
I just tried `--in-place=""` on my mac and it didn't work. Are you saying neither double-quotes nor single quotes will work if there is a space between them and the `-i`? If so, then the only solution I can think of for both linux and mac is to run `-i.tmp` and then delete the .tmp file, or you could just put a bit of guidance in the Makefile ...

adamgreig:
I've updated the Makefile to not use `-i` at all; could you check if this works for you please?

Issue #3, opened 2018-04-12
docs.rs only shows empty crates


jonas-schievink:
For example: https://docs.rs/stm32f4/0.1.0/stm32f4/

The features added in https://github.com/onur/docs.rs/pull/73 can be used to make docs.rs build the crates with all features enabled:

```toml
[package.metadata.docs.rs]
all-features = true
```

jonas-schievink:
Actually, scratch that. `cargo doc --open --all-features` takes 10 **minutes** and uses up to 12 GB of memory, so I doubt the docs.rs builder would like that. I'd still like to find *some* solution for hosted docs, though.

adamgreig:
I was trying to look for that feature to make docs.rs build all features and couldn't find it anywhere, so thanks! But I fear you're right and it won't enjoy the intense RAM usage.

At the moment I have a sort of halfway towards being docs thing at https://stm32.agg.io/rs but it's not really what you want. I think I might be able to build them locally and host somewhere, but it would be nice if they could be on docs.rs. Could always add a link in the crate READMEs though.

adamgreig:
I uploaded docs for just stm32f405 [here](https://stm32.agg.io/rs/stm32f405/stm32f4/stm32f405/index.html) to try it out, but it takes a lot of time to generate and upload and about 500MB per device. I think I can probably wrangle that for everything but it'l take a while and is a bit extreme. I'll look in to generating them on the server at least.

gbip:
Maybe you could (ab)use Travis CI to generate the doc ?

jonas-schievink:
Travis only provides like 2 GB of RAM

gbip:
Generating the doc for 1 micro-controller took roughly 1.5 Gb on my computer, so I guess it is still doable, don't you think ?

phil-opp:
Maybe a dumb question, but: Have you actually tried to generate the docs on docs.rs? Or do you just think that the docs.rs builder is not capable enough? Because normal svd2rust crates like [stm32f7x6](https://crates.io/crates/stm32f7x6) seem to [work fine on docs.rs](https://docs.rs/stm32f7x6/0.2.0/stm32f7x6/). (I know that this crate is different because it contains code for multiple different devices, but I think it's worth a try.)

adamgreig:
I haven't actually tried. We'd just need to set
```[package.metadata.docs.rs]
all-features = true```
[...]

PR #29, opened 2018-06-18
[#19] Use form to split device files into multiple folders

sjroe:
Implements #19

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,8 +31,9 @@ $(1)/src/%/mod.rs: svd/%.svd.patched
 	mkdir -p $$(@D)
 	cd $$(@D); svd2rust -i ../../../$$<
 	export DEVICE=$$$$(basename $$< .svd.patched); \
-        sed "s/crate :: Interrupt/crate :: $$$${DEVICE} :: Interrupt/" $$(@D)/lib.rs > $$(@D)/mod.rs
-	form -i $$(@D)/mod.rs -o $$(@D)/
+        sed -i "s/crate :: Interrupt/crate :: $$$${DEVICE} :: Interrupt/" $$(@D)/lib.rs
+	form -i $$(@D)/lib.rs -o $$(@D)/
+	mv $$(@D)/lib.rs $$(@D)/mod.rs
 	rm $$(@D)/build.rs
 	find $$(@D) -name *.rs -exec rustfmt {} +
 endef
--- a/Makefile
+++ b/Makefile
@@ -30,13 +30,11 @@ define crate_template
 $(1)/src/%/mod.rs: svd/%.svd.patched
 	mkdir -p $$(@D)
 	cd $$(@D); svd2rust -i ../../../$$<
-	rustfmt $$(@D)/lib.rs
-	export DEVICE=$$$$(basename $$< .svd.patched); \
-        sed "1,6d;10d;s/crate::Interrupt/crate::$$$${DEVICE}::Interrupt/" $$(@D)/lib.rs > $$(@D)/mod.rs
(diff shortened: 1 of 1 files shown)


adamgreig:
Thanks for implementing this! I'll try and check it over this evening but two quick notes first:

* Do you delete the various extern crate and config lines that the first part of the sed command used to delete, or do they stay in the new mod.rs? (We delete lines 1-6 and 10 of the svd2rust lib.rs).
* We can't use `sed -i` (see #25) so prefer just running sed and outputting to a new file and deleting the old one... (yes this is pretty annoying, blame OS X!)

sjroe:
Apologies, I mistakenly removed the line removal when I was experimenting with not running `rustfmt` before making running the `sed` line.  This has been reworked.

I wasn't aware of OS X's version of `sed` didn't support -i but should have assumed that you weren't using it for a good reason.

adamgreig:
I've tested this out and it seems to work fine (thanks for fixing it up!) but I'm not sure whether to merge. Using form increases generation time about 4x on my desktop (12s -> 39s), without changing crate compilation or documentation time noticeably. Currently the auto-generated files are not uploaded to GitHub (since there are many hundreds of megabytes of them in total), so the nicer underlying files would only be present on the crates.io archive. I don't think it helps us generate the crate documentation (#3), since they still require shocking quantities of RAM and disk.

Basically I'm not sure what the advantage would be to using Form in this case, but I'm still open to merging it if there are good reasons. Maybe you or @gbip would have some ideas?

phil-opp:
Another advantage of using form is that you can look at the source more easily (without the need to open a file with 400k+ lines). This is especially important when using the `src` link in rustdoc, since loading this amount of lines in the browser stresses my laptop considerably. Also, it seems like rustdoc/firefox is not able not handle anchors to lines >50k (jumps to and highlights a wrong line).

adamgreig:
That's a good point. I don't remember the memory usage being any lower but I'll check.

If rustdoc can't build them anyway then perhaps an option is to just have using Form be an option when making which is disabled by default -- when building locally (whether to generate your own documentation or just for browsing the source) you could have it enabled, but otherwise the default (and quicker) option is to skip that step.

adamgreig:
Memory usage isn't any better sadly. The STM32F4 crate took over 12G to build all features in 9 minutes on my desktop, but I think something might have got oom killed as it finished pretty much right as it maxed out my RAM. Even viewing the generated documentation is a bit intense.

phil-opp:
> Memory usage isn't any better sadly.

That's a pity, but thanks for checking!

> If docs.rs can't build them anyway then perhaps an option is to just have using Form be an option when making which is disabled by default -- when building locally (whether to generate your own documentation or just for browsing the source) you could have it enabled, but otherwise the default (and quicker) option is to skip that step.

The version on crates.io is pre-generated, right? So we can't just enable a cargo feature in the crate, but would need to clone and re-generate the project. Wouldn't it make more sense to just use form by default and accept the longer generation times, because generation is only done seldomly (on new releases) while `cargo doc` is used by most users (especially if there is no docs.rs documentation)?

adamgreig:
Good point - so maybe the versions uploaded to Cargo are built using Form, but have an option to not use it when building locally for test/development where it meaningfully slows down that cycle.

phil-opp:
Sounds good to me!

adamgreig:
I modified this to have an optional target 'form' in the Makefile which can be called to run Form over all the generated sources, but isn't run by default. I can use it before building releases, but otherwise it's very slow and the huge number of files slows down git operations etc a lot too.

PR #30, opened 2018-06-19
Adding PLLR for stm32f76x and stm32f77x mcu.

mathk:


adamgreig:
Thanks! Should this be added to `stm32f7x9` as well? It doesn't seem to be in the SVD for it.

adamgreig:
I checked and I don't think the F7x6 should have PLLR, but F7x5 and F7x9 should do. It's a confusing naming scheme...

| Family | PLLR? | Ref Man |
|----------|------------|-----------|
| F7x2 | No | RM0431 |
| F7x3 | No  | RM0431 |
| F7x5 | Yes | RM0410 |
| F7x6 | No | RM0385 |
| F7x7 | Yes | RM0410 |
| F7x9 | Yes | RM0410 |

mathk:
Oh yes let me update that. BTW where did you check it? Did you looks on all the datasheet one by one ?

adamgreig:
I checked the reference manuals - there are only three (RM0431, RM0410, RM0385) but you have to check the page of at least one microcontroller in each family (e.g. [STM32F779NI](http://www.st.com/en/microcontrollers/stm32f779ni.html)) to see which reference manual goes with it, linked in the documents table at the bottom (or just remember them, which is mostly what I do...).

I might try and make a table describing which chips are in which family and which SVD and link it to the reference manuals, it'd make checking SVDs a lot easier for everyone.

PR #31, opened 2018-06-21
Add VOS enum.

mathk:
Maybe this can be more generic as other MCU might got this option too.

### DIFF
--- a/devices/stm32f0x0.yaml
+++ b/devices/stm32f0x0.yaml
@@ -4,3 +4,4 @@ _include:
  - ../peripherals/gpio/gpio_v2.yaml
  - ../peripherals/crc/crc_basic.yaml
  - ../peripherals/wwdg/wwdg.yaml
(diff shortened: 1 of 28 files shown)


mathk:
Don't forgot to squash before merging

adamgreig:
Thanks for starting the PWR peripheral. Looking at how it's done on various STM32s, I think it might be best to split it into pwr_v1 and pwr_v2, with v1 having CR and CSR, and v2 having CR1, CR2, CSR1, CSR2. That way it will be easier to add the rest of the PWR fields which are surprisingly common between devices ([check out the table](https://agg.io/u/stm32/stm32f/PWR_0x40007000.html)).

It looks like L0 and L1 would use v1, L4 would use v2; F0, F1, F2, F3, F4 would use v1, F7 would use v2. We could then add the PWR peripheral to most devices.

PR #33, opened 2018-06-26
Fix SAI2 interrupt.

mathk:


adamgreig:
Oops, good spot, thanks. Please could you update the description as well as the name?

Issue #41, opened 2018-09-06
STM32H7x3: Ethernet_MAC peripheral broken


ehntoo:
From what I can tell by the reference manual, the STM32H7x3 uses the same ethernet MAC as the STM32F7x9 and company. The SVD is seriously broken in this regard - it lists the DMA-related registers under "Ethernet_MAC" rather than under an "Ethernet_DMA" peripheral, and all of the other registers are missing entirely.

Is the preferred method for fixing an error this egregious to use the normal yaml patching mechanisms? Just seeking input before I put together a PR. :)

*edit*
Upon further inspection, the peripheral is _not_ the same as the F7x9. The SVD issues still stand, though.

adamgreig:
Yikes, that's shocking.

Yes, the normal YAML patching mechanisms would suffice to fix this - you can add entire new peripherals with new registers with new fields, and probably just wholesale delete the existing Ethernet_MAC since it's full of lies. For an example of doing this, check out [stm32f7x9](https://github.com/adamgreig/stm32-rs/blob/master/devices/stm32f7x9.yaml#L3) where we have to add an ADC_Common peripheral. A PR to fix it would be very appreciated!

This sort of thing you might hope ST themselves would fix with an updated SVD file but I wouldn't hold your breath.

ehntoo:
Fixed by #42 :)

PR #42, opened 2018-09-07
Add missing stm32h7 ethernet

ehntoo:
Addresses Issue #41 

This PR pulls in all the missing ethernet registers from the stm32h7 reference manual (RM0433 https://www.st.com/resource/en/reference_manual/dm00314099.pdf)

Some things to note - this ethernet peripheral's registers are grouped significantly differently from others in the STM32 family. I've reproduced the DMA, MAC, MTL register groups from the manual, but added a separate MMC group, as previous family members seem to have had.

There are also a number of ethernet statistic registers such as "TX_PACKET_COUNT_GOOD" which do not share the "MAC" register prefix - I have left them in the "Ethernet_MAC" register group.

I also noticed an aberration in addressing for the MACARPAR register - it looks from the register map as if it should be at offset 0xB10, but the manual is consistent in placing it at 0xAE0. ST's own HAL code puts it at 0xB10 (https://raw.githubusercontent.com/fhsakaci/STM32H743-LwIP/master/STM32Cube_FW_H7_V1.3.0/Drivers/CMSIS/Device/ST/STM32H7xx/Include/stm32h743xx.h), so I've chalked that up as an error in the reference manual.

Let me know if there are any changes I should make or if you'd like the commits squashed. :)

### DIFF
--- a/devices/stm32h7x3.yaml
+++ b/devices/stm32h7x3.yaml
@@ -4,6 +4,1762 @@ _svd: ../svd/stm32h7x3.svd
 _modify:
   Flash:
     name: FLASH
+  Ethernet_MAC:
+    name: Ethernet_DMA
+    baseAddress: 0x40029000
+    description: "Ethernet: DMA controller operation"
+
+_add:
+  Ethernet_MAC:
+    baseAddress: 0x40028000
+    addressBlock:
+      offset: 0x0
+      size: 0x400
+    registers:
+      MACCR:
+        description: Ethernet MAC Operating mode configuration register
+        addressOffset: 0x0
+        access: read-write
+        resetValue: 0x00000000
+        fields:
+          RE:
+            description: Receiver Enable
+            bitOffset: 0
+            bitWidth: 1
+          TE:
+            description: Transmitter Enable
+            bitOffset: 1
+            bitWidth: 1
+          PRELEN:
+            description: Preamble Length for Transmit Packets
+            bitOffset: 2
+            bitWidth: 2
+          DC:
+            description: Deferral Check
+            bitOffset: 4
+            bitWidth: 1
+          BL:
+            description: Back-Off Limit
+            bitOffset: 5
+            bitWidth: 2
+          DR:
+            description: Disable Retry
+            bitOffset: 8
+            bitWidth: 1
+          DCRS:
+            description: Disable Carrier Sense During Transmission
+            bitOffset: 9
+            bitWidth: 1
+          DO:
+            description: Disable Receive Own
+            bitOffset: 10
+            bitWidth: 1
+          ECRSFD:
+            description: Enable Carrier Sense Before Transmission in Full-Duplex Mode
+            bitOffset: 11
+            bitWidth: 1
+          LM:
+            description: Loopback Mode
+            bitOffset: 12
+            bitWidth: 1
+          DM:
+            description: Duplex Mode
+            bitOffset: 13
+            bitWidth: 1
+          FES:
+            description: MAC Speed
+            bitOffset: 14
+            bitWidth: 1
+          JE:
+            description: Jumbo Packet Enable
+            bitOffset: 16
+            bitWidth: 1
+          JD:
+            description: Jabber Disable
+            bitOffset: 17
+            bitWidth: 1
+          WD:
+            description: Watchdog Disable
+            bitOffset: 19
+            bitWidth: 1
+          ACS:
+            description: Automatic Pad or CRC Stripping
+            bitOffset: 20
+            bitWidth: 1
+          CST:
+            description: CRC stripping for Type packets
+            bitOffset: 21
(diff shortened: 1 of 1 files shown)


adamgreig:
This looks great, thanks! Did you automatically generate it?

I'm going to check it against the ref man this weekend. It might be worth moving the ethernet stuff into its own file, both since there's a lot of it and since the upcoming STM32H value-line devices might have the same problem. You can see how a few parts (e.g. the stm32f7 already) use the `common_patches` folder and just include YAML from there.

ehntoo:
It's semi-generated using copy-and-paste and a whole lot of Vim macro abuse. There's a little room for error, but things should be _fairly_ accurate. :) Were I to do this again, I think I'd spend some more time and put together a proper script to scrape things out of the reference manual in a more automated fashion.

I'll pull things into a separate patch file tomorrow, it seems like a good way to approach this.

ehntoo:
Well, "tomorrow" turned into 2 and a half weeks, but I finally got back to this. I extracted the modifications into a separate patch file, which makes the device file _much_ nicer to read again.

I did [report this](https://community.st.com/s/question/0D50X00009hnZLRSA2/stm32h7-svd-bugs) on the ST community forums, and it's now been "reported internally", but we'll see if/when anything comes of that.

devices/common_patches/h7_ethernet.yaml
```diff
@@ -0,0 +1,2082 @@
...
+            bitWidth: 1
+          TLPIEN:
+            description: Transmit LPI Entry
+            bitOffset: 0 
+            bitWidth: 1
+      MACLTCR:
+        description: Ethernet MAC LPI timers control register
+        addressOffset: 0xD4
+        access: read-write
+        resetValue: 0x00000000
```
adamgreig: This one is `0x03E80000`

devices/common_patches/h7_ethernet.yaml
```diff
@@ -0,0 +1,2082 @@
...
+            description: Sub-nanosecond Increment Value
+            bitOffset: 8 
+            bitWidth: 8
+      MACSTSR:
+        description: System time seconds register
+        addressOffset: 0x0B08
+        resetValue: 0x00000000
+        access: read-only
+        fields:
+          TSSS:
```
adamgreig: Should be `TSS`

devices/common_patches/h7_ethernet.yaml
```diff
@@ -0,0 +1,2082 @@
...
+            description: Timestamp Sub-seconds
+            bitOffset: 0 
+            bitWidth: 31
```
adamgreig: Should be 32

devices/common_patches/h7_ethernet.yaml
```diff
@@ -0,0 +1,2082 @@
...
+            bitWidth: 2
+          PPSEN0:
+            description: Flexible PPS Output Mode Enable
+            bitOffset: 4 
+            bitWidth: 1
+          PPSCTRL:
+            description: PPS Output Frequency Control
+            bitOffset: 0 
+            bitWidth: 4
+      MACPPSCR:
```
adamgreig: I don't understand what's going on here and briefly thought I was going mad. The reference manual lists `MACPPSCR` twice, both at the same offset, with the same `TRGTMODSEL0` and `PPSEN0` but totally different `PPSCTRL`/`PPSCMD` in the bottom 4 bits. We can't just list both registers identically with the same offset but who knows what the correct field is. Any ideas?

ehntoo: This had me super confused at first after you pointed it out - it looks like what's actually going on is that PPSEN0 modifies the behavior of bits [0:3] - when PPSEN0 is 0, bits 0:3 act as PPSCTRL, but when PPSEN0 is 1, bits 0:3 act as PPSCMD.

That being said, I have no idea what the right course for handling this in the crate is.

ehntoo:
Thanks for the thorough review! I addressed the issues you caught save for the oddity of PPSCMD/PPSCTRL. Perhaps the best way to handle that is just to change the field name to combine the two?

devices/common_patches/h7_ethernet.yaml
ehntoo: Looks like some Intel/Altera parts use the same (or similar) Synopsys IP for their ethernet peripherals, and their docs are a little clearer on this point: https://www.intel.co.jp/content/dam/altera-www/global/ja_JP/pdfs/literature/hb/cyclone-v/hps.pdf

adamgreig:
Updates look good, thanks. Maybe just call it `PPSCMD` and document that it's an alias for `PPSCTRL`?

ehntoo:
Updated the MACPPSCR register to remove the duplicate. I wasn't quite sure how to smash the descriptions together, but this seemed fine to me.

PR #45, opened 2018-09-12
Add IWDG peripherals to the stm32f0xx family

octronics:
Documents the IWDG fields and add it to the devices of the STM32F0 familly.


adamgreig:
Great, thanks!

PR #46, opened 2018-09-12
Add advanced CRC peripheral

octronics:
Extends the basic CRC with reversal mode fields. This peripheral is used in at least the STM32F0 familly. Not all devices have this mode but it seems there is only on svd files per family. So it describe registers and fields that doesn't appear in all devices.

I have documented the POLYSIZE field has well, but I don't know where to put it. Is storing it on `crc_with_polysize.yaml` and include it on the devices that has it seems a good ideas? Please let me know.

### DIFF
--- a/devices/stm32f0x0.yaml
+++ b/devices/stm32f0x0.yaml
@@ -2,6 +2,6 @@ _svd: ../svd/stm32f0x0.svd
 
 _include:
  - ../peripherals/gpio/gpio_v2.yaml
- - ../peripherals/crc/crc_basic.yaml
+ - ../peripherals/crc/crc_advanced.yaml
  - ../peripherals/wwdg/wwdg.yaml
  - ../peripherals/pwr/pwr_v1.yaml
--- a/devices/stm32f0x0.yaml
+++ b/devices/stm32f0x0.yaml
@@ -3,5 +3,6 @@ _svd: ../svd/stm32f0x0.svd
 _include:
  - ../peripherals/gpio/gpio_v2.yaml
  - ../peripherals/crc/crc_advanced.yaml
+ - ../peripherals/crc/crc_with_polysize.yaml
  - ../peripherals/wwdg/wwdg.yaml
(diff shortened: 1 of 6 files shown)


adamgreig:
Looks good, thanks!

> Is storing it on crc_with_polysize.yaml and include it on the devices that has it seems a good ideas? Please let me know.

Yes, this sounds like a good idea. If you want to do that in this PR I'll merge it together.

octronics:
Not all devices on the stm32f0 family have this feature but the svd in not specific to a device. How do you managed that?

adamgreig:
That's annoying. The SVDs all have the POLYSIZE field and we can't restrict specific devices beyond the SVD level, so in this case I think it will be OK to put POLYSIZE in all of them and users will just have to not use it on f03/f04/f05 devices.

I think having the value names start with numbers will cause problems though. Could you rename them to Polysize32, Polysize16, etc please?

octronics:
Is https://agg.io/u/stm32/ yours?

adamgreig:
Yes, it is.

octronics:
Thanks. It help me. It will be good to extend the device coverage at https://stm32.agg.io/rs/ with it. Is its source code public?

adamgreig:
Yes, that page is generated by [makehtml.py](https://github.com/adamgreig/stm32-rs/blob/master/scripts/makehtml.py) and updates automatically from master branch every 15min.

octronics:
makehtml.py generate the Device coverage pages, isn't it? What generates the https://agg.io/u/stm32/ peripherals coverage across family?

adamgreig:
Ah, that one is [htmlcomparesvd.py](https://github.com/adamgreig/stm32-rs/blob/master/scripts/htmlcomparesvd.py). It works on a list of SVDs, so I run it once for each family and then once with a few members of every family and add the hyperlinks at the top manually.

octronics:
Thanks for all. I'll look at it.

PR #50, opened 2018-09-13
Add a sticky header and first columns to Peripheral Compare page

octronics:
Helps a lot when searching on which family peripheral exists. CSS only implementation. Tested on firefox.  
The styling can be improve, but first let me know if that seems a good idea.

### DIFF
--- a/scripts/htmlcomparesvd.py
+++ b/scripts/htmlcomparesvd.py
@@ -91,9 +91,27 @@ def who_has_what_register_fields(parts, peripheral, register):
 
 
 def html_table_peripherals(parts, peripherals):
-    out = ["<table><thead><tr><th>Peripheral</th><th>Address</th>"]
+    out = ["""
+    <style>
+    table thead tr {
+        position: sticky;
+        top: 0;
+        z-index: 5;
+        background: white;
+    }
+    td:first-child,
+    th:first-child {
+        position: sticky;
+        left: 0;
+        z-index: 6;
+        background: white;
+    }
+    </style>
+    """]
+    out.append("<table><thead><tr><th>Peripheral</th><th>Address</th>")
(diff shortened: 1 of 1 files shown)


adamgreig:
I like the idea and it works for me in Firefox but only the column sticks in Chrome, not the top row. Not sure why...

octronics:
Perhaps those can help:
```
position: -webkit-sticky;
position: -moz-sticky;
position: -ms-sticky;
position: -o-sticky;
```
I don't have a chrome browser so can't test them.

adamgreig:
It turns out to be [this](https://stackoverflow.com/questions/44001954/css-only-sticky-table-headers-in-chrome), making the selector `table thead tr th` solves it.

Maybe it would be worth adding to the other tables too?

octronics:
I can take a look at it. But don't expect something before next week.

adamgreig:
:+1: no worries. Thanks for the other PRs too!

octronics:
Refactors page generation by adding the table on a template with minimal markup to have valid web page. Also gives page a title. And cleanups.

adamgreig:
This looks good!

The sticky headers still don't work in Chrome: they need `table thead tr th` for the selector, but we don't have the final `th` here at the moment. I'm happy with all the rest.

octronics:
Done. Can you test it with chrome. While at it, I catch a small visual glitch.

octronics:
Sorry I press the wrong button.

adamgreig:
Works fine, thank you!

adamgreig:
I've updated the hosted version ([here](https://agg.io/u/stm32)) and added a script `htmlcomparesvdall.sh` which generates that version (i.e. runs for each family and makes a top level summary etc).

PR #55, opened 2018-09-16
field access in tables

burrbull:


adamgreig:
Thanks for what looks like some good changes to the HTML script! When you think it's ready to merge, please could you write a short description of what's been done and let me know it's ready to review?

burrbull:
Changes in bitfield tables:
- for each field added access option (so far only `r`, `w` or `rw`)
- centered offset numbers
- if field is in both (high and low) part of table, it will be highlighted

Other:
- added navigation list of peripherals on top of page
- register address offset changed to uppercase

That's all for now.

adamgreig:
Thanks, this looks really good!

PR #59, opened 2018-09-24
Adding some smarts to 'make html'

MattCatz:
Makehtml.py now only generates the html if there is not an associated html file or if the svd is newer. This is because the 'make html' target is much faster than the svd2rust target and is good for quickly checking your work. Now you can do it faster. I also added in a few corrections to the Makefile, such as PHONY designation.


adamgreig:
Looks good, thanks!

PR #62, opened 2018-09-26
Documents DMA peripherals found on F0, F1, F3, L0 and L4 families

octronics:
Use `dma_v1` name as it is available. All DMA registers are documented.


peripherals/dma/dma_v1.yaml
```diff
@@ -0,0 +1,74 @@
...
+      Bit16: [1, "16-bit size"] # of HalfWord
+      Bit32: [2, "32-bit size"] # or Word
+    "[MP]INC":
+      Disabled: [0, "Increment mode disabled"]
+      Enabled: [1, "Increment mode enabled"]
+    CIRC:
+      Disabled: [0, "Circular buffer disabled"]
+      Enabled: [1, "Circular buffer enabled"]
+    DIR:
+      FromPheripheral: [0, "Read from pheripheral"]
```
octronics: Typo

peripherals/dma/dma_v1.yaml
```diff
@@ -0,0 +1,74 @@
...
+      Clear: [1, "Clears the TEIF flag in the ISR register"]
+    "CHTIF*":
+      NoEffect: [0, "No effect"]
+      Clear: [1, "Clears the HTIF flag in the ISR register"]
+    "CTCIF*":
+      NoEffect: [0, "No effect"]
+      Clear: [1, "Clears the TCIF flag in the ISR register"]
+    "CGIF*":
+      NoEffect: [0, "No effect"]
+      Clear: [1, "Clears the GIF, TEIF, HTIF, TCIF flags in the ISR register"]
```
octronics: Is the `NoEffect` necessary?

adamgreig: These are write-only registers and the reset value is 0x0 so I wouldn't have thought so. The only use case I can find is if your code is always writing some value to the field (e.g. writing an argument that is supplied) and needs to either write 0 or 1. However in many other places we've just got the `Clear` without a corresponding `NoEffect` and it's not caused problems yet, so I think leave it out.

adamgreig:
All looks good to me except the typo you identified.

peripherals/dma/dma_v1.yaml
```diff
@@ -52,7 +48,7 @@
       Disabled: [0, "Circular buffer disabled"]
       Enabled: [1, "Circular buffer enabled"]
     DIR:
-      FromPheripheral: [0, "Read from pheripheral"]
+      FromPeripheral: [0, "Read from pheripheral"]
```
adamgreig: The description still contains the typo

octronics:
Thanks! I didn't catch this. I can rebase and merge those little commits if you want.

adamgreig:
Sure, go for it.

PR #71, opened 2018-10-07
Rename ETH_MACIIDR field TD to MD to align with RM0090

### LINKED ISSUE #69: STM32F4 family ETH_MACIIDR RM0090 discrepancy
peripherals/eth/eth_mac.yaml says:

```
  MACMIIDR:
    TD: [0, 0xFFFF]
```

However the RM0090 says the field is support to be called `MD`. I guess the reason it is called `TD` is that some other series (or special sibling) call it `TD` in the documentation but I'm wondering how we want to treat such discrepancies...

therealprof:
Fixes #69

Signed-off-by: Daniel Egger <daniel@eggers-club.de>

### DIFF
--- a/devices/common_patches/ethernet_macmiidr.yaml
+++ b/devices/common_patches/ethernet_macmiidr.yaml
@@ -1,0 +1,5 @@
+Ethernet_MAC:
+  MACMIIDR:
+    _modify:
+      TD:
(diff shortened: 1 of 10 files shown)


adamgreig:
LGTM, thanks!

PR #77, opened 2018-10-22
f0: replace JITOFF_D* by CKMODE in ADC CFGR2 to match RM0360 and RM0091

aurelj:
This fixes misnamed fields in ADC CFGR2 for all stm32f0.
The fields JITOFF_D2 and JITOFF_D4 don't exist in the stm32l0 reference manual, which instead contains a CKMODE field.
This CKMODE field is the same as in stm32l0 (for which svd files are correct).


adamgreig:
Looks good. I thought this was a weird error but it's explained in revision 2 and 4 of the revision history in the reference manuals; they used to be called JITOFF_D2/D4, then DIV2/DIV4, and now CKMODE. I guess the SVDs just never got updated.

PR #80, opened 2018-10-31
stm32f429: add missing RCC reset/enable bits

birkenfeld:
This is ported from a similar PR I made to the `stm32f429` crate. Your version of patching SVDs is much nicer :)

The only things now missing compared to their SVD patch is:

* descriptions for the MAC registers - you already have descriptions for the individual bits though

I guess I can add these to the global MAC yaml?

* a DEVICE_ID peripheral to read out the three unique ID words at 0x1FFF7A10

Do you want to have this included? It's probably relevant to all STMs?

### DIFF
--- a/devices/stm32f429.yaml
+++ b/devices/stm32f429.yaml
@@ -52,6 +52,142 @@ FMC:
       addressOffset: "0x11C"
       alternateRegister: ""
 
+# Add missing bits to RCC_CR
+RCC:
+  CR:
+    _add:
+      PLLSAIRDY:
+        description: PLLSAI clock ready flag
+        bitOffset: 29
+        bitWidth: 1
+        access: read-only
+      PLLSAION:
+        description: PLLSAI enable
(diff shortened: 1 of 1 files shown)


adamgreig:
Thanks, this looks good!

> descriptions for the MAC registers - you already have descriptions for the individual bits though

What descriptions do you mean? I think most of them should already be included for the STM32F429 via https://github.com/adamgreig/stm32-rs/blob/master/peripherals/eth/eth_mac.yaml which it includes.

If there's any missing that's where they should be added.

> a DEVICE_ID peripheral to read out the three unique ID words at 0x1FFF7A10

I'm not keen on adding a whole new peripheral that's not really a peripheral at all to get this. I'm also not sure the address is the same for all STM32s or even all STM32F4s. I'd rather this be a method in an appropriate HAL crate that just does a `core::ptr::read(0x1FFF7A10)` etc than here.

PR #81, opened 2018-11-01
eth_mac: add field descriptions

birkenfeld:
This is purely docstrings. There are, however, two things I have noticed that require renamings (using the STM32F429 ref. manual as the authority):

* `RAM` in `MACFFR` should be `PAM` (Pass all multicast)
* The fields in `MACDBGR` are completely wrong - they seem to be copied from `ETH_MMCCR`.

What is the approach to such changes?

### DIFF
--- a/peripherals/eth/eth_mac.yaml
+++ b/peripherals/eth/eth_mac.yaml
@@ -112,7 +112,7 @@
     BFD:
       Enabled: [0, "Address filters pass all received broadcast frames"]
       Disabled: [1, "Address filters filter all incoming broadcast frames"]
-    RAM:
+    PAM:
       Disabled: [0, "Filtering of multicast frames depends on HM"]
       Enabled: [1, "All received frames with a multicast destination address are passed"]
     DAIF:
@@ -141,6 +141,7 @@
       BFD:
         description: Broadcast frames disable
       RAM:
+        name: PAM
         description: Pass all multicast
       DAIF:
         description: Destination address unique filtering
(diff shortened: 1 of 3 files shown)


birkenfeld:
Ah, by the way: is this the only way to do it (needing the `_modify` blocks separate from the field values)?

peripherals/eth/eth_mac.yaml
```diff
@@ -57,6 +57,39 @@
...
+      RD:
+        description: Retry disable
+      APCS:
+        description: Automatic pad/CRC stripping
+      BL:
+        description: Back-off limit
+      DC:
+        description: Deferral check
+      TE:
+        description: Transmit enable
```
adamgreig: Technically `Transmit*ter* enable` and `Receive*r* enable`.

peripherals/eth/eth_mac.yaml
```diff
@@ -94,12 +127,41 @@
...
+      HPF:
+        description: Hash or perfect filter
+      SAF:
+        description: Source address filter
+      SAIF:
+        description: Source address inverse filtering
+      PCF:
+        description: Pass control frames
+      BFD:
+        description: Broadcast frames enable
```
adamgreig: Broadcast frames **disable**

birkenfeld:
Fixed up. Can you have a look at the question about renamings from the PR description?

adamgreig:
> RAM in MACFFR should be PAM (Pass all multicast)

You can put `name` as well as `description` in the `_modify` block to fix this. The `RAM` above it (L115, where the values are described) will need updating to say `PAM`.

> The fields in MACDBGR are completely wrong - they seem to be copied from ETH_MMCCR.

Ugh. At the moment there's no support for _deleting_ fields (or registers). I'd leave this out of this PR, and open an issue to track it. The solution is probably to add support for deleting fields and then adding the correct fields to the register (which can already be done with `_add`).

PR #83, opened 2018-11-01
stm32f4: add missing RCC bits for all cpus

birkenfeld:
I went through and made the requisite changes part of https://github.com/adamgreig/stm32-rs/commit/ed701bf0e8c95a3644e023c320314f3899f102be for the whole f4xx family.

There's quite a bit of repeat, but in small chunks. Is it worth making common patches for individual parts?


adamgreig:
Thanks! I think it might be worth making a common patch for the ones that show up most, like SPI4/5, SAIPLL, and UART7/8. You could use them in the f429 as well. Ones that only show up once or twice are probably less worthwhile making common patches.

birkenfeld:
Ok, updated.

This also fixes a bug in my commit for the F429 where there were two `RCC:` keys in the YAML, so the first one didn't get used at all.

psycowithespn:
It looks like some of these can't use the common patches. For example, F410 is missing SPI5RST and SPI5LPEN. SPI5EN is already in the original SVD, so using the common patch causes a duplicate type and method.

It looks like 410, 411, 412, and 446 are affected similarly.

birkenfeld:
> It looks like some of these can't use the common patches. For example, F410 is missing SPI5RST and SPI5LPEN. SPI5EN is already in the original SVD, so using the common patch causes a duplicate type and method.

You're absolutely correct, and I should have done a complete local build. I kind of assumed that either svdpatch or svd2rust would complain about these cases... time to write patches for both :)

Also, is there an objection to changing CI to run the full build incl. svd2rust and cargo build with all features? Hmm, of course the latter may well take a very long time...

birkenfeld:
Ok, this change now builds. I noticed that the patches for SPI6 and UART7/8 are now only used in one device; however I haven't checked any non-F4 devices yet. Should I keep them in `common_patches`?

adamgreig:
I'd move the ones that are only used in one place back to that device. At some point I think `common_patches/` is going to need tidying up anyway but better if we can avoid cluttering it too much now.

Could you rename those peripherals in their descriptions (`SPI 4` -> `SPI4`)?

birkenfeld:
Done, but be aware that these are *very* inconsistent already in the fields present in the SVDs.

adamgreig:
Yep, sadly, but we do what we can...

Thanks.

birkenfeld:
BTW, how is the policy on new releases? This is the last thing I'd need to switch from `stm32f429` and sadly I can't use a `git =` dependency with this crate...

adamgreig:
@birkenfeld I've released v0.4.0.

PR #85, opened 2018-11-05
svdpatch: check for conflicts when adding items

birkenfeld:
This is now checked for peripherals, interrupts, registers, and fields.


adamgreig:
Good idea, thanks.

PR #87, opened 2018-11-05
Ethernet: fixup MACDBGR definition

birkenfeld:
NOTE: This depends on #86 being merged to work!


devices/common_patches/ethernet_mac_regs.yaml
```diff
@@ -0,0 +1,57 @@
+Ethernet_MAC:
+  MACMIIDR:
+    _modify:
+      TD:
+        name: MD
```
adamgreig: This is fine here but `devices/common_patches/f4_ethernet.yaml` also contains the same patch; I think we should remove it from there and keep it here.

PR #89, opened 2018-11-05
travis: do a "cargo check" (no codegen) for all crates

birkenfeld:
This shows a failure in stm32l100 that seems legit...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,19 @@
-language: python
-python:
-  - "3.6"
+language: rust
+
+rust:
+  - stable
+
+addons:
+  apt:
+    packages:
+      - python3-yaml
+
 install:
-  - pip install pyyaml
+  - cargo install svd2rust
+  - rustup component add rustfmt-preview
+
+# disabling incremental build gives a nice speedup for check...
 script:
   - cd svd && ./extract.sh && cd ..
-  - make patch
+  - make -j$(nproc) all
+  - CARGO_INCREMENTAL=0 make -j$(nproc) check
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 all: patch svd2rust
 
-.PHONY: patch svd2rust form clean-rs clean-patch clean-html clean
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -1,8 +1,17 @@
...
+language: rust
+
+rust:
+  - stable
+
 install:
-  - pip install pyyaml
+  - wget https://bootstrap.pypa.io/get-pip.py
+  - python3 get-pip.py --user
+  - python3 -m pip install --user pyyaml
```
adamgreig: `python3-yaml` is on the Travis apt whitelist, so maybe we could replace this with
```
addons:
  apt:
    packages:
      - python3
```

Makefile
```diff
@@ -46,6 +50,10 @@ $(1)/src/%/.form: $(1)/src/%/mod.rs
 	mv $$$$(dirname $$<)/lib.rs $$<
 	find $$$$(dirname $$<) -name "*.rs" -exec rustfmt {} +
 	touch $$$$(dirname $$<)/.form
+
+$(1)/src/%/.check: $(1)/src/%/mod.rs
+	cd $(1) && cargo check --target-dir ../target/check/$$* --features $$*
+
```
adamgreig: You should also touch `.check` so make sees it get created.

adamgreig:
I'm tracking down the build failure, by the way. Half resolved.

adamgreig:
If you rebase onto master the build will hopefully succeed.

.travis.yml
birkenfeld: Nice!

birkenfeld:
Passing :fireworks: The build takes about 30 minutes on Travis, which I guess is fair enough, it has to build 46 crates.

Each build uses its own target directory since otherwise the 2 cores offered by Travis can't straightforwardly be used in parallel, since the same crate is built multiple times leading to interlocks. On the other hand, the artifacts that could be shared only take a few seconds to compile.

I also tried building each crate in one go with all supported features, but it led to OOM and timeouts.

adamgreig:
Looks good, thanks!

PR #90, opened 2018-11-15
Change `package.metadata.docs.rs` for stm32l4

nickray:
Generally, would make sense to align documented devices with availability of Nucleo boards.

### DIFF
--- a/stm32l4/Cargo.toml
+++ b/stm32l4/Cargo.toml
@@ -19,7 +19,7 @@ optional = true
 version = "0.6.4"
 
 [package.metadata.docs.rs]
-features = ['rt', 'stm32l4x1', 'stm32l4x5']
+features = ['rt', 'stm32l4x2', 'stm32l4x6']
 
 [features]
 default = []


adamgreig:
Are there any others you think we should change? I originally aimed for a combination of "few/small enough to not fail the doc build" and "cover most peripherals available" since of course people use STM32s in many more contexts than the Nucleos; but it's as good a selection criteria as any. Ideally we'd have all the chips in and I will want to get to that point.

I'm happy to merge this change, just wondering if you'd like to see any others done in the same PR.

nickray:
Mmh, admittedly I'm being egotistical, as our security key uses an L432, so I'd like to point to your official docs ;) Having all the docs built would be best of course. I could offer to figure out the "top" Nucleos for the other chips if you like? This would just be me googling though; also not sure how much docs.rs can handle?

adamgreig:
Sadly how much docs.rs can handle seems to be a very trial-and-error thing; it might have changed recently as they updated how they do the backend too. Of course, to find out requires releasing a new version on crates.io...

nickray:
I built `--all-features` for stm32l4 and tried deploying to Netlify, which however broke Netlify. Admittedly, this tries to deploy 2GB, which is a little over the top. But even deploying on my own server, the Javascript seems to regularly get stuck, even crashing browser tabs; not sure why this kind of documentation needs Javascript in the first place. All in all, I'm tempted to close this pull request, as I fear replacing 4x1/4x5 with 4x2/4x6 might break your next documentation deploy, given there are more peripherals.

Do you think alternatively there might be a way to make the generated documentation's size more reasonable, and possibly suppress the JS functionality? Re. deploying on Netlify the main obstacle seems to be that the docs have "many little files".

adamgreig:
I believe the Javascript is for the search functionality (which runs in the browser using a pre-built index file; on these crates that file is huge so it takes a load of time and memory to load it and search it). I don't think there is any way to disable it short of manually editing the generated files.

Not really sure what the long term solution is. I think the autogenerated documentation is not that useful for svd2rust crates anyway since it can be very difficult to find what you're looking for, so perhaps we need some other form of generated documentation.

adamgreig:
On balance I'm inclined to try this and hope it works; we can always roll it back if it breaks the docs.rs build.

PR #92, opened 2018-11-20
fix APB1ENR1.USBFSEN for l4x2 and l4x3

nickray:
I believe this fixes https://github.com/adamgreig/stm32-rs/issues/91


adamgreig:
Thanks, looks good!

Issue #37, opened 2018-07-15
STM32L4x1, x2: USB_SRAM incorrect


adamgreig:
The SVD for STM32L4x1,4x2 contains a USB_SRAM "peripheral", and a USB_FS peripheral which derives from the USB_SRAM peripheral. Unfortunately this isn't correct: USB_FS should have all the registers inside USB_SRAM (RM0394 43.6.1), and USB_SRAM should have a collection of different registers (RM0394 43.6.2). However, the USB_SRAM really is just a big block of memory, and isn't strictly partitioned into registers -- it's more a set of descriptors which can start at any offset.

So the SVD should really contain USB_FS with all the registers currently in USB_SRAM, and USB_SRAM should cease to exist at all, or at least be a peripheral with no registers.

nickray:
I think L4x1 has no USB, whereas L4x3 does, and its SVD is correct(ish)

PR #94, opened 2018-11-21
attempt to fix USB peripheral for stm32l4x2

nickray:
I hope this is correct!

- strangely, stm32l4x3 is not affected by https://github.com/adamgreig/stm32-rs/issues/37
- I renamed the entire peripheral to just `USB` as this is what both l4x3 and f103 do (considering the latter because I will port `mvirkkunen/stm32f103xx-usb`)
- the interrupt was listed under `USB_FS`, it is missing from both l4x3 and f103...
- I don't know if the previous baseAddress `0x4000_6C00` is needed anywhere?
- regarding the `USB_{ADDR,COUNT}n_{TX,RX}` registers, both l4x3 and f103 don't seem to have them; the `usb-device` implementation seems to hardcode it (https://github.com/mvirkkunen/stm32f103xx-usb/blob/master/src/endpoint.rs#L49) - I may send a separate PR if it makes sense to patch them in

### DIFF
--- a/devices/stm32l4x2.yaml
+++ b/devices/stm32l4x2.yaml
@@ -12,6 +12,27 @@ RCC:
         name: USBFSEN
         description: USB FS clock enable
 
+# cf. <https://github.com/adamgreig/stm32-rs/issues/37>
+# we call the resulting peripheral `USB` instead of `USB-FS`
+# to be aligned with `mvirkkunen/stm32f103-usb`
+
+_modify:
+  USB_SRAM:
+     name: USB
+     # without quotes, get less readable value 1073768448
+     baseAddress: "0x40006800"
+
+USB:
+  _add:
+    _interrupts:
+      # the interrupt is listed under USB_FS, which we delete
+      USB:
+        description: USB event interrupt through EXTI
+        value: 67
+
+_delete:
+  - USB_FS
+
 # Merge the thousands of individal bit fields into a single field for each
 # CAN filter register. This is not only much easier to use but also saves
 # a huge amount of filespace and compilation time etc -- as much as 30% of all


adamgreig:
> strangely, stm32l4x3 is not affected by #37

The mysteries of ST's SVDs only grow.

> I renamed the entire peripheral to just USB as this is what both l4x3 and f103 do (considering the latter because I will port mvirkkunen/stm32f103xx-usb)

Checks out. The reference manual also just calls it "USB" which is my usual source of truth.

> the interrupt was listed under USB_FS, it is missing from both l4x3 and f103...

We should add it to 4x3 and 103 in another PR ideally.

> I don't know if the previous baseAddress 0x4000_6C00 is needed anywhere?

It doesn't really belong in the SVD, though a HAL would have to know about it. It could sensibly be put into the linker script so you could easily place a descriptor there.

> regarding the USB_{ADDR,COUNT}n_{TX,RX} registers, both l4x3 and f103 don't seem to have them; the usb-device implementation seems to hardcode it (https://github.com/mvirkkunen/stm32f103xx-usb/blob/master/src/endpoint.rs#L49) - I may send a separate PR if it makes sense to patch them in

I haven't had a chance to look into this situation but would welcome a separate PR.

nickray:
> > I don't know if the previous baseAddress 0x4000_6C00 is needed anywhere?
> 
> It doesn't really belong in the SVD, though a HAL would have to know about it. It could sensibly be put into the linker script so you could easily place a descriptor there.

Can you explain what you mean by that? E.g. point to a place where this is done? Then I can maybe supply a PR for it.

adamgreig:
I don't think it can be done by the SVD files in any meaningful way. It's more a constant that either the HAL or the end user needs to know, the same way they might know the RAM section that's CCM/DTCM etc. It needs to end up in the linker script sections ideally, which is done per-application at the moment...

Issue #97, opened 2018-11-25
STM32F4 I2C has inconsistent derivedFrom


psycowithespn:
For the F4 series, F410 and F413 both use I2C2 as the RegisterBlock I2CX type. All other F4 devices use I2C3. For reference, F2 uses I2C2, the rest of the ecosystem uses I2C1.

My naive solution would be to mark one of the peripherals as a new base and provide a name of another to strip everything but name, baseAddress, and the interrupts. These would be moved to the new base and the derivedFrom fields of the other instances would get updated accordingly.

This is in reference to an attempt to handle the 3 I2C peripheral instances generically. (See https://github.com/therealprof/stm32f4xx-hal/pull/37)

PR #98, opened 2018-11-27
Allow rebasing peripherals

psycowithespn:
This is the implementation of my idea in #97.

Specify in a device yaml as follows (I'm sure this can be improved, just not sure how):
```yaml
_rebase:
  I2C1: I2C3
```

It was a coin toss as to when the rebase occurs, before or after modifying peripherals.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,13 @@ _add:
                 description: ADC global interrupt
                 value: 18
 
+# Reorder the heirarchy of peripherals with 'deriveFrom'.
+_rebase:
+    # The KEY peripheral steals everything but 'interrupt', 'name',
+    # and 'baseAddress' elements from the VALUE peripheral.
+    # Peripherals that were 'deriveFrom="VALUE"' are now 'deriveFrom="KEY"'.
+    I2C1: I2C3
+
 # An STM32 peripheral, matches an SVD <peripheral> tag.
 # Does not match any tag with derivedFrom attribute set.
 "GPIO*":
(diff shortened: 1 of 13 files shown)


adamgreig:
Nice! I'll have time to check this more carefully tomorrow evening, but perhaps you could include a second commit in this PR showing some example uses in the existing YAMLs? It will help check what it should do and also ensure the CI passes.

adamgreig:
Could you also add an example usage to the project README, since that's where all the patch options are currently documented?

psycowithespn:
Done and done.

I've noticed that this can mess with spacing of the rebased peripheral of the resulting svd. Looks like just the first moved element, though. Is this an issue?

adamgreig:
Thanks!

> I've noticed that this can mess with spacing of the rebased peripheral of the resulting svd.

Just whitespace changes don't matter. You can tweak this using `Element.tail` [like this](https://github.com/adamgreig/stm32-rs/blob/master/scripts/svdpatch.py#L101) if you want.

This looks good to me and I can see the use case for the HALs, but it's definitely a breaking change so I'm going to wait on merging it until svd2rust 0.14 comes out and we can do a big break to 0.5 with the new svd2rust as well.

psycowithespn:
> You can tweak this using `Element.tail` [like this](https://github.com/adamgreig/stm32-rs/blob/master/scripts/svdpatch.py#L101) if you want.

Thanks for the hint. XPath's last() didn't seem to work right, so I had to resort to 2 loops.

adamgreig:
On second thoughts I'll merge this into master here now so we can start using it, and then it will just be included in the next release, which will be 0.5 and come out once svd2rust 0.14 is done.

PR #99, opened 2018-11-27
Reset and clock control for AHB/APB peripherals on stm32f0

jessebraham:
If I understand things correctly (which I'm not exactly confident about :wink:), this should add support for the `APB2ENR` and `APB1ENR` registers for the STM32F0 series of devices.

This information was pulled from the reference for the [STM32F0X0 devices](https://www.st.com/content/ccc/resource/technical/document/reference_manual/cf/10/a8/c4/29/fb/4c/42/DM00091010.pdf/files/DM00091010.pdf/jcr:content/translations/en.DM00091010.pdf) specifically.

`APB2ENR` is missing `USART6EN`, and `APB1ENR` is missing `USBEN`, as they threw errors during compilation and frankly I don't understand the project well enough how to fix it yet.

Let me know if this is correct and/or if any changes need to be made. If I messed it up I can always give it another go. Thanks for all the hard work that has gone into this project!

### DIFF
--- a/devices/stm32f0x0.yaml
+++ b/devices/stm32f0x0.yaml
@@ -1,6 +1,22 @@
 _svd: ../svd/stm32f0x0.svd
 
 RCC:
+  APB2RSTR:
+    _add:
+      USART6RST:
+        description: "USART6 reset"
+        bitOffset: 5
+        bitWidth: 1
+  AHBENR:
+    _add:
+      IOPDEN:
+        description: "I/O port D clock enable"
+        bitOffset: 20
+        bitWidth: 1
+      DMAEN:
+        description: "DMA clock enable"
+        bitOffset: 0
+        bitWidth: 1
   APB2ENR:
     _add:
       USART6EN:
--- a/devices/stm32f0x0.yaml
+++ b/devices/stm32f0x0.yaml
@@ -1,5 +1,13 @@
 _svd: ../svd/stm32f0x0.svd
 
+RCC:
+  APB2ENR:
+    _add:
+      USART6EN:
+        description: "USART6 clock enable"
+        bitOffset: 5
+        bitWidth: 1
+
 _include:
  - ./common_patches/rename_RCC_CIR_HSI14RDYIE_field.yaml
  - ./common_patches/merge_I2C_CR2_SADDx_fields.yaml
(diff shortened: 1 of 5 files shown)


aurelj:
Nice !
Your patch looks OK.

Regarding missing `USART6EN` and `USBEN`, you can handle them by adding something like this at the beginning or your APB2ENR and APB1ENR:
```
  APB2ENR:
    _add:
      USART6EN:
        description: USART6 clock enable
        bitOffset: 5
        bitWidth: 1
  APB1ENR:
    _modify:
      USBRST:
        name: USBEN
```
Note that `USART6EN` is already present in `stm32f0x1.svd`, so you will probably need to add it into individual `devices/stm32f0x?.yaml` files (excluding f0x1).

And while you are at it, the `AHBENR` could benefit of the same treatment. Oh, and `APB2RSTR`, `APB1RSTR` and `AHBRSTR` as well...

jessebraham:
@aurelj , thanks for the input! I'll look into making the suggested changes tonight some time, just wanted to get confirmation on this before I did too much.

adamgreig:
Thanks, the commits so far look good! We could merge them as they are, but @aurelj 's suggestions are good - if you want to try adding those other fields (and even working on the other registers) please go ahead!

jessebraham:
I think with the addition of these last two commits this pull request should cover everything I initially set out to do plus the additional suggestions. I did my best to review everything, but let me know if you spot any issues. Thanks!

aurelj:
@jessebraham Great, thanks for taking care of my suggestions.

Maybe adding `DMAEN` could be done in the common `rcc_f0.yaml`, but that's quite minor.

Otherwise, LGTM.

jessebraham:
Heh, not sure what I was thinking with the `DMAEN` thing :smile: It's obviously trivial to fix, but if it's not an issue I/somebody can update it in a future pull request.

adamgreig:
Thanks again for the PR! I'm happy to merge this in now, but for future reference - when you have an error that's common between many devices, e.g. the USART5/6 and DMAEN here, it's often simpler to make one or two files in `devices/common_patches/` and then include them in the affected devices. It helps keep the changes in one place in case they need altering in the future, too.

PR #103, opened 2018-12-09
Fix handling of UART stop bit on STM32F4

aurelj:
On UART (contrary to USART), the 2 bits stop field of CR2 only accept values 0b00 (1 stop bit) or 0b10 (2 stop bits). So basically, this is actually a 1 bit field.

The currently defined value for Stop2 is incorrect for a 2 bits fields, it should be 2 instead of 1.

This patch actually changes the stop field to only 1 bit, making the value of Stop2 correct, and changing the unsafe .bits() method to a safe .bit() method.

### DIFF
--- a/peripherals/usart/uart_uart.yaml
+++ b/peripherals/usart/uart_uart.yaml
@@ -3,4 +3,4 @@
   CR2:
     STOP:
       Stop1: [0, "1 stop bit"]
-      Stop2: [1, "2 stop bits"]
+      Stop2: [2, "2 stop bits"]


adamgreig:
I think it would be better to leave the field as-is and change the Stop2 value to 2 instead, since technically the field still exists as 2 bits and that's how it's described in the reference manual; it's just not permitted to write 0b01 and 0b11 to the field (which we express by not including them in the enumeration). I'd rather keep the fields aligned to the reference manual where possible.

adamgreig:
Thanks!

PR #107, opened 2018-12-11
Collect registers into arrays

burrbull:


adamgreig:
This looks promising though I won't have time to check it carefully for a couple of days. Is it ready to merge?

Please could you add some text to the README to describe what the purpose is and add the syntax?

burrbull:
No sense in a hurry.
I tested only few tests, so I can't say if there are issues.

burrbull:
Now, I think it's ready to merge.

scripts/svdpatch.py
```diff
@@ -486,6 +543,10 @@ def process_peripheral(svd, pspec, peripheral, update_fields=True):
     pcount = 0
     for ptag in iter_peripherals(svd, pspec):
         pcount += 1
+        # Handle register arrays
+        for rspec in peripheral.get("_array", {}):
+            rmod = peripheral["_array"][rspec]
+            process_peripheral_regs_array(ptag, rspec, rmod)
```
adamgreig: Please could you move array handling to below addition handling? We might need to add/modify/delete registers that will become part of the array, which is impossible if array handling is processed first.

scripts/svdpatch.py
```diff
@@ -329,6 +329,63 @@ def process_register_delete(rtag, fspec):
     for ftag in list(iter_fields(rtag, fspec)):
         rtag.find('fields').remove(ftag)
 
+def spec_ind(spec):
```
adamgreig: A docstring here would help clarify what the function is doing.

scripts/svdpatch.py
```diff
@@ -329,6 +329,63 @@ def process_register_delete(rtag, fspec):
...
+            process_peripheral_register(ptag, name, rmod)
+            ET.SubElement(rtag, 'dim').text = str(dim)
+            ET.SubElement(rtag, 'dimIndex').text = dimIndex
+            ET.SubElement(rtag, 'dimIncrement').text = str(dimIncrement)
+        else:
+            raise SvdPatchError("{}: registers cannot be collected into {} array"
+                    .format(ptag.findtext('name'), rspec))
+    else:
+        raise SvdPatchError("{}: registers {} not found"
+                .format(ptag.findtext('name'), rspec))
```
adamgreig: We can avoid some unnecessary nesting if you swap the if/raise around: check `if dim==0` then raise, check `if not check_offsets` and then raise.

scripts/svdpatch.py
```diff
@@ -354,22 +355,27 @@ def process_peripheral_regs_array(ptag, rspec, rmod):
...
-        registers.append([rname, rtag, rname[li:len(rname)-ri], int(rtag.findtext('addressOffset'), 0)])
+        registers.append([rtag, rname[li:len(rname)-ri], int(rtag.findtext('addressOffset'), 0)])
     dim = len(registers)
-    if dim > 0:
-        registers = sorted(registers, key=lambda r: r[3])
-        regarray = registers[0]
+    if dim == 0:
+        raise SvdPatchError("{}: registers {} not found"
+                .format(ptag.findtext('name'), rspec))
+    else:
```
adamgreig: We don't need the `else` here at all: the `raise` above will return from the function, so the rest of the code can be unindented. Same on the second check later.

adamgreig:
Thanks for this feature! I've merged it as-is since the only remaining change is very minor and we need arrays to back-patch the USB patches from stm32f103xx.

scripts/svdpatch.py
burrbull: In my opinion if we get zero size register specification matching, then patch is incorrect, and we should make signal. Maybe just print("Warn: ...") ?

adamgreig: I agree the patch would be incorrect - we can raise the error as you do here, but then we don't need to put the rest of the code in `else` because it will only run if the error wasn't raised anyway.

burrbull: Ok. Now I understand what you want.

PR #111, opened 2018-12-14
STM32F103 USB patches

adamgreig:
This is actually all I could find in the stm32f103xx patch file related to USB; I'm not sure if it will be enough to fix the issues you were having @therealprof? I think `USBPRE` was the main one?

If this is enough I'd like to get 0.5 released ASAP with svd2rust 0.14 support; we can always add more of the stm32f103xx patches later too. I think combined with the new array support from #107 we should be good for USB on STM32F103 at least, which will hopefully enable stm32f103xx-usb to work too. @mvirkkunen?

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,8 +50,8 @@ devices should be supported to some level.
 
 * Install `svd2rust`: `cargo install svd2rust`
 * Install `form`: `cargo install form`
-* Install latest rustfmt: `cargo uninstall rustfmt; rustup component add rustfmt-preview`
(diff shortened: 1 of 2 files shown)


therealprof:
@adamgreig Works nicely!

adamgreig:
Thanks for checking!

therealprof:
As soon as the new version has landed we can enable the stm32f1xx-hal CI and think about a first release on crates.io.

PR #114, opened 2018-12-15
USART6 common patch for STM32F0, move DMAEN from device to rcc_f0.yaml peripheral file

jessebraham:
This is mostly just cleaning up the changes I made in #99 

`DMAEN` was moved into the common `rcc_f0.yaml` file rather than being repeated in each individual device file, and I created a common patch for the `USART6` peripheral on F0-series devices.

### DIFF
--- a/devices/common_patches/f0_usart6.yaml
+++ b/devices/common_patches/f0_usart6.yaml
@@ -1,0 +1,13 @@
+RCC:
+  APB2RSTR:
+    _add:
+      USART6RST:
+        description: "USART6 reset"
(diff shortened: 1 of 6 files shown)


adamgreig:
Thanks, looks good. In general I'd rather keep adding/removing registers to patches instead of in the peripheral files -- ideally the device files/patches generate a good SVD and the peripheral files add enums/ranges (and maybe more descriptions) to the peripherals, but it's not a hard split.

jessebraham:
I'll keep that in mind moving forward, thanks!

PR #118, opened 2018-12-17
F0 RCC registers

octronics:
Documents some F0 family RCC registers.

### DIFF
--- a/peripherals/rcc/rcc_f0.yaml
+++ b/peripherals/rcc/rcc_f0.yaml
@@ -403,3 +403,24 @@ RCC:
       Div_14: [13, "PREDIV input clock divided by 14"]
       Div_15: [14, "PREDIV input clock divided by 15"]
       Div_16: [15, "PREDIV input clock divided by 16"]
+  CR2:
+    HSI48CAL: [0, 255]
+    HSI48RDY:
+      _read:
+        NotReady: [0, "HSI48 oscillator ready"]
+        Ready: [1, "HSI48 oscillator ready"]
+    HSI48ON:
+      "Off": [0, "HSI48 oscillator off"]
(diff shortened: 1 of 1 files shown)


peripherals/rcc/rcc_f0.yaml
```diff
@@ -350,3 +385,57 @@ RCC:
...
+      Div_15: [14, "PREDIV input clock divided by 15"]
+      Div_16: [15, "PREDIV input clock divided by 16"]
+  CFGR3:
+    "USART*SW":
+      PCLK: [0, "PCLK selected as USART clock source"]
+      SYSCLK: [1, "SYSCLK selected as USART clock source"]
+      LSE: [2, "LSE selected as USART clock source"]
+      HSI: [3, "HSI selected as USART clock source"]
+    USBSW:
+      HSI48: [0, "HSI48 selected as USB clock source"]
```
adamgreig: Annoyingly on the F0x0 a `0` here is "USB clock disabled" rather than "HSI48 selected as USB clock source" according to RM0360. We could take this field out of this file and split it into two files, one for the F0x0 and one for the rest of the family?

aurelj: BTW, I have a patch set mostly ready which among other things is adding documentation for RCC fields of F0x1, F0x2 and F0x8 which are not in F0x0.
I was just waiting for this PR to be merged to rebase my patch set and send a new PR.
Having different documentation for USBSW/HSI48 would be trivial after my patch set.
So if you want, you could just merge this PR as is and I will add a fix for this USBSW in my patch set.

adamgreig: Perfect, let's do that. Thanks.

PR #125, opened 2018-12-30
Makefile: simplify rustfmt call

aurelj:
rustfmt will recurse through submodules itself.

This is required with new version of `form` which generate modules as xxx.rs instead of xxx/mod.rs.


adamgreig:
Thanks, looks good!

PR #127, opened 2019-01-03
Fix USB OTG FS peripheral

albru123:
This should fix some of the flaws the USB OTG FS peripheral has for F4 devices. I currently included the fixes in all devices with the same reference manual I used (RM0090), but it possibly applies to all devices with the same OTG FS macrocell (e.g. F103)


devices/common_patches/otg_fs_fixes.yaml
```diff
@@ -0,0 +1,96 @@
...
+
+  DIEPCTL0:
+    _modify:
+      EPENA:
+        access: read-write
+
+  DOEPCTL0:
+    _modify:
+      EPENA:
+        access: read-only
```
adamgreig: Is this correct? From looking at the RM it seems like it should be write-only if anything.

devices/common_patches/otg_fs_fixes.yaml
```diff
@@ -0,0 +1,96 @@
...
+      displayName: DIEPMSK
+
+    FS_DOEPMSK:
+      name: DOEPMSK
+      displayName: DOEPMSK
+
+  DIEPCTL0:
+    _modify:
+      EPENA:
+        access: read-write
```
adamgreig: This one is marked `r` in the RM so I'm not really sure what's going on with `EPENA`.

albru123: The RM is wrong on this one (actually there are more access mistakes like that in the RM). It makes no sense for the bit to be `r`, because it would prevent from using the USB endpoint completely.

See this: <https://community.st.com/s/question/0D50X0000AAIG4fSQH/f429-usb-otg-in-endpoint-0-enable-bit-is-readonly-ref-manual-mistake>

devices/common_patches/otg_fs_fixes.yaml
albru123: This was supposed to be `EPDIS`, not `EPENA` actually. Though the `EPENA` bit might be read-only too, as it should have no effect on the endpoint operation.

I'll change it to EPDIS for now, but both `DOEPCTL0` and `DIEPCTL0` probably need to be discussed with someone from ST who actually has access to more documentation about this if we want to get it 100% correct.

PR #128, opened 2019-01-04
Docupdates

x37v:
* updated links that were redirecting from old repo
* fixed broken device crate links that wern't working because of case sensitivity
* added markdown to generated doc strings so we can get links in the docs


adamgreig:
Thanks! The changes will take effect in the next release.

PR #133, opened 2019-01-05
Faster Travis Builds

HarkonenBade:
Make the `make check` invocation run linearly (`-j1`) but with a common target dir to prevent rebuilding of deps.


adamgreig:
Nice, looks like an easy win.

PR #135, opened 2019-01-07
Stm32f0x0128 SPI

x37v:
Stm32f0x0,1,2,8 now have SPI enumerations and some updated field names.

`spi_common.yaml` includes registers that the stm32f0x0 doesn't have, so I extracted those into a common file and then include that in `spi_v1.yaml` and `spi_v2.yaml`.

I'm open to name changes for files, I am not sure what the scheme is for that.


aurelj:
Great ! Patches look good to me.

x37v:
> Great ! Patches look good to me.

@aurelj thanks! did you mean to click `approve` or is that not the process here?

adamgreig:
Thanks for the PR @x37v! The changes all look good but I'd like one tweak to the file names to make the structure more clear: could you rename `spi_f0x0` to `spi_v2_without_UDR_CHSIDE`? It seems a bit weird to me that `spi_v2` (which so many platforms use) is "based on" the `spi_f0x0`, but if we rename it it makes more sense to me.

x37v:
@adamgreig I'm with you on the naming, wasn't quite sure how to name it, will address that when I'm back at home.

x37v:
okay @adamgreig, updated!

jessebraham:
This looks great, thanks @x37v !

adamgreig:
Great, thanks!

PR #139, opened 2019-01-14
Rcc documentation for stm32f2

aurelj:
This is just a very quick pass at stm32f2 RCC. It is probably not exhaustive, and I may have a deeper look at stm32f2 later on. Still I think this quick PR should already be quite useful and I thought it might be worth considering if you want to make a release.


adamgreig:
Good timing, this just snuck in for the release!

PR #144, opened 2019-01-27
Add the missing usart interrupts for the stm32f401.

jkristell:
Fiddling around with the USARTs on the stm32f401 i was noticing the interrupts was missing. I'm not sure this is the right way to add them, and the other MCUs in the family are probably missing them as well. But this seems to work for this MCU at least.

The information from the stm32fxx-hal[1] could probably be used to add the right USARTs to the right device.

[1] https://github.com/stm32-rs/stm32f4xx-hal/blob/master/src/serial.rs


adamgreig:
Thanks, and sorry it's taken so long to merge!

jkristell:
Thanks!

PR #150, opened 2019-02-09
Document SPI registers for F1, F2 and F3

aurelj:


peripherals/spi/spi_v1_without_FRF_FRE_UDR_CHSIDE.yaml
```diff
@@ -0,0 +1,10 @@
+# SPIv1 is used on F1 and F4
+
+_include:
+  - "spi_common.yaml"
+
+"SPI*,I2S*":
+  CR1:
+    DFF:
+      EightBit: [0, "8-bit data frame format is selected for transmission/reception"]
+      SixteenBit: [16, "16-bit data frame format is selected for transmission/reception"]
```
adamgreig: Should be `1` instead of `16`?

aurelj: Indeed, right. Fixed now.

PR #153, opened 2019-02-16
Uart fixes and documentation on F1, F2 and F7

aurelj:
This also add support for a new `_derive` instruction to mark an existing peripheral as derivedFrom another one, and allows to set a derivedFrom for peripheral `_add`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,16 @@ _add:
                 description: ADC global interrupt
                 value: 18
 
+# A whole new peripheral can also be created as derivedFrom another peripheral.
+_add:
+    USART3:
+        derivedFrom: USART1
+        baseAddress: "0x40004800"
+        interrupts:
+            USART3:
+                description: USART3 global interrupt
+                value: 39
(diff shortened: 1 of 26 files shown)


aurelj:
> Sorry this has been sitting so long! The UART changes all look fine to me, but the new `_derive` functionality has already been implemented as `_rebase`; do you think you could update the PR to remove that and leave the rest?

Nope, `_rebase` and `_derive` are different things:
- `_rebase` is when you have I2C3 fully defined and I2C1 marked as derivedFrom I2C3 while you want the opposit.
- `_derive` is when you have both I2C1 and I2C3 fully defined and identical, and that you want to remove the definition of I2C3 and replace it by a derivedFrom I2C1.

So I think the new `_derive` is useful and should be kept.

adamgreig:
Aha, got you. Okay.

adamgreig:
Thanks for the explanation, I've updated the README to clarify it too.

PR #159, opened 2019-03-01
Fix stm32f3 README about what module to use for STM32F318 and STM32F378

aurelj:
STM32F318 is documented in RM0366 and not in RM0316 so it should use
the stm32f301 module.

STM32F378 is documented in RM0313 and not in RM0316 so it should use
the stm32f373 module.

Also the st.com links for stm32f373 and stm32f3x8 where inverted.

### DIFF
--- a/stm32_part_table.yaml
+++ b/stm32_part_table.yaml
@@ -125,6 +125,7 @@ stm32f3:
     rm_url: https://www.st.com/resource/en/reference_manual/dm00094350.pdf
     members:
       - STM32F301
+      - STM32F318
 
   stm32f302:
     url: https://www.st.com/content/st_com/en/products/microcontrollers/stm32-32-bit-arm-cortex-mcus/stm32-mainstream-mcus/stm32f3-series/stm32f302.html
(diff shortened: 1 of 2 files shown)


aurelj:
I just noticed that this part of the README is generated from stm32_part_table.yaml so I also updated it.

adamgreig:
Good spot, thanks. It's pretty annoying that the 318 and 378 are listed on the F3x8 page but share the RM with the 301 and 373! I wonder if it would be better to split each into their own section with a link to the 3x8 page but to the correct respective RM?

aurelj:
Not sure what you mean by separate section ? They surely shouldn't have a separate module ?
Do you mean like 2 lines in the tables both with stm32f301 as the module and RM0366 but with different devices and links ? I don't think the current stm32_part_table.yaml format can acomodate for this.
Anyway, having the 318 and 378 listed along the correct module is a quite important fix, finding the correct web page on the st website for a specific device is not hard at all...

adamgreig:
You're right, I guess there's not really any way to split them up again. Let's go with your proposed changes.

PR #119, opened 2018-12-19
Documents some USB registers

octronics:
Documents the USB peripheral on F0, F1, F3, L0 and L1 families.

- Unprefix USB peripherals
- Documents registers.
- Rename some fields.
- Partial revert a commit that document USB field for stm32f103 device.

I write those some time ago. Perhaps they are good enough to be merged in or perhaps not.

PR #162, opened 2019-03-06
Adds documentation for USB registers

lichtfeind:
This adds Documentation for some of the USB registers and merges the EP*R register for some devices in an array.

The pull request is based of https://github.com/stm32-rs/stm32-rs/pull/119 with the following changes: 

 * dropped USB_SRAM rename for stm32l4 as this does not build and i'm not using that device.
 * add the EP*R array to STM32f0x2

I tested this for the STM32F042K6.


adamgreig:
Thanks for these (and thanks @octronics for #119 in the first place). Sorry it's taken so long to review!

PR #166, opened 2019-03-08
xmlformat

burrbull:
Format patched svds with `xmllint --format`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ addons:
   apt:
     packages:
       - python3-yaml
+      - libxml2-utils
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ addons:
   apt:
     packages:
       - python3-yaml
-      - libxml2-utils
+#      - libxml2-utils
 
 env:
   global:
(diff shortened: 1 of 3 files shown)


adamgreig:
I like the idea of this a lot, but I'm not sure how much it will change the existing SVD content, making it harder to diff the patched/unpatched SVDs. Not that it's especially easy already... Any idea how much this changes the stock parts of the SVD? I wonder if the Python XML writer can be made to tidy the output instead.

burrbull:
On the contrary, it simplifies the comparison and verification of the result. I used it many times with SVD and never had problems or lost data.

adamgreig:
Do you mean running it on the stock SVD files does not change them, or do you run it on both before comparing?

burrbull:
I ran it on patched files. But we can add this into `extract` script to check stock SVDs also.

burrbull:
Or I can separate it in additional instruction in Makefile, like:
`make svdformat`

aurelj:
This would indeed be a very useful feature to compare SVDs, but it wouldn't be useful for someone who only want to generate the final crates.
So I think that adding a new instruction in the Makefile that would format both the original SVD and the patched one would be a great solution. Something like:
`make xmlformat`

burrbull:
I've tried to create separate command `svdformat`, but I'm not sure I do all right.
I'm ignoramus in makefiles.

adamgreig:
I've merged these along with a change to the Makefile so it outputs `%.svd.patched` alongside the other SVDs, which I think is cleaner but still serves the same purpose.

PR #167, opened 2019-03-08
tim CCRx

burrbull:
~~Blocked by #https://github.com/japaric/svd/pull/68~~


burrbull:
I found simple workaround. It's strange I did not think about this before.

adamgreig:
I'm fine with these changes, but I notice svd2rust does _not_ convert CCR into arrays with this, I think because we get `dimIndex` equal `1,2,3,4` but svd2rust requires `0,1,2,3` to convert to an array?

https://github.com/rust-embedded/svd2rust/blob/4e94f2f/src/generate/peripheral.rs#L674-L682

Are there any changes to make before merging this?

burrbull:
> I'm fine with these changes, but I notice svd2rust does _not_ convert CCR into arrays with this, I think because we get `dimIndex` equal `1,2,3,4` but svd2rust requires `0,1,2,3` to convert to an array?
> 
> https://github.com/rust-embedded/svd2rust/blob/4e94f2f/src/generate/peripheral.rs#L674-L682
> 
> Are there any changes to make before merging this?

All is as it should work.
The goal of PR is not to create CCR[4] array. The goal is to crete 1 type CCR from several types CCR1, CCR2, etc.

Issue #136, opened 2019-01-10
STM32F411 missing USART interrupts


korken89:
As the title states, so the issue is known until there is a patch.

musitdev:
Same for STM32F7xx some USART / UART interrupt are missing like the USART6. Can't find why because they are declared in f7_interrupts.yaml.

PR #170, opened 2019-03-08
Allow all interrupts to be deleted by svdpatch.py and fixed adding interrupts to derivedFrom peripherals

### LINKED ISSUE #163: Can't delete or modify interrupts?
Hey! I'm working on generating patches from the cube mx db to fix the interrupts for the stm32f411 but I can't find a way to delete or modify interrupts in a yaml patch.

The readme suggests 
```
ADC_Common:
  _modify:
     _interrupts: 
      FPU: 
        value: 123
```
but that doesn't seem to work.

I've also tried 
```
ADC_Common:
  _delete:
    _interrupts:
      *
```
and various other permutations.

From a brief look at the python it looks like deleting interrupts isn't intended functionality and I can't see how the modify would work on interrupts either but I'm not a fan of python so I could well have missed something.

The SVD for the 411 seems to be a mess, it makes me wonder how STM generate them. How you can have the FPU interrupt under the ADC_Common peripheral if it's being generated from a database I have no idea; I've got to conclude they've got some intern manually editing them to match some other system.

Cheers!

cs2dsb:
The interrupts on the F411 are completely wrong and I've generated a correct list from the CubeMX db.

This change allows you to add
```
_delete:
  - _interrupts
```
at the device level to clear out all interrupts.

I also encountered a bug in svdpatch.py that prevented it adding interrupts to peripherals derivedFrom other peripherals (DMA1 from DMA2 in f411 for example) so I made a change there too.

Currently I've just updated stm32f411.yaml as that's what I'm working on but in theory there is a lot of commonality (and similarly broken/missing interrupts) across the f4 family. I did a quick diff between the f401 and the f411 xml files from CubeMx and the interrupts are identical except the f401 has an additional SPI peripheral and associated interrupt.

I'm going to keep working on the tool I used to generate the correct list of interrupts and will submit more PRs but thought this was a reasonable first step. I also wanted to get some feedback on if you think this is a good way to go? 

Closes #163 and #136 

Cheers,

Daniel

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## Unreleased
+
+* svdpatch.py - added functionality to delete all interrupts from a device level delete
+  ```
+  _delete:
+    - _interrupts
+  ```
+* svdpatch.py - changed behaviour to allow adding interrupts to perpherals that are derivedFrom another peripheral
+* F411 - deleted all interrupts and added a complete list of correct interrupts
+
 ## v 0.6.0 2019-01-14
 
 Family-specific:
--- a/devices/stm32f411.yaml
+++ b/devices/stm32f411.yaml
@@ -5,6 +5,10 @@ _rebase:
   I2C1: I2C3
   SPI1: I2S2ext
 
+_delete:
+  # Nuke the interrupts because they are a mess and a complete list is added below
+  - _interrupts
+
 "SPI*":
   SR:
     _modify:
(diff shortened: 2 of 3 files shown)


adamgreig:
I've implemented the modify/delete interrupt functionality and the missing F411 interrupts in #268. Many interrupts are still associated with the wrong peripheral but this doens't cause any problems that I'm aware of. It might be nice to fix one day.

PR #171, opened 2019-03-09
Link to docs.rs documentation page from each crate README

### LINKED ISSUE #155: Link to docs.rs documentation page from each crate README

jessebraham:
Closes #155


adamgreig:
Thanks for the PR!

As the README files are auto-generated by `scripts/makecrates.py` it would be better to add the change there (in the `README_TPL` template, where `{crate}` will expand to `stm32f0` etc), otherwise the changes will be overwritten every time the script is run (to add a new device or update version number, etc).

The actual change to the README itself looks great.

jessebraham:
Ahh I didn't realize that, thanks. I'll get that fixed up today.

PR #172, opened 2019-03-11
Flash peripheral support for stm32f0xx

jessebraham:
This adds full coverage of the Flash peripheral for all stm32f0 devices. Wasn't too sure what to document for `nBOOT0` and `nBOOT1`; they could probably be improved. Functionally I'm fairly confident it's all good to go, however.

Let me know if anything needs fixing.


jessebraham:
Oops, good eye! Got those fixed up.

PR #173, opened 2019-03-11
F3 syscfg / comp / opamp

aurelj:
split stm32f3 syscfg_comp_opamp into 3 separate devices, syscfg, comp and opamp to match the datasheet and fix all registers that need to be fixed.


adamgreig:
There's a few bits where we could use the new `_split` etc but on the whole I think best to just get it merged first.

PR #175, opened 2019-03-19
rename DBG to DBGMCU

aurelj:
Some of the families name this device DBGMCU and others DBG, even though the Reference Manual for all those families are describing this device in the exact same way.
It is presented in a DBG chapter which contains various kind of devices such as ETM or TPIU as well as DBGMCU, each with their own register set.
So for consistency across families, and for better matching of the reference manual, I propose that we name this device DBGMCU in all families.

### DIFF
--- a/devices/common_patches/dbgmcu.yaml
+++ b/devices/common_patches/dbgmcu.yaml
@@ -1,0 +1,10 @@
+_include:
+ - dbgmcu_rename.yaml
+
+DBGMCU:
+  _strip:
+    - DBGMCU_
+  APB1_FZ:
+    _modify:
+      DBG_IWDEG_STOP:
+        name: DBG_IWDG_STOP
--- a/devices/common_patches/f7_dbgmcu.yaml
+++ b/devices/common_patches/f7_dbgmcu.yaml
@@ -1,0 +1,163 @@
+_add:
+  DBGMCU:
+    description: Debug support
+    baseAddress: 0xE0042000
+    addressBlock:
+      offset: 0x0
+      size: 0x400
+    registers:
+      IDCODE:
+        description: IDCODE
+        addressOffset: 0x0
+        size: 0x20
+        access: read-only
+        fields:
+          DEV_ID:
+            description: DEV_ID
+            bitOffset: 0
+            bitWidth: 12
+          REV_ID:
+            description: REV_ID
+            bitOffset: 16
+            bitWidth: 16
+      CR:
+        description: Control Register
+        addressOffset: 0x4
+        size: 0x20
+        access: read-write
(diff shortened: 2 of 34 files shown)


devices/common_patches/dbgmcu.yaml
```diff
@@ -0,0 +1,17 @@
...
+DBGMCU:
+  _modify:
+    DBGMCU_IDCODE:
+      name: IDCODE
+    DBGMCU_CR:
+      name: CR
+    DBGMCU_APB1_FZ:
+      name: APB1_FZ
+    DBGMCU_APB2_FZ:
+      name: APB2_FZ
```
adamgreig: These could now be replaced with a `_strip: DBGMCU_` (since #187)

devices/common_patches/f7_dbgmcu.yaml
```diff
@@ -0,0 +1,163 @@
...
+          TRACE_IOEN:
+            description: TRACE_IOEN
+            bitOffset: 5
+            bitWidth: 1
+          TRACE_MODE:
+            description: TRACE_MODE
+            bitOffset: 6
+            bitWidth: 2
+      APB1_FZ:
+        description: Debug MCU APB1 Freeze registe
```
adamgreig:
```suggestion
        description: Debug MCU APB1 Freeze register
```

devices/common_patches/f7_dbgmcu.yaml
```diff
@@ -0,0 +1,163 @@
...
+          DBG_CAN1_STOP:
+            description: DBG_CAN1_STOP
+            bitOffset: 25
+            bitWidth: 1
+          DBG_CAN2_STOP:
+            description: DBG_CAN2_STOP
+            bitOffset: 26
+            bitWidth: 1
+      APB2_FZ:
+        description: Debug MCU APB2 Freeze registe
```
adamgreig:
```suggestion
        description: Debug MCU APB2 Freeze register
```

devices/common_patches/dbgmcu.yaml
aurelj: Oh, yes, much nicer !

devices/common_patches/f7_dbgmcu.yaml
aurelj: Fixed

devices/common_patches/f7_dbgmcu.yaml
aurelj: Fixed

aurelj:
New patch set should take care of all your comments.

adamgreig:
Thanks!

PR #176, opened 2019-03-20
Use cortex-m latest 0.6.0

mathk:
This might also require a version change of all sub component. Let me know.

### DIFF
--- a/scripts/makecrates.py
+++ b/scripts/makecrates.py
@@ -48,7 +48,7 @@ license = "MIT/Apache-2.0"
 [dependencies]
 bare-metal = "0.2.4"
 vcell = "0.1.0"
-cortex-m = "0.5.8"
+cortex-m = ">=0.5.8,<0.7"
 
 [dependencies.cortex-m-rt]
 optional = true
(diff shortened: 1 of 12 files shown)


adamgreig:
Thanks! This will require a version bump but we'll do that at release, it doesn't need to be in the PR.

However I'd like to support cortex-m version `">0.5.8,<0.7"` to allow people to use the stm32-rs crates with 0.5.8 _and_ 0.6, since at the moment we'd be completely compatible with both and it helps prevent splitting the ecosystem. Could you update the PR for that please?

mathk:
Done with ">=0.5.8,<0.7"

adamgreig:
Sorry, I should have noticed this the first time.. you also need to make the change in `makecrates.py`

https://github.com/stm32-rs/stm32-rs/blob/master/scripts/makecrates.py#L51

which is what generates the `Cargo.toml` files for each crate, otherwise these changes would get blown away next time.

mathk:
> Sorry, I should have noticed this the first time.. you also need to make the change in `makecrates.py`
> 
> https://github.com/stm32-rs/stm32-rs/blob/master/scripts/makecrates.py#L51
> 
> which is what generates the `Cargo.toml` files for each crate, otherwise these changes would get blown away next time.

Done

adamgreig:
Thanks!

PR #180, opened 2019-03-21
stm32h7x3: add SPI peripheral documentation

jordens:
Two things are not covered:

The SVD has all six SPI peripherals the same and derived from SPI1.
In reality only SPI123 are as documented. SPI456 have 16 bit wide TX and
RX registers and DSIZE is limited accordingly.

Furthermore, there should be explicit byte and half-word access to the
TX and RX data registers. This is how the access-transfer packing ratio
is determined.

### DIFF
--- a/peripherals/spi/spi_v3.yaml
+++ b/peripherals/spi/spi_v3.yaml
@@ -1,0 +1,281 @@
+"SPI*":
+  CR1:
+    IOLOCK:
+      Unlocked: [0, "IO configuration unlocked"]
+      Locked: [1, "IO configuration locked"]
+    TCRCI:
+      AllZeros: [0, "All zeros TX CRC initialization pattern"]
+      AllOnes: [1, "All ones TX CRC initialization pattern"]
+    RCRCI:
+      AllZeros: [0, "All zeros RX CRC initialization pattern"]
+      AllOnes: [1, "All ones RX CRC initialization pattern"]
+    CRC33_17:
+      Disabled: [0, "Full size (33/17 bit) CRC polynomial is not used"]
+      Enabled: [1, "Full size (33/17 bit) CRC polynomial is used"]
+    SSI:
+      SlaveSelected: [0, "0 is forced onto the SS signal and the I/O value of the SS pin is ignored"]
+      SlaveNotSelected: [1, "1 is forced onto the SS signal and the I/O value of the SS pin is ignored"]
+    HDDIR:
+      Receiver: [0, "Receiver in half duplex mode"]
+      Transmitter: [1, "Transmitter in half duplex mode"]
+    CSUSP:
+      _write:
+        NotRequested: [0, "Do not request master suspend"]
+        Requested: [1, "Request master suspend"]
+    CSTART:
+      NotStarted: [0, "Do not start master transfer"]
+      Started: [1, "Start master transfer"]
+    MASRX:
+      Disabled: [0, "Automatic suspend in master receive-only mode disabled"]
+      Enabled: [1, "Automatic suspend in master receive-only mode enabled"]
+    SPE:
+      Disabled: [0, "Peripheral disabled"]
+      Enabled: [1, "Peripheral enabled"]
+
+  CR2:
+    TSER: [0, 65535]
+    TSIZE: [0, 65535]
+
+  CFG1:
+    MBR:
+      Div2: [0, "f_spi_ker_ck / 2"]
+      Div4: [1, "f_spi_ker_ck / 4"]
+      Div8: [2, "f_spi_ker_ck / 8"]
+      Div16: [3, "f_spi_ker_ck / 16"]
+      Div32: [4, "f_spi_ker_ck / 32"]
+      Div64: [5, "f_spi_ker_ck / 64"]
+      Div128: [6, "f_spi_ker_ck / 128"]
+      Div256: [7, "f_spi_ker_ck / 256"]
+    CRCEN:
+      Disabled: [0, "CRC calculation disabled"]
+      Enabled: [1, "CRC calculation enabled"]
+    CRCSIZE: [3, 31]
+    TXDMAEN:
+      Disabled: [0, "Tx buffer DMA disabled"]
+      Enabled: [1, "Tx buffer DMA enabled"]
+    RXDMAEN:
+      Disabled: [0, "Rx buffer DMA disabled"]
+      Enabled: [1, "Rx buffer DMA enabled"]
+    UDRDET:
+      StartOfFrame: [0, "Underrun is detected at begin of data frame"]
+      EndOfFrame: [1, "Underrun is detected at end of last data frame"]
+      StartOfSlaveSelect: [2, "Underrun is detected at begin of active SS signal"]
+    UDRCFG:
+      Constant: [0, "Slave sends a constant underrun pattern"]
+      RepeatReceived: [1, "Slave repeats last received data frame from master"]
+      RepeatTransmitted: [2, "Slave repeats last transmitted data frame"]
+    FTHVL:
+      OneFrame: [0, "1 frame"]
+      TwoFrames: [1, "3 frames"]
+      ThreeFrames: [2, "3 frames"]
+      FourFrames: [3, "4 frames"]
+      FiveFrames: [4, "5 frames"]
+      SixFrames: [5, "6 frames"]
+      SevenFrames: [6, "7 frames"]
+      EightFrames: [7, "8 frames"]
+      NineFrames: [8, "9 frames"]
+      TenFrames: [9, "10 frames"]
+      ElevenFrames: [10, "11 frames"]
+      TwelveFrames: [11, "12 frames"]
+      ThirteenFrames: [12, "13 frames"]
+      FourteenFrames: [13, "14 frames"]
+      FifteenFrames: [14, "15 frames"]
+      SixteenFrames: [15, "16 frames"]
+    DSIZE: [3, 31]
+
+  CFG2:
+    AFCNTR:
+      NotControlled: [0, "Peripheral takes no control of GPIOs while disabled"]
+      Controlled: [1, "Peripheral controls GPIOs while disabled"]
+    SSOM:
+      Asserted: [0, "SS is asserted until data transfer complete"]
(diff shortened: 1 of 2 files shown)


peripherals/spi/spi_v3.yaml
```diff
@@ -0,0 +1,281 @@
+"SPI*":
+  CR1:
+    IOLOCK:
+      Unlocked: [0, "IO configuration unlocked"]
+      Locked: [1, "IO configuration locked"]
+    TCRCI:
+      AllZeros: [0, "All zeros TX CRC initialization pattern"]
+      AllOnes: [1, "All ones TX CRC initialization pattern"]
+    RCRCI:
```
adamgreig: Looks like the RM has `TCRCINI` and `RCRCINI` instead, so maybe we should also patch the SVD and update these?

peripherals/spi/spi_v3.yaml
```diff
@@ -0,0 +1,281 @@
...
+      StartOfFrame: [0, "Underrun is detected at begin of data frame"]
+      EndOfFrame: [1, "Underrun is detected at end of last data frame"]
+      StartOfSlaveSelect: [2, "Underrun is detected at begin of active SS signal"]
+    UDRCFG:
+      Constant: [0, "Slave sends a constant underrun pattern"]
+      RepeatReceived: [1, "Slave repeats last received data frame from master"]
+      RepeatTransmitted: [2, "Slave repeats last transmitted data frame"]
+    FTHVL:
+      OneFrame: [0, "1 frame"]
+      TwoFrames: [1, "3 frames"]
```
adamgreig: Should be `"2 frames"`

peripherals/spi/spi_v3.yaml
```diff
@@ -0,0 +1,281 @@
...
+      Div8: [2, "f_spi_ker_ck / 8"]
+      Div16: [3, "f_spi_ker_ck / 16"]
+      Div32: [4, "f_spi_ker_ck / 32"]
+      Div64: [5, "f_spi_ker_ck / 64"]
+      Div128: [6, "f_spi_ker_ck / 128"]
+      Div256: [7, "f_spi_ker_ck / 256"]
+    CRCEN:
+      Disabled: [0, "CRC calculation disabled"]
+      Enabled: [1, "CRC calculation enabled"]
+    CRCSIZE: [3, 31]
```
adamgreig: This is correct _but_ svd2rust will not generate any safe API to write to this field because it can't stop you writing 0-2 at runtime. One option would be to set this field to `[0, 31]`, which would permit invalid values 0-2 to be written, but I don't think that causes any more serious problem than any other misconfiguration of the peripheral (e.g. `CRCSIZE` must be a multiple of `DSIZE` which we cannot enforce). That way svd2rust generates a safe method to write integers to the field.

peripherals/spi/spi_v3.yaml
```diff
@@ -0,0 +1,281 @@
...
+      EightFrames: [7, "8 frames"]
+      NineFrames: [8, "9 frames"]
+      TenFrames: [9, "10 frames"]
+      ElevenFrames: [10, "11 frames"]
+      TwelveFrames: [11, "12 frames"]
+      ThirteenFrames: [12, "13 frames"]
+      FourteenFrames: [13, "14 frames"]
+      FifteenFrames: [14, "15 frames"]
+      SixteenFrames: [15, "16 frames"]
+    DSIZE: [3, 31]
```
adamgreig: Same comment as `CRCSIZE`: set to `[0, 31]`?

peripherals/spi/spi_v3.yaml
```diff
@@ -0,0 +1,281 @@
...
+
+  CFG2:
+    AFCNTR:
+      NotControlled: [0, "Peripheral takes no control of GPIOs while disabled"]
+      Controlled: [1, "Peripheral controls GPIOs while disabled"]
+    SSOM:
+      Asserted: [0, "SS is asserted until data transfer complete"]
+      NotAasserted: [1, "Data frames interleaved with SS not asserted during MIDI"]
```
adamgreig: Typo, should be `NotAsserted`

peripherals/spi/spi_v3.yaml
```diff
@@ -0,0 +1,281 @@
...
+    SSOE:
+      Disabled: [0, "SS output is disabled in master mode"]
+      Enabled: [1, "SS output is enabled in master mode"]
+    SSIOP:
```
adamgreig: Missing values?

peripherals/spi/spi_v3.yaml
```diff
@@ -0,0 +1,281 @@
...
+    UDRIE:
+      Masked: [0, "Underrun interrupt masked"]
+      NotMasked: [1, "Underrun interrupt not masked"]
+    TXTFIE:
+      Masked: [0, "Transmission transfer filled interrupt masked"]
+      NotMasked: [1, "Transmission transfer filled interrupt not masked"]
+    EOTIE:
+      Masked: [0, "End-of-transfer interrupt masked"]
+      NotMasked: [1, "End-of-transfer interrupt not masked"]
+    DPXPIE:
```
adamgreig: This matches the SVD but the RM says `DXPIE` which I think is correct; we should probably fix the SVD and update the entry here.

peripherals/spi/spi_v3.yaml
```diff
@@ -0,0 +1,281 @@
...
+      Loaded: [1, "Additional number of SPI data to be transacted was reloaded"]
+    MODF:
+      NoFault: [0, "No mode fault detected"]
+      Fault: [1, "Mode fault detected"]
+    TIFRE:
+      NoError: [0, "TI frame format error detected"]
+      Error: [1, "TI frame format error detected"]
+    CRCE:
+      Match: [0, "No CRC error detected"]
+      NoMatch: [1, "CRC error detected"]
```
adamgreig: Do you think `NoError`/`Error` would more clearly match the field name and RM?

peripherals/spi/spi_v3.yaml
```diff
@@ -0,0 +1,281 @@
...
+      TXCRC:
+
+  RXCRC:
+    _read:
+      RXCRC:
+
+  UDRDR:
+    UDRDR: [0, 4294967295]
+
+  CGFR:
```
adamgreig: We should rename this `I2SCFGR` to match the RM.

peripherals/spi/spi_v3.yaml
```diff
@@ -0,0 +1,281 @@
...
+    CKPOL:
+      SampleOnRising: [0, "Signals are sampled on rising and changed on falling clock edges"]
+      SampleOnFalling: [1, "Signals are sampled on falling and changed on rising clock edges"]
+    CHLEN:
+      Bits16: [0, "16 bit per channel"]
+      Bits32: [1, "32 bit per channel"]
+    DATLEN:
+      Bits16: [0, "16 bit data length"]
+      Bits24: [1, "16 bit data length"]
+      Bits32: [2, "16 bit data length"]
```
adamgreig: Descriptions should be 24/32 respectively

peripherals/spi/spi_v3.yaml
```diff
@@ -0,0 +1,281 @@
...
+    PCMSYNC:
+      Short: [0, "Short PCM frame synchronization"]
+      Long: [1, "Long PCM frame synchronization"]
+    I2SSTD:
+      Philips: [0, "I2S Philips standard"]
+      LeftAligned: [1, "MSB/left justified standard"]
+      RightAligned: [2, "LSB/right justified standard"]
+      Pcm: [3, "PCM standard"]
```
adamgreig: Should be `PCM` unless there's a reason otherwise?

peripherals/spi/spi_v3.yaml
```diff
@@ -0,0 +1,281 @@
...
+      Empty: [0, "Rx buffer empty"]
+      NotEmpty: [1, "Rx buffer not empty"]
+
+  IFCR:
+    "*C":
+      _write:
+        Clear: [1, "Clear interrupt flag"]
+
+  TXDR:
+    TXDR: [0, 4294967295]
```
adamgreig: This and `CRCPOLY` might be clearer as `0xFFFFFFFF`

peripherals/spi/spi_v3.yaml
jordens: Ack.

peripherals/spi/spi_v3.yaml
jordens: ACK

peripherals/spi/spi_v3.yaml
jordens: ACK

peripherals/spi/spi_v3.yaml
jordens: ACK

peripherals/spi/spi_v3.yaml
jordens: ACK

peripherals/spi/spi_v3.yaml
jordens: ACK. Added.

peripherals/spi/spi_v3.yaml
jordens: ACK. Renamed.

peripherals/spi/spi_v3.yaml
jordens: ACK.

peripherals/spi/spi_v3.yaml
jordens: ACK

peripherals/spi/spi_v3.yaml
jordens: ACK

peripherals/spi/spi_v3.yaml
jordens: ACK. I assume you mean capitalizing the variant, not shortening the doc.

peripherals/spi/spi_v3.yaml
jordens: ACK

adamgreig:
Thanks!

PR #182, opened 2019-03-27
stm32h7 dma and dmamux fixes

jordens:
* Maybe the prefix stripping could be automated in `svdpatch`. It would also apply to a bunch of other peripherals in that stm32h7 SVD.
* ~~DMAMUX1/DMAMUX2 are handled individually and not by pattern match because the README states that renames must not use pattern matches.~~ Pattern renames seem to work just fine in this case.
* The DMA1 interrupts are in fact listed the SVD, but under wrong names (e.g. `DMA_STR0`) ~~and they don't appear in the crate output~~. AFAICT `svdpatch` does not support renaming interrupts (#163 #170) and therefore they are just added with the correct names/description/values and override the wrong ones.

### DIFF
--- a/devices/stm32h7x3.yaml
+++ b/devices/stm32h7x3.yaml
@@ -356,6 +356,16 @@ DMA1:
         description: DMA1 Stream6
       # DMA1_STR7 is correct
 
+"DMAMUX*":
+  _strip:
+    - DMAMUX1_
+    - DMAMUX2_
+  _array:
+    "C*CR":
+      name: "CCR%s"
+    "RG*CR":
+      name: "RGCR%s"
+
 _include:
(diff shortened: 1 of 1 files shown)


jordens:
Let's hold this until after #187

adamgreig:
@jordens sorry for taking a while to get around to this, do you want to re-do now #187 is in?

burrbull:
These registers could be collected into arrays: DMAMUX2_C0CR, DMAMUX2_C1CR, .. => CR[x]

jordens:
Updated, including C*CR and RG*CR arrays.

adamgreig:
Thanks!

PR #187, opened 2019-04-15
Strip register name prefix

burrbull:
As proposed in #182 .


jordens:
Nice!

adamgreig:
Great idea, thanks!

PR #195, opened 2019-04-28
usb::epr as array, safe access to members

nickray:


adamgreig:
Can you just include `peripherals/usb/usb.yaml` which fully documents the EP?R register?

nickray:
Maybe. I think the f103 version I followed just changed :)

Will check, perhaps consider this PR WIP for now; won't really know until [I actually use it](https://github.com/nickray/stm32l43x-usbd).

adamgreig:
I had a quick try and it definitely applies cleanly, but I haven't checked the reference manual to ensure everything in `usb.yaml` is correct for the L4x2. There's not too much there though -- if everything in `usb.yaml` matches up with the fields in the RM then I think we could just add that to the includes as well. I'm excited to see how your USB crate goes!

nickray:
Turns out `usb_with_LPM.yaml` is the correct patch...

Only one question: Is this correct? https://github.com/stm32-rs/stm32-rs/blob/fa3350020d36f40c67aa13282558ddae82c8f308/peripherals/usb/usb.yaml#L78-L79

The RM defines `BTABLE[15:3]`, is this offset (bits 2:0 reserved, forced to zero) implied by other things? If not, how do I express it in patch language?

adamgreig:
It's correct in that it precisely matches the reference manual which is generally what we try to follow (except for obvious errors). It means when writing the BTABLE field you'll need to shift right by three bits yourself. You could also write to the entire BTABLE _register_ (albeit requiring an unsafe block) with an offset. Since you have to compute the BTABLE offset relative to the SRAM start address anyway I don't think it's too bad to have to do a shift?

adamgreig:
Thanks for checking the patches, looks good!

PR #206, opened 2019-05-03
stm32h7x3: RCC definitions and patches

richardeoin:
* CFGR: rename fields MCOxSEL to MCOx to agree with RM
* BDCR: rename field VSWRST to BDRST to agree with RM. Rename RTCSRC to RTCSEL.
* CRRCR: rename RC48CAL to HSI48CAL
* rename RC48 to HSI48 in CR, CRRCR, CIER, CIFR and CICR
* PLL1DIVR: enumerate all even division ratios for DIVP1. The RM says
  "Note that odd division factors are not allowed." Perhaps a little
  excessive
* DxCCIPR: rename *SRC fields to *SEL to agree with RM
* APB1LRSTR: rename HDMICECRST to CECRST. APB1L(LP)ENR: rename
  HDMICECEN to CECEN.
* AHB1(LP)ENR: rename USBxOTG(LP)EN to USBxOTGHS(LP)EN. rename
  USBxULPI(LP)EN to USBxOTGHSULPI(LP)EN.
* AHB1ENR: move USB2OTGHSULPIEN to bit 18 from 28
* AHB2(LP)ENR: rename CAMITF to DCMI
* AHB3LPENR: rename FLITF to FLASH
* Optional patch to add HSICFGR/CSICFGR for Revision V
  devices. Available devices appear to be Revsion Y, so this is
  commented out

Written against RM0433 Rev 6 04-Apr-2019

PR #204, opened 2019-05-03
stm32h7x3: add missing DBGMCU

jordens:
close #197

* not a breaking change, just adding a missing peripheral

PR #208, opened 2019-05-05
stm32h7 svd v1.4

### LINKED ISSUE #191: stm32h7 ethernet naming
(Except for some registers in `Ethernet_MAC` and the registers in `Ethernet_MMC`):

The STM32H7 ethernet peripherals and registers are currently names as `Ethernet_$X.$X$Y` (where e.g. `$X=DMA` and `$Y=DSR`). The ref manual typically names them `ETH_$X$Y`.

 @ehntoo Does it make sense to remove the `$X.$X` duplication and name that stuff `ETH_$X.$Y` (stripping the `$X` from the register name) or `ETH.$X$Y` (merging the ethernet peripherals)?

jordens:
## Major SVD changes:

* Several other devices have been added. Among the changes is the split of `stm32h7x3` into `stm32h743x`, `stm32h753x`, `stm32h7a3x` (??), `stm32h7b3x` (??).
* The new devices don't vary that much from `stm32h7x3` and the latter still seems to be a good superset. There is some RAM/ECC controller and Accelerator in the h7x7/h7x5 CM4/CM7 devices.
* Most of the upstream changes in the SVD are compatible with the current patches.
* There would be fallout w.r.t. RCC (if the pending #206 gets merged)
* and with Ethernet (added MAC, MTL, MMC peripherals upstream with more registers and fields but with somewhat worse register descriptions)

## Status

* Tested on a real project (with lots of RCC and Ethernet peripheral usage).
* Removed redundant `_strip` patched
* Ethernet MMC registers integrated into MAC
* IMO this would resolve and close #191
* Better Ethernet field descriptions taken kept
* Some wrong Ethernet DMA read-only fields made read-write. The reference manual also shows them read-only but the STM32-HAL happily uses and configures them. Conceptually they also should be read-write.
* A couple Ethernet registers have "Tx" or "Rx" with a lower case "x" in their name. The referenc manual is inconsistent (lower case in the index, upper case in some descriptions). This now uses the lower case "x" which leads to an added underscore in the register name in rust.


richardeoin:
We also need to watch out for revision V / Y here. I assume we want to maintain Y for now, since that's past and current distributor stock.

These SVDs appear to be for Y, so that's okay.

The upstream SVD files appear to reveal some of ST's roadmap - even Google doesn't know about some of these parts! Looks like there's some CM7/CM4 big.LITTLE coming up.

jordens:
Is V actually available or is it some intermediate, experimental or future chip that may get a different name?

richardeoin:
ST claim it's available - ie. You'd get one if you ordered direct

richardeoin:
[AN5312: Migration from RevY to RevV for STM32H743/753 and STM32H750 Value line microcontrollers](https://www.st.com/content/ccc/resource/technical/document/application_note/group1/95/22/7c/0c/57/de/4b/f9/DM00609692/files/DM00609692.pdf/jcr:content/translations/en.DM00609692.pdf)

jordens:
ACK. Interesting. THanks for digging that out.
What about sticking with Y (or the superset of Y and V) and then later decide whether to fork or do something else?

richardeoin:
That's sensible, until V turns up in the wild

jordens:
This is ready for review now. Let me know if you'd like things changed or commits squashed.

adamgreig:
Thanks. Do you think we're best merging this first, then doing #203 #204 and #206 after?

jordens:
Yes. That sequence is fine. I think #203 and #204 may pass cleanly. I'll check once this is merged.
#206 will probably need revision.

adamgreig:
Two unexpected differences between this PR and current master:

* LPIET in ETH_MACLETR seems to have shifted down a few bits
* TQS in MTLTXQOMR has gone from 9 bits to 3 bits wide

Other register and field changes match the latest RM so seem good, though I'm not completely sure we haven't snuck in a few Rev-V items like `BOOSTVDDSEL` and `ODEN`, and the new ADC resolution field update?

Several interrupts have moved in a way that seems bad, though I don't think this is really a problem since we just use the overall list of interrupts:

* TIM4 interrupt is now in TIM5 peripheral
* USART1 interrupt is now in USART2 peripheral, and 2 in 3, and 3 in 4
* RTC_ALARM interrupt now in EXTI peripheral
* ETH_WKUP interrupt now in LPTIM4 peripheral
* SWPMI1 interrupt now in TIM2 peripheral
* WKUP interrupt now in DELAY_Block_SDMMC1 peripheral

I wrote a new tool to help with this sort of comparison which I'll tidy up and put into the repository too.

adamgreig:
Try it out: download https://agg.io/u/h7x3_mmaps.tar.gz, extract, and diff with your favourite diff tool.

jordens:
* LPIET: fixed (to the RM values).
* TQS: kept. The RM has the RX and TX queues at 2 kB size (fig 776, page 2779) which is TXQ=7. I am pretty sure that the width of the field is actually only 3. A 256 kB deep TX queue (0x3ff*256) is unrealistic. Also for RQS the RM has a 3 bit wide field.
* I would propose keeping/allowing Rev-V items as long as they are a superset of the Rev-Y stuff and then forking/splitting Rev-V once that doesn't work anymore (i.e. once someone wants to use the HSICAL/CSICAL stuff).
* IRQ positions: agreed. I had fixed the DMA irqs which were placed and named wrong. The bad placement alone is probably something we can live with, especially since we don't have all tools to handle that.

adamgreig:
LGTM, I agree with your points. Will merge once CI builds and then get on with #203, #204.

jordens:
Nice memory map! This flattened and sorted style is probably the only sensible approach to comparing.

PR #214, opened 2019-05-09
Add PAR to dmav2 cluster

### LINKED ISSUE #213: Where is SxPAR gone?
DMA `RegisterBlock`s used to have and `.s0par` field. That doesn't seem to be the case anymore.
From what I see, all other `Sx` registers have been moved to `.st[x]`, but that doesn't seem to be the case for `SxPAR`.

adamgreig:
Fixes #213 

@burrbull I think you added this `dma_v2.yaml` the first time, was this just a mistake or is there a reason to not have PAR?

### DIFF
--- a/devices/common_patches/dma/dma_v2.yaml
+++ b/devices/common_patches/dma/dma_v2.yaml
@@ -7,11 +7,13 @@
         bitWidth: 4
   _cluster:
     "ST%s":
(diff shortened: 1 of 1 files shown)


burrbull:
Yes. It's mistake. Very big mistake.

adamgreig:
I have released v0.7.1 with just this change.

PR #215, opened 2019-05-10
stm32h7x3: spi `DXPIE`/`TXPIE` enables are in fact read-write

richardeoin:
These two bits may be written by software, although they may also be
cleared by hardware. The RM says "TXPIE is set by software and cleared
by TXTF flag set event."

The bitfield in the RM describes these bits as `rs`, rather than
`rw`. Perhaps why they are mis-classified as read-only in the SVD.


jordens:
Agreed

PR #217, opened 2019-05-19
stm32l4: add nBOOT0 and nSWBOOT0

nickray:


adamgreig:
bors cancel

nickray:
Hmm, ok :) I think the last commit should fix this.

BTW, should `make patch` maybe clean out `.deps`? When moving around the patch file, I got stuck for a moment; it seems generating the contents of `.deps` does not take too long?

PR #222, opened 2019-05-30
Make makehtml.py work with bitRange and on Windows

benpye:
This makes makehtml.py work with `bitRange` as well as `bitOffset`/`bitWidth` - `bitRange` is used in some non ST svd files.

Additionally, on Windows `multiprocessing.Pool` behaves differently - it spawns a new process for each worker, meaning that globals (`args`) aren't available, using a `ThreadPool` instead mitigates this and is more suitable as this should be IO bound, not CPU bound.

### DIFF
--- a/scripts/makehtml.py
+++ b/scripts/makehtml.py
@@ -8,9 +8,13 @@ peripheral and register and their level of coverage.
 """
 
 import shutil
+import sys
 import os.path
 import argparse
-import multiprocessing
+if sys.platform == 'win32':
+    import multiprocessing.pool
+else:
+    import multiprocessing
 import xml.etree.ElementTree as ET
 from jinja2 import Environment, PackageLoader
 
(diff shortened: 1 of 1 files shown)


adamgreig:
> Additionally, on Windows multiprocessing.Pool behaves differently - it spawns a new process for each worker, meaning that globals (args) aren't available, using a ThreadPool instead mitigates this and is more suitable as this should be IO bound, not CPU bound.

I just tried this and get 21.78 seconds runtime with 102% CPU (i.e. single core), whereas master gets 2.82 seconds with 548% CPU, so I'm pretty sure it is actually CPU bound and this ends up being a big regression. On Linux multiprocessing.Pool _also_ spawns a new process for each worker (that's the point) but globals should still be available. How is this usually dealt with on Windows?

No problem to the `bitRange` change but I don't want to merge the multiprocessing bit as-is.

benpye:
Understood - didn't realise that this regressed on Linux. You are correct on Linux it's a new process but the process is forked so it retains the address space of the parent, on Windows this is an entirely separate process.

I'll have to look at this to see if there's some other way of making this work on Windows, though the 'easy' solution might be to use `ThreadPool` on Windows and `Pool` on Linux.

adamgreig:
> I'll have to look at this to see if there's some other way of making this work on Windows, though the 'easy' solution might be to use ThreadPool on Windows and Pool on Linux.

That would be fine with me.

benpye:
Sorry for taking so long, finally got back to fix this!

PR #233, opened 2019-06-16
adc f1

burrbull:


peripherals/adc/adc_f1.yaml
```diff
@@ -0,0 +1,156 @@
...
+        Initialize: [1, "Initialize calibration register"]
+    CAL:
+      _read:
+        Complete: [0, "Calibration completed"]
+        NotComplete: [1, "Calibrating"]
+      _write:
+        Start: [1, "Enable calibration"]
+    CONT:
+      Single: [0, "Single conversion mode"]
+      Continious: [1, "Continuous conversion mode"]
```
adamgreig:
```suggestion
      Continuous: [1, "Continuous conversion mode"]
```

peripherals/adc/adc_f1.yaml
```diff
@@ -0,0 +1,156 @@
...
+    ADON:
+      Disable: [0, "Disable ADC conversion/calibration and go to power down mode"]
+      Enable: [1, "Enable ADC and to start conversion"]
```
adamgreig: We use `Disabled`/`Enabled` in `adc_v2.yaml` so should probably be consistent here.

peripherals/adc/adc_f1.yaml
```diff
@@ -0,0 +1,156 @@
...
+      Cycles1_5: [0, "1.5 ADC clock cycles"]
+      Cycles7_5: [1, "7.5 ADC clock cycles"]
+      Cycles13_5: [2, "13.5 ADC clock cycles"]
+      Cycles28_5: [3, "28.5 ADC clock cycles"]
+      Cycles41_5: [4, "41.5 ADC clock cycles"]
+      Cycles55_5: [5, "55.5 ADC clock cycles"]
+      Cycles71_5: [6, "71.5 ADC clock cycles"]
+      Cycles239_5: [7, "239.5 ADC clock cycles"]
+  "SQR?":
+    "SQ*": [0, 17]
```
adamgreig: Is it worth adding `L` for `SQR1`? At least setting it to [0, 15] would allow safely writing it, or better yet an enum for each number (since you have to write N-1 to the field). In my experience `L` is pretty commonly used.

peripherals/adc/adc_f1.yaml
```diff
@@ -0,0 +1,156 @@
...
+  CR2:
+    EXTSEL:
+      Tim1Cc1: [0, "Timer 1 CC1 event"]
+      Tim1Cc2: [1, "Timer 1 CC2 event"]
+      Tim1Cc3: [2, "Timer 1 CC3 event"]
+      Tim2Cc2: [3, "Timer 2 CC2 event"]
+      Tim3Trgo: [4, "Timer 3 TRGO event"]
+      Tim4Cc4: [5, "Timer 4 CC4 event"]
+      Exti11: [6, "EXTI line 11/TIM8_TRGO event (TIM8_TRGO is available only in high-density and XL-density devices)"]
+      Swstart: [7, "SWSTART"]
```
adamgreig: This list is correct for ADC1 but it seems a shame to leave out ADC2 which has the same list. On the other hand ADC3 is a totally different list and is set to derive from ADC2 here, so adding it to ADC2 would make ADC3 incorrect. Kind of a pain to solve nicely so I'm OK with this as-is if you want, it just means ADC2 and ADC3 won't get enumerations for EXTSEL.

peripherals/adc/adc_f1.yaml
```diff
@@ -0,0 +1,156 @@
...
+    JEXTSEL:
+      Tim1Trgo: [0, "Timer 1 TRGO event"]
+      Tim1Cc4: [1, "Timer 1 CC4 event"]
+      Tim2Trgo: [2, "Timer 2 TRGO event"]
+      Tim2Cc1: [3, "Timer 2 CC1 event"]
+      Tim3Cc4: [4, "Timer 3 CC4 event"]
+      Tim4Trgo: [5, "Timer 4 TRGO event"]
+      Exti15: [6, "EXTI line15/TIM8_CC4 event (TIM8_CC4 is available only in high-density and XL-density devices)"]
+      Jswstart: [7, "JSWSTART"]
```
adamgreig: `DMA` field is missing but I think could be specified here (it's applicable for ADC1 and ADC3, but same problem with ADC2 as mentioned above, so at least adding it for ADC1 would be useful).

peripherals/adc/adc_f1.yaml
burrbull: I thought about adding "_copy" functionality that will make full copy of peripheral registers instead of `derive` it (copy ADC3 from ADC2 in this case) and then modify them independently.

adamgreig:
Any thoughts on `DMA` field? If you want to leave it with `EXTSEL` and resolve later/with DMA3 that's fine and we can merge now, or it could be added just for ADC1 for now.

burrbull:
Done. Forget. Sorry.

adamgreig:
Great, thanks!

bors r+

PR #237, opened 2019-06-18
rtc f1

burrbull:
I made enums for `interrupt enable` fields listen/unlisten.
I suggest to make an exception for such type of fields (I mean interrupt control) to distinguish them from others.

### DIFF
--- a/devices/stm32f101.yaml
+++ b/devices/stm32f101.yaml
@@ -60,6 +60,7 @@ _include:
  - ../peripherals/usart/uart_uart.yaml
  - ../peripherals/usart/uart_usart.yaml
  - ../peripherals/rcc/rcc_f1.yaml
+ - ../peripherals/rtc/rtc_f1.yaml
  - common_patches/dma/dma_v1.yaml
  - ../peripherals/dma/dma_v1.yaml
  - ../peripherals/iwdg/iwdg.yaml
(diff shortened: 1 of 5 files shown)


jordens:
Why shouldn't they be called enabled/disabled?

burrbull:
I'm not against `enabled/disabled`.

I just think it is better somehow differ interrupt control fields from peripheral configuration fields.
Now enabled/disabled are just synonyms for set_bit/clear_bit.

jordens:
The argument for `(en|dis)abled` is predictability and consistency with other configuration bits elsewhere. The prevailing style is to not differentiate fields that enable/disable something where practicable and possible. That avoids situations where these bits could be called any of `mask(ed)?`, `listen(ed|ing)?`, `bypass(ed)?`, `suppressed`, `power_down`, `assert(ed)?`, `high`, `clear(ed)?`, `zero`, `activ(ated|e)`, `off` etc. If you call the `*IE` flags `listen/unlisten` then in some other device/peripheral some field that enables a similar IRQ-like flag could be called differently which would be confusing and make the devices/peripherals less compatible. I would also try to stick to the SVD wording/description as much as possible and reasonable. The SVD calls these flags `*IE` and describes them predominantly as `enabled/disabled`. There should also be a natural and direct match between the enum name and the enum doc wording.

Issue #147, opened 2019-01-30
stm32f4xx TIM3/4 ARR should be 16 bits, is 32


cbiffle:
Per the STM32F4 family reference manual, section 18.2, the timer sizes for TIM2-TIM5 are

> 16-bit (TIM3 and TIM4) or 32-bit (TIM2 and TIM5)

But `stm32::stm32f407::tim3::arr::_ARRW` takes a `u32`. It should take a `u16`.

cbiffle:
This also affects the other timer-sized registers on TIM3/TIM4: CCR1, CCR2, CNT, etc.

jessebraham:
This seems to be related to #132 as well

Issue #132, opened 2019-01-05
stm32f0x0 TIM3 ranges incorrect


x37v:
TIM3 is a 16 bit timer for STM32f0x0, but the patching is done based on a 32bit timer so the ranges are 16 bits too wide.
The SVD indicates that there are `_H`, high values, for these registers but the description indicates that they are only for TIM2, which these devices don't have (I don't think).

adamgreig:
I believe this is essentially fixed now (by afd9b28 / #241).

PR #241, opened 2019-06-24
[timers] make TIM3/4 16-bit and some definitions

richardeoin:
* Where TIM3 was `derivedFrom` 32-bit timer TIM2, `_copy` it and use
  correct 16-bit ranges for `CNT` rather than incorrect 32-bit
* Add some definitions specific to Advanced-control timers TIM1/TIM8,
  in the `BDTR` register
* `OCxPE` definitions (`CCMR` output preload)


burrbull:
Closes #147, #132 .

richardeoin:
I've run `cargo check` with this PR on the following HAL crates, with no errors.

stm32f3xx-hal
stm32l0xx-hal
stm32l4xx-hal
stm32f1xx-hal
stm32f0xx-hal
stm32f4xx-hal

These crates are accessing the whole register in an `unsafe` block eg.
```rust
tim.arr.write(|w| unsafe { w.bits(cast::u32(arr)) });
```

Therefore changing the size of the field doesn't break these crates.

Thoughts @adamgreig ? I'm happy to scale back the changes if this PR is too much.

adamgreig:
Thanks for checking them! I'm happy with the PR, just need a bit more time to review it properly. I think it's close to time for another release so I'll try and get pending PRs cleared and look to release this weekend or early next week.

adamgreig:
Thanks for this PR! The timers are one of the biggest pains to get right across all the devices, I think.

I'm happy with all the contents but ideally the SVD modifications would go into `common_patches/` leaving just the enumeration descriptions in `peripherals/`. That said, in this case it might be less elegant since you've nicely factored out the various `_mixed`/`_16bit` and `_l` options. What do you think? I could be convinced to merge this as-is but if you think it would be easy enough to split into patches vs peripheral descriptions that would be nice.

richardeoin:
Sure, refactoring into `common_patches/` is preferable and wasn't too difficult. Thanks to `M-x dired-do-query-replace-regexp` for the heavy lifting :-)

It does mean there's two trees mapping `tim2345.*` to `tim2_common_.*`, but the mapping in `peripherals/` is simplified as there's no need for `_l` there.

PR #242, opened 2019-06-24
stm32h7x3: RCC additional definitions

richardeoin:
* Also rename `MCO` clock selection to show which all output is selected. A breaking change!


richardeoin:
* Fixed a typo on `RNGSEL` in RM0433. Cross referenced with STM32CubeMX

PR #247, opened 2019-07-06
stm32h7: Rename stm32h7x3 into a total of 4 devices

richardeoin:
stm32h743/753 with revision v versions

The resulting patched SVDs for stm32h753 (not rev v) is identical to
the previous output for stm32h7x3. The stm32h743 is almost identical,
less the CRYP peripheral and interrupts.

A less awful alternative to the symbolic links would be appreciated...

ST have released v1.5 SVDs (we currently have v1.4). However the only
change in the release is to remove/hide the 7a3x / 7b3x SVDs. So the
archive is not updated in this PR.


richardeoin:
CI will fail: merge #241 first

burrbull:
```
__main__.MissingFieldError: Could not find TIM12:CNT.CNT
Makefile:38: recipe for target 'svd/stm32h743.svd.patched' failed
```

burrbull:
bors try

richardeoin:
> __main__.MissingFieldError: Could not find TIM12:CNT.CNT
Makefile:38: recipe for target 'svd/stm32h743.svd.patched' failed

More fallout from #241, now builds for me

> What's with the docs being restricted to the non-V only?

An omission in `makecrates.py`, fixed now.

stm32h7/README.md
```diff
@@ -38,4 +38,4 @@ https://docs.rs/svd2rust/0.14.0/svd2rust/#peripheral-api
 
 | Module | Devices | Links |
 |:------:|:-------:|:-----:|
-| stm32h7x3 | STM32H743, STM32H753 | [RM0433](https://www.st.com/resource/en/reference_manual/dm00314099.pdf), [st.com](https://www.st.com/en/microcontrollers/stm32h743-microprocessors-753.html) |
+| stm32h7x3 | STM32H743, STM32H753 | [RM0433](https://www.st.com/resource/en/reference_manual/dm00314099.pdf), [st.com](https://www.st.com/en/microcontrollers-microprocessors/stm32h743-753.html) |
```
adamgreig: This list is now incorrect; you need to update [stm32_part_table.yaml](https://github.com/stm32-rs/stm32-rs/blob/master/stm32_part_table.yaml#L328) to contain an entry for each generated module (`stm32h743`, `stm32h742v`, etc).

richardeoin: Now resolved, I think

adamgreig: Sorry, I meant that because we split these devices into multiple SVD files and therefore multiple modules and features in the stm32h7 crate, they each need their own higher-level entry in that YAML file, each with just one device member. This part of the README tells users which feature/module to use for a given device.

I think probably the list of features enabled for docs on stm32h7 in makecrates.py will need cutting down too -- having all four there will probably cause docs.rs to OOM kill the process. For other families we've picked just one or two devices that are representative of the family or are more widely used.

richardeoin: Got it, there should be an entry under `sth32h7:` for each feature.

I've set it to only build stm32h743/stm32h743v for docs.rs (`743` seems more
common, on current ST dev boards etc.)

adamgreig:
That's prefect, thanks!

bors r+

PR #248, opened 2019-07-08
Numbers

burrbull:
Breaking change.


richardeoin:
This feels like a regression for me. For example, `Bit12` refers to a specific bit (the 13th LSb) which is not intended here. `TwelveBit` refers to a group of 12 bits as intended. It'll also break a number of HAL crates.

I appreciate the spelling of numbers may be unfamiliar to non-english speakers (particularly 10-20) though.

This isn't worth much debate (see bikeshedding), will fix downstream `h7` if it is merged.

jordens:
I'm also in favor of the current style (i.e. semantic correctness of `FifteenBit` over brevity of `Bit15`).

adamgreig:
I tend to agree with @richardeoin here and prefer `TwelveBit` to `Bit12` for an ADC resolution, but if it were technically possible I think `12Bit` would be better. What do people think of `Bits12`, and similarly `Clocks4`, `Samples2` as an alternative?

adamgreig:
Any thoughts on `Bits12` etc, @richardeoin / @jordens / anyone else?

An issue with `TwelveBit`, beyond being longer and harder to type, is it really doesn't scale well - I think we'll want to keep using `Bytes256` rather than `TwoHundredAndFiftySixBytes` (!), and the same goes for sample and clock counts (`ThirtyNinePointFiveCycles`?). Using `BitsN`, `ClocksN`, `SamplesN` etc allows consistency across various counting fields.

Alternatively we could keep using `TwelveBit` where reasonable, e.g. in ADC resolution, and perhaps in SPI and UART word length, and use numerals for larger numbers. Right now we have a mix of both so it would be nice to at least make a decision.

jordens:
That sounds good to me. Spelling out larger or fractional numbers is indeed unergonomic. `Bits12` for new or otherwise unergonomic enums seems nice to me.
Maybe break the existing enums only when and if there are other breaking changes in the respective area. But if you prefer to harmonize them now then that's fine.

PR #249, opened 2019-07-08
sai

burrbull:


devices/common_patches/sai/sai_v1.yaml
```diff
@@ -9,3 +9,33 @@
       "?SR": {}
       "?CLRFR": {}
       "?DR": {}
+  "?CR1":
+    _modify:
+      OutDri:
+        name: OUTDRIV
+      SAI?EN:
+        name: SAIEN
```
adamgreig: I get why you do this and I think it's the best compromise, but jeez it sure sucks that they have SAIAEN and SAIBEN in otherwise identical register blocks.

peripherals/sai/sai.yaml
```diff
@@ -0,0 +1,173 @@
...
+    MONO:
+      Stereo: [0, "Stereo mode"]
+      Mono: [1, "Mono mode"]
+    SYNCEN:
+      Asynchronous: [0, "audio sub-block in asynchronous mode"]
+      Internal: [1, "audio sub-block is synchronous with the other internal audio sub-block. In this case, the audio sub-block must be configured in slave mode"]
+      External: [2, "audio sub-block is synchronous with an external SAI embedded peripheral. In this case the audio sub-block should be configured in Slave mode"]
+    CKSTR:
+      RisingEdge: [0, "ignals generated by the SAI change on SCK rising edge, while signals received by the SAI are sampled on the SCK falling edge"]
+      FallingEdge: [1, "Signals generated by the SAI change on SCK falling edge, while signals received by the SAI are sampled on the SCK rising edge"]
```
adamgreig: These values are a bit confusing - RM0090 talks about 0 being "data strobing edge is falling edge of SCK". Did another RM have a clearer explanation?

Also, missing the first letter S in Signals.

peripherals/sai/sai.yaml
```diff
@@ -0,0 +1,173 @@
...
+      SlaveRx: [3, "Slave receiver"]
+
+  "*CR2":
+    COMP:
+      NoCompanding: [0, "No companding algorithm"]
+      MuLaw: [2, "μ-Law algorithm"]
+      ALaw: [3, "A-Law algorithm"]
+    CPL:
+      First: [0, "1’s complement representation"]
+      Second: [1, "2’s complement representation"]
```
adamgreig: I think these would be better as `OnesComplement` and `TwosComplement`

burrbull:
Travis complete successful.

PR #252, opened 2019-07-13
Apply eth patches to f2 chips

MabezDev:
This PR applies the same patch as https://github.com/stm32-rs/stm32-rs/pull/73 but for the F2 chips, as they use the same ethernet mac. It might be a good idea to change the patch name from `f4_ethernet.yaml` to something more general, what do you think?


adamgreig:
> It might be a good idea to change the patch name from f4_ethernet.yaml to something more general, what do you think?

Yes please, could you rename them in this PR? Maybe just `f2_f4_ethernet.yaml`.

MabezDev:
Done :)

adamgreig:
It doesn't look like you've moved the patch file or updated the name in the F4 devices it's included in?

MabezDev:
Whoops, a bit of sleep deprivation creeping in I think. That should do it.

PR #253, opened 2019-07-14
fsmc

burrbull:


devices/common_patches/fsmc/fsmc_sramfix.yaml
```diff
@@ -0,0 +1,33 @@
+"FMC,FSMC":
+  BCR1:
+    _add:
+      WRAPMOD:
+        description: WRAPMOD
+        bitOffset: 10
+        bitWidth: 1
+  "BCR[2-4]":
+    _add:
+      CCLKEN:
```
adamgreig: From RM0090 and RM0386, "The CCLKEN bit of the FMC_BCR2..4 registers is don’t care. It is only enabled through the
FMC_BCR1 register.". So I guess that's why it's not in the SVD. Should we include it/does it do something useful on other devices?

devices/common_patches/fsmc/fsmc_sramfix.yaml
```diff
@@ -0,0 +1,33 @@
...
+        description: CCLKEN
+        bitOffset: 20
+        bitWidth: 1
+  BCR?:
+    _add:
+      WFDIS:
```
adamgreig: I don't think RM0090 has WFDIS at all, but it covers e.g. F429 which uses this patch file.

adamgreig: Also, WFDIS in RM0386 is dont-care for BCR2-4, only used in BCR1.

peripherals/fsmc/fsmc_sd.yaml
```diff
@@ -0,0 +1,85 @@
...
+      NoDelay: [0, "No clock cycle delay"]
+      Clocks1: [1, "One clock cycle delay"]
+      Clocks2: [2, "Two clock cycles delay"]
+    RBURST:
+      Disabled: [0, "Single read requests are not managed as bursts"]
+      Enabled: [1, "Single read requests are always managed as bursts"]
+    SDCLK:
+      Disabled: [0, "SDCLK clock disabled"]
+      Mul2: [2, "SDCLK period = 2 x HCLK period"]
+      Mul3: [3, "SDCLK period = 3 x HCLK period"]
```
adamgreig: I wonder if these two should be `Div2` and `Div3` since they decrease the clock frequency. I think so far we've only used `MulX` for _increasing_ clock frequency, and use `Div` even when the RM talks about period.

peripherals/fsmc/fsmc_sd.yaml
```diff
@@ -0,0 +1,85 @@
...
+      Enabled: [1, "Interrupt is generated if RE = 1"]
+    COUNT: [0, 8191]
+    CRE:
+      Clear: [1, "Refresh Error Flag is cleared"]
+
+  SDSR:
+    BUSY:
+      NotBusy: [0, "SDRAM Controller is ready to accept a new request"]
+      Busy: [1, "SDRAM Controller is not ready to accept a new request"]
+    MODES2:
```
adamgreig: Should we include `MODES1` too, or is there a reason to leave it out?

peripherals/fsmc/fsmc_wfdis.yaml
```diff
@@ -0,0 +1,8 @@
+"FMC,FSMC":
+  "BCR?":
+    WFDIS:
+      Enabled: [0, "Write FIFO enabled"]
+      Disabled: [1, "Write FIFO disabled"]
+    CCLKEN:
+      Enabled: [1, "FMC_CLK clock output to external memory devices enabled"]
+      Disabled: [0, "FMC_CLK clock output to external memory devices disabled"]
```
adamgreig: The description here is a bit misleading - the clock output is always enabled during synchronous access, and CCLKEN controls whether it is _also_ output during asynchronous access (i.e. always).

devices/common_patches/fsmc/fsmc_sramfix.yaml
burrbull: I've added some bits to make BCRx identical to collect them into array, but now I'm not sure is it really needed to collect into array.

burrbull:
Split BCR1 and BCR2-4.

Please, review again.

adamgreig:
Looks good, thanks!

PR #254, opened 2019-07-14
some can enums

burrbull:


burrbull:
I think it's better to merge as is.

The remaining changes require a very careful study of the issue.

PR #258, opened 2019-07-16
F1 RCC PLLMUL fix

### LINKED ISSUE #257: rcc_f0_f1_f3_common has two values with the same name
In `rcc_f0_f1_f3_common.yaml`, the `RCC.CFGR.PLLMUL` patch contains `Mul16` twice but with different values.  As a single `enumeratedValue` can't have multiple values, the second one ends up overriding the first.  This means that if the user would manually set `PLLMUL` to the first bit pattern, they would get back `PLLMULR::_Reserved` when reading back the value.

https://github.com/stm32-rs/stm32-rs/blob/f11c937464cf6782d0d19b27c26854ed2cdc0665/peripherals/rcc/rcc_f0_f1_f3_common.yaml#L21-L22

This is sort of an annoying situation.  I think that preferably the SVD format would support specifying multiple bit patterns for one enumeration, and then just using the first of them when writing.  But without changing the file format, the only work-around I can think of is naming the second one something like `Mul16_2`, and then seeing if `svd2rust` could be modified to detect this and merge them together for reads, so that the consumer doesn't have to care what the actual bit pattern was.

burrbull:
Should close #257 .

### DIFF
--- a/devices/stm32f100.yaml
+++ b/devices/stm32f100.yaml
@@ -38,7 +38,7 @@ _include:
  - ../peripherals/usart/uart_common.yaml
  - ../peripherals/usart/uart_uart.yaml
  - ../peripherals/usart/uart_usart.yaml
- - ../peripherals/rcc/rcc_f1.yaml
+ - ../peripherals/rcc/rcc_f100.yaml
  - ../peripherals/rcc/rcc_cfgr2_prediv.yaml
  - common_patches/dma/dma_v1.yaml
  - ../peripherals/dma/dma_v1.yaml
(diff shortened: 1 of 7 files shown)


kyrias:
This looks very wrong to me.  This removes most of the `PLLMUL` values that apply to the non-connectivity line F1's, and adds `Mul6_5` to all F1s even though it only applies to the connectivity line.

AFAICT the proper thing to do would be to separate the connectivity line patches from the rest of the F1 line.

Additionally, calling the last one `MulMax` feels wrong to me, since it doesn't say what the multiplier actually is, hmm.

PR #264, opened 2019-07-21
stm32h7: Rename stm32h7x3: ADC CR boost field is wider for rev v

richardeoin:
Documented in RM0433 Section 24.6.3


richardeoin:
Missed a bit from https://github.com/stm32-rs/stm32-rs/pull/247

PR #271, opened 2019-08-02
generics, svd2rust 0.16

burrbull:
Add support of new svd2rust.

Depends on https://github.com/rust-embedded/svd2rust/pull/362

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,10 +43,11 @@ svd/%.svd.formatted: svd/%.svd.patched
 define crate_template
 $(1)/src/%/mod.rs: svd/%.svd.patched
 	mkdir -p $$(@D)
-	cd $$(@D); svd2rust -i ../../../$$<
+	cd $$(@D); svd2rust -g -i ../../../$$<
 	rustfmt $$(@D)/lib.rs
 	sed "1,10d" $$(@D)/lib.rs > $$@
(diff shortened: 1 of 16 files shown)


adamgreig:
The `lib.rs` files are generated from `makecrates.py` so that needs updating: https://github.com/stm32-rs/stm32-rs/blob/master/scripts/makecrates.py#L89

Makefile
```diff
@@ -24,6 +24,10 @@ RUST_DIRS := $(foreach crate, $(CRATES), \
                $(patsubst devices/$(crate)%.yaml, \
                           $(crate)/src/$(crate)%/, \
                           $(wildcard devices/$(crate)*.yaml)))
+GENERICS := $(foreach crate, $(CRATES), \
+               $(patsubst devices/$(crate)%.yaml, \
+                          $(crate)/src/generic.rs, \
```
adamgreig: This isn't right, we don't need to loop over all devices. It can just be:
```make
GENERICS := $(foreach crate, $(CRATES), $(crate)/src/generic.rs)
```
or even:
```make
clean-rs:
    rm -rf $(RUST_DIRS)
    rm -f */src/generic.rs
```

burrbull: Fixed. Rebased.

PR #272, opened 2019-08-02
Improve USB OTG FS patch, apply patch for F401 and F411

Disasm:


burrbull:
Have you try to use _strip to remove prefixes?

Disasm:
Hmm, no. I will try it.

PR #274, opened 2019-08-06
OTG HS patches and register unprefix

albru123:
Similarly to OTG FS fixes, this patch should improve and unify OTG register naming, fix incorrect register addresses, and remove registers that shouldn't be there and collect appropriate registers into arrays.


devices/stm32f446.yaml
```diff
@@ -130,3 +130,4 @@ _include:
  - ../peripherals/fsmc/fsmc_sd.yaml
  - ../peripherals/tim/tim_ccm_v1.yaml
  - ../peripherals/sai/sai.yaml
+ - common_patches/otg_hs_remove_prefix.yaml
```
Disasm: I don't think it's important, but I placed patch lines between `- common_patches/rtc/rtc_bkpr.yaml` and `- common_patches/tim/tim_ccr.yaml` for consistency.

albru123: Moved :)

Disasm:
Depends on: https://github.com/albru123/stm32-rs/pull/1

albru123:
Merged albru123/stm32-rs#1

PR #275, opened 2019-08-06
Set correct number of nvic priority bits available on stm32l0xx

kitzin:
I was using the stm32l0 crate with RTFM and noticed some funky behavior. The number of NVIC priority bits are two on the m0/m0+, and it was wrong in the SVD file.


adamgreig:
hmmm.... what???

bors retry

PR #276, opened 2019-08-07
stm32h7: Dual core

richardeoin:
Add stm32h747cm7

This is the most common dual core part from this series (dev boards
available etc.). This target is for the BIG core (Cortex-M7).

Add DSI host peripheral.

Some issues remaining to be resolved, particularly ADC and DMAMUX.


richardeoin:
Ha yes, the SVD for this leaves much to be desired.

Some of this is done with `pdftotext` on the RM followed by emacs lisp/regexps. That worked well for the DSI peripheral, but the more complicated dual-bank dual-core flash required manual work.

I live in hope that ST might have a future SVD that fixes at least the ADC common register issues, but who knows.

PR #277, opened 2019-08-09
Use UTF-8 for opening YAML files

albru123:
By default Python uses default system encoding for opening files, which might cause parsing errors, because all the YAML files are UTF-8 (at least to my knowledge).


adamgreig:
bors retry

PR #278, opened 2019-08-09
Mark extracted SVDs as precious to prevent auto deletion by Make

adamgreig:
This prevents Make from deleting the SVD files it considers as intermediate, and then re-creating them using the extraction script.


adamgreig:
bors try

PR #279, opened 2019-08-09
Do not generate .deps on clean

Disasm:


Makefile
```diff
@@ -100,4 +100,6 @@ clean: clean-rs clean-patch clean-html
 	@mkdir -p .deps
 	python3 scripts/makedeps.py $< > $@
 
--include $(patsubst devices/%.yaml, .deps/%.d, $(YAMLS))
+$(PATCHED_SVDS): $(patsubst devices/%.yaml, .deps/%.d, $(YAMLS))
```
adamgreig: Could we stick the `.deps/%.d` dependency on L37 with `svd/%.svd.patched` instead?

Disasm: Hmm, this way make treats dep files as intermediate and removes them after build.

adamgreig: I did a quick test and that seems to work great, ensuring each SVD only depends on its on `.d` file too.

adamgreig: Ah, you're right about the intermediate business. Maybe stick them in .PRECIOUS too?

Disasm: Fixed!

adamgreig:
Thanks! Could you rebase off master, as it now contains `.PRECIOUS` for the `svd/%.svd` files as well.

adamgreig:
Incidentally I think you could do `.PRECIOUS: .deps/%.d` rather than having to build the list.

Disasm:
Yeah, I see. Didn't know about this.

Disasm:
bors r=adamgreig

PR #285, opened 2019-08-26
stm32h7: dualcore based on stm32h7x5 instead

richardeoin:
As far as I can tell the H7x7 and H7x5 are the same die, but binned
for different voltage-temperature ranges (and different
packages). They share RM0399 with no mention of differences.

The `stm32h7x5` SVD has a different set of errors to the `stm32h7x7` SVD,
but overall it's not quite so bad. Therefore use this instead.

I used the diff between the two SVDs to highlight errors and resolved
according to the RM. The previous problems with ADC and DMAMUX are
resolved.

Add separate feature for H755/H757, which has `CRYP` peripheral.


svd/extract.sh
```diff
@@ -3,8 +3,16 @@ for f in vendor/*.zip; do
...
-ln -s stm32h743x.svd stm32h743v.svd || true
-ln -s stm32h753x.svd stm32h753.svd || true
-ln -s stm32h753x.svd stm32h753v.svd || true
-ln -s stm32h7x7_cm7.svd stm32h747cm7.svd || true
+rm stm32h743.svd || true
+rm stm32h743v.svd || true
+rm stm32h753.svd || true
+rm stm32h753v.svd || true
+rm stm32h747cm7.svd || true
+rm stm32h757cm7.svd || true
```
adamgreig: Why remove these files only to recreate them immediately? The `cp` call should be fine overwriting them unless I'm missing something...

devices/common_patches/h7_common_dualcore.yaml
```diff
@@ -220,33 +220,88 @@ _modify:
...
+      RDATA:
+        bitWidth: 32
   SQR1:
     _modify:
       L3:
         name: L
+  SMPR1:
+    _add:
+      SMP0:
+        description: ADC channel 19 sampling time selection
```
adamgreig: Should this be `channel 0`?

svd/extract.sh
richardeoin: Good point, these are vestigial from previous use of `ln -s`

devices/common_patches/h7_common_dualcore.yaml
```diff
@@ -409,18 +496,54 @@ RCC:
...
+    _add:
+      DSIRST:
+        description: DSI block reset
+        bitOffset: 4
+        bitWidth: 1
+  AHB1RSTR:
+    _add:
+      ARTRST:
+        description: ART block reset
+        bitOffset: 15
```
adamgreig: This one should be `bitOffset: 14`

devices/common_patches/h7_common_dualcore.yaml
```diff
@@ -454,12 +584,54 @@ RCC:
...
+        bitOffset: 29
+        bitWidth: 1
+      ITCM1EN:
+        description: D1 ITCM block enable
+        bitOffset: 30
+        bitWidth: 1
+      AXISRAMEN:
+        description: AXISRAM block enable
+        bitOffset: 31
+        bitWidth: 1
```
adamgreig: The RM says `This location is 'Reserved' for register RCC_C1_AHB3ENR` for bits 28-31, so perhaps these additions should be `AHB3ENR` only?

devices/common_patches/h7_common_dualcore.yaml
richardeoin: Agreed

devices/common_patches/h7_common_dualcore.yaml
richardeoin: Agreed

devices/common_patches/h7_common_dualcore.yaml
richardeoin: That's how I interpret that footnote too. Removing from `C1_AHB3ENR`

richardeoin:
I can live with excessive numbers of peripherals (bar bus scalability issues), but it does boggle the mind how ST have so many issues in their own SVDs 🤦‍♀️

PR #288, opened 2019-09-27
stm32h7: split ethernet peripheral for dualcore parts

richardeoin:
Split ethernet peripheral into (MAC, DMA, MTL).

This brings it in line with single core parts.

### DIFF
--- a/devices/common_patches/h7_ethernet_combined_desc.yaml
+++ b/devices/common_patches/h7_ethernet_combined_desc.yaml
@@ -1,0 +1,18 @@
+# Applies to H7 where MAC and MTL registers are combined under
+# the Ethernet_MAC peripheral
+
+# Create separate DMA and MTL periperhals
+_include:
+  - ./h7_ethernet_dma.yaml
+  - ./h7_ethernet_mtl.yaml
+
+Ethernet_MAC:
+  # Delete DMA and MTL parts from MAC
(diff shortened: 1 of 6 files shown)


devices/common_patches/h7_ethernet_combined_desc.yaml
```diff
@@ -0,0 +1,25 @@
...
+  _include:
+    - ./h7_ethernet_desc_mac.yaml
+
+Ethernet_MTL:
+  _include:
+    - ./h7_ethernet_desc_mtl.yaml
+
+Ethernet_MAC:
+  _include:
+    - ./h7_ethernet_desc_mac.yaml
```
adamgreig: Sorry for not looking at this earlier - should this line be `Ethernet_DMA`?

richardeoin: Actually, this is redundant. There's no need for separate description listings for `MTL` or `DMA`, because the descriptions are included in the standalone peripheral definitions here:

```
# Create separate DMA and MTL periperhals
_include:
  - ./h7_ethernet_dma.yaml
  - ./h7_ethernet_mtl.yaml
```

The desc_mac definitions are particularly misleading, as they're already included a few lines above. I'm therefore resolving by removing both description includes.

PR #291, opened 2019-10-04
Fix typo in TIM22RST field name

jonas-schievink:
This is why the HAL had to disable TIM22 on the x2 and x3 chips. I wasn't able to fully test the HAL now since the generated Rust code changed too much (presumably due to a newer svd2rust version), but the "no method named `tim22rst` error is gone now.


jonas-schievink:
Closing in favor of https://github.com/stm32-rs/stm32-rs/pull/292

PR #292, opened 2019-10-04
stm32l0xx RCC fixes

jonas-schievink:


jonas-schievink:
Might as well close this one in favor of https://github.com/stm32-rs/stm32-rs/pull/293

PR #293, opened 2019-10-04
STM32L0 fixes

jonas-schievink:
This builds on top of https://github.com/stm32-rs/stm32-rs/pull/292 and applies all necessary fixes from the L0x1 family to the L0x2 and L0x3 families.

It also fixes a few newly discovered issues in all 3 families.


devices/stm32l0x3.yaml
```diff
@@ -1,18 +1,115 @@
...
+  CCR:
+    # VLCDEN is only available on l0x1 and l0x3
+    # It's also missing in the l0x3 SVD
+    _add:
+      VLCDEN:
+        bitOffset: 24
+        bitWidth: 1
+        description: "VLCD reading circuitry enable"
+        Disabled: [0, "VLCD reading circuitry disabled"]
+        Enabled: [1, "VLCD reading circuitry enabled"]
```
adamgreig: This won't work to add the field values, you need to write them separately outside of the `_add` bit. It would suffice to just add `VLCDEN:` indented to the same level as `_add` after the `description` line.

PR #294, opened 2019-10-13
RCC naming fixups for STM32G4 devices

ajfrantz:
These renamings bring the SVDs into alignment with RM0440.

### DIFF
--- a/devices/common_patches/g4_rcc.yaml
+++ b/devices/common_patches/g4_rcc.yaml
@@ -1,0 +1,237 @@
+# Edits required to match RM0440.
+RCC:
+  _modify:
+    PLLSYSCFGR:
+      name: PLLCFGR
+    CCIPR1:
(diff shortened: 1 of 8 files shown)


devices/common_patches/g4_rcc.yaml
```diff
@@ -0,0 +1,237 @@
...
+      USBSMEN:
+        description: USB device clocks enable during Sleep and Stop modes
+        bitOffset: 23
+        bitWidth: 1
+      I2C3SMEN:
+        description: I2C3 clocks enable during Sleep and Stop modes
+        bitOffset: 30
+        bitWidth: 1
+
+  APB1SMENR1:
```
adamgreig: I think this should be `APB1SMENR2`?

ajfrantz: Good catch.

PR #296, opened 2019-10-20
Update documentation of the makecrates.py script

aurelj:
It was missing build.rs and README.md in the list of files it
autogenerate.


adamgreig:
bors retry

PR #299, opened 2019-10-31
Update STM32F7 SVD files

rfuest:
This PR updates the STM32F7 SVD files to version 1.4 in preparation for adding additional STM32F7 devices in future PRs.

The stm32f7x5 SVD file was split into separate files for the STM32F745 and STM32F765 devices, which I reproduced by also splitting the cargo feature.

The SVD files and also the reference manuals refer to both LTDC interrupts as `LCD-TFT` which results in a conflict. I've changed them to `LTDC` and `LTDC_ER` for now, but there might be better names.


adamgreig:
Thanks for this PR! At first glance this mostly looks good but I'll try and review the changes in detail soon. I did notice it looks like you manually changed the readme, cargo.toml, lib.rs, and build.rs files; they're automatically generated by `scripts/makecrates.py` based on the parts table, so you'll need to update that as well: https://github.com/stm32-rs/stm32-rs/blob/master/stm32_part_table.yaml#L289

rfuest:
OK, I think this wasn't mentioned in the readme. But I've updated the parts table now and did rerun the `makecrates.py` script.

PR #300, opened 2019-11-01
Fix SYSCFG CFGR3 register fields

jonas-schievink:
I have compared the result against all 3 reference manuals and it looks
correct.

`ENREF_HSI48` was accidentally removed in #293, but it's only absent from the l0x1, not the others.

### DIFF
--- a/devices/common_patches/l0_syscfg_cfgr.yaml
+++ b/devices/common_patches/l0_syscfg_cfgr.yaml
@@ -36,12 +36,12 @@ SYSCFG:
       - ENBUF_BGAP_ADC
       - EN_BGAP
     _add:
-      # REF_LOCK is wrongly marked as writeable
+      # REF_LOCK is wrongly marked as write-only, but it's "rs" - read/set
       REF_LOCK:
         description: SYSCFG_CFGR3 lock bit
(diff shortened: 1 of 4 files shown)


devices/common_patches/l0_syscfg_cfgr.yaml
```diff
@@ -35,6 +36,12 @@ SYSCFG:
       - ENBUF_BGAP_ADC
       - EN_BGAP
     _add:
+      # REF_LOCK is wrongly marked as writeable
+      REF_LOCK:
+        description: SYSCFG_CFGR3 lock bit
+        bitOffset: 31
+        bitWidth: 1
+        access: read-only
```
adamgreig: Are you sure this is `read-only`? In RM0367 Rev 5 I see:

image

And its access is marked as "rs" i.e. read/set.

jonas-schievink: You're right, will fix tomorrow

jonas-schievink: Okay, looks like the SVDs are still wrong by marking the bit write-only, so now they're patched to read-write.

PR #302, opened 2019-11-13
Add COMP2 and COMP4_6 interrupts for STM32F3x4

adamgreig:
To-do: check other devices using `common_patches/f3_comp246` to see if they also require these interrupts.

### DIFF
--- a/devices/common_patches/f3_comp1234567.yaml
+++ b/devices/common_patches/f3_comp1234567.yaml
@@ -4,6 +4,10 @@ _include:
 
 COMP:
   _add:
+    _interrupts:
+      COMP7:
+        description: COMP7 interrupt combined with EXTI line 33
+        value: 66
     COMP3_CSR:
       description: control and status register
       addressOffset: 0x24
--- a/devices/common_patches/f3_comp246.yaml
+++ b/devices/common_patches/f3_comp246.yaml
@@ -6,6 +6,13 @@ _add:
       offset: 0
       size: 0x400
       usage: registers
+    interrupts:
+      COMP1_2_3:
+        description: COMP1_2_3 interrupt combined with EXTI lines 21, 22, 29
+        value: 64
+      COMP4_5_6:
+        description: COMP4_5_6 interrupt combined with EXTI lines 30, 31, 32
+        value: 65
     registers:
       COMP2_CSR:
         description: control and status register
--- a/devices/stm32f303.yaml
+++ b/devices/stm32f303.yaml
@@ -22,6 +22,11 @@ RCC:
         name: UART5EN
 
 SYSCFG:
+  _delete:
+    _interrupts:
+      - "COMP*"
+
+SYSCFG:
   _strip:
     - SYSCFG_
   _add:
--- a/devices/stm32f373.yaml
+++ b/devices/stm32f373.yaml
@@ -46,6 +46,11 @@ SYSCFG:
       - BYP_ADD_PAR
 
 COMP:
+  _add:
+    _interrupts:
+      COMP1_2_3:
+        description: COMP1_2_3 interrupt combined with EXTI lines 21, 22
+        value: 64
   CSR:
     _add:
       COMP1_INP_DAC:
@@ -53,6 +58,11 @@ COMP:
         bitOffset: 1
         bitWidth: 1
         access: read-write
+
+CAN:
+  _delete:
+    _interrupts:
+      - COMP1_2
 
 _include:
  - common_patches/4_nvic_prio_bits.yaml
--- a/devices/stm32f3x8.yaml
+++ b/devices/stm32f3x8.yaml
@@ -232,6 +232,11 @@ SYSCFG:
         bitOffset: 8
         bitWidth: 2
 
+EXTI:
+  _delete:
+    _interrupts:
+      - COMP7
+
 _include:
  - common_patches/4_nvic_prio_bits.yaml
  - ./common_patches/merge_I2C_CR2_SADDx_fields.yaml


aurelj:
Those COMP interrupts are missing from the COMP block of all the stm32f3 svd.

Here is a list of those interrupts from the RMs:

- stm32f301:
  * COMP2 (64) COMP2 interrupt combined with EXTI Lines 22 interrupt.
  * COMP4_6 (65) COMP4 & COMP6 interrupts combined with EXTI Lines 30 and 32 interrupts respectively.
- stm32f302:
  * COMP1_2 (64) COMP1 & COMP2 interrupts combined with EXTI Lines 21, 22 interrupts.
  * COMP4_6 (65) COMP4 & COMP6 interrupts combined with EXTI Lines 30 and 32 interrupts.
- stm32f303: (STM32F303x6/8)
  * COMP2 (64) COMP2 interrupt combined with EXTI Lines 22 interrupt.
  * COMP4_6 (65) COMP4 & COMP6 interrupts combined with EXTI Lines 30 and 32 interrupts respectively.
- stm32f303: (STM32F303xB/C/D/E)
  * COMP1_2_3 (64) COMP1 & COMP2 & COMP3 interrupts combined with EXTI Lines 21, 22 and 29 interrupts.
  * COMP4_5_6 (65) COMP4 & COMP5 & COMP6 interrupts combined with EXTI Lines 30, 31 and 32 interrupts.
  * COMP7 (66) COMP7 interrupt combined with EXTI Line 33 interrupt
- stm32f3x4:
  * COMP2 (64) COMP2 interrupt combined with EXTI Lines 22 interrupt.
  * COMP4_6 (65) COMP4 & COMP6 interrupts combined with EXTI Lines 30 and 32 interrupts respectively.
- stm32f3x8: (f328)
  * COMP2 (64) COMP2 interrupt combined with EXTI Lines 22 interrupt.
  * COMP4_6 (65) COMP4 & COMP6 interrupts combined with EXTI Lines 30 and 32 interrupts respectively.
- stm32f3x8: (f358, f398)
  * COMP1_2_3 (64) COMP1 & COMP2 & COMP3 interrupts combined with EXTI Lines 21, 22 and 29 interrupts.
  * COMP4_5_6 (65) COMP4 & COMP5 & COMP6 interrupts combined with EXTI Lines 30, 31 and 32 interrupts.
  * COMP7 (66) COMP7 interrupt combined with EXTI Line 33 interrupt
- stm32f373:
  * COMP1_2 (64) Comparator 1/comparator 2 global interrupts (EXTI21/EXTI22)

Some svd have some of those interrupts defined in the wrong device block:

- stm32f303 have them all in the SYSCFG block
- stm32f3x8 have COMP7 in the EXTI block
- stm32f373 have COMP1_2 in the CAN block

I propose to add COMP1_2_3 and COMP4_5_6 in `common_patches/f3_comp246.yaml` and to add COMP7 in `common_patches/f3_comp1234567.yaml`, and then to handle stm32f373 separatly by adding COMP1_2_3 in `stm32f373.yaml`. What do you think about this ?

adamgreig:
Thanks for the detailed review! Adding the interrupts to `f3_comp*.yaml` makes sense, though we'll also have to delete the incorrect ones, probably in each device file for 303/3x8/373.

I'd rather each device has the correct interrupt names, so e.g. the f301 would have COMP2 not COMP1_2_3 as the interrupt name. Otherwise they won't match the reference manual and be more confusing to use. Do you think it's worth still doing them in a common patch or just adding per-device?

aurelj:
> Thanks for the detailed review! Adding the interrupts to `f3_comp*.yaml` makes sense, though we'll also have to delete the incorrect ones, probably in each device file for 303/3x8/373.

Yes, indeed, deleting will need to be done in each device file.

> I'd rather each device has the correct interrupt names, so e.g. the f301 would have COMP2 not COMP1_2_3 as the interrupt name. Otherwise they won't match the reference manual and be more confusing to use. Do you think it's worth still doing them in a common patch or just adding per-device?

I personally prefer more uniformity (COMP1_2_3 for all devices), but I understand that strictly matching the RM also has its merit (better discoverability).

The point here is that if we choose COMP1_2_3 for the f303, it will also apply to the STM32F303x6/8 which only have a COMP2 (only one svd file for 2 different feature set). And if we don't match exactly the RM for STM32F303x6/8, maybe we also don't need to exactly match the f301 one.

Anyway, I think both options are valid and I will let you choose the one we implement.

adamgreig:
> The point here is that if we choose COMP1_2_3 for the f303, it will also apply to the STM32F303x6/8 which only have a COMP2 (only one svd file for 2 different feature set). And if we don't match exactly the RM for STM32F303x6/8, maybe we also don't need to exactly match the f301 one.

Ugh, that is definitely good reason to just unify the names, so let's use COMP1_2_3 and COMP4_5_6. I'll update this PR soon to do that.

adamgreig:
I've updated this branch, is this what you had in mind? The resulting interrupts (from `scripts/interrupts.py`) are:

```
STM32F3x4
41:64 COMP1_2_3: COMP1_2_3 interrupt combined with EXTI lines 21, 22, 29 (in COMP)
42:65 COMP4_5_6: COMP4_5_6 interrupt combined with EXTI lines 30, 31, 32 (in COMP)

STM32F373
59:64 COMP1_2_3: COMP1_2_3 interrupt combined with EXTI lines 21, 22 (in COMP)

STM32F303
61:64 COMP1_2_3: COMP1_2_3 interrupt combined with EXTI lines 21, 22, 29 (in COMP)
62:65 COMP4_5_6: COMP4_5_6 interrupt combined with EXTI lines 30, 31, 32 (in COMP)
63:66 COMP7: COMP7 interrupt combined with EXTI line 33 (in COMP)

STM32F3x8
63:64 COMP1_2_3: COMP1_2_3 interrupt combined with EXTI lines 21, 22, 29 (in COMP)
64:65 COMP4_5_6: COMP4_5_6 interrupt combined with EXTI lines 30, 31, 32 (in COMP)
65:66 COMP7: COMP7 interrupt combined with EXTI line 33 (in COMP)

STM32F302
59:64 COMP1_2_3: COMP1_2_3 interrupt combined with EXTI lines 21, 22, 29 (in COMP)
60:65 COMP4_5_6: COMP4_5_6 interrupt combined with EXTI lines 30, 31, 32 (in COMP)

STM32F301
54:64 COMP1_2_3: COMP1_2_3 interrupt combined with EXTI lines 21, 22, 29 (in COMP)
55:65 COMP4_5_6: COMP4_5_6 interrupt combined with EXTI lines 30, 31, 32 (in COMP)
```

aurelj:
Yes ! Patch looks great. Exactly what I had in mind.
Thanks !

PR #304, opened 2019-11-15
Fix priority bits on G0 family

dotcypress:
@adamgreig could you review and merge this PR?


adamgreig:
What source are you getting 3 prio bits from? I'm not familiar with the G0 family especially, but RM0454 suggests 2 bits:

image

dotcypress:
@adamgreig yes, 2 bits, my bad. Fixed.

PR #314, opened 2019-11-29
Document the EXTI registers on every existing stm32

aurelj:


aurelj:
Just updated to also include the newly added STM32F730 device.

PR #316, opened 2019-12-06
Add STM32F730 device

rfuest:
This PR adds support for the STM32F730 and is based on the existing `stm32f7x2.yaml`. Some initial tests on real hardware were successful.


adamgreig:
oops, wrong button...

bors r+

PR #318, opened 2019-12-14
stm32h7: fix EXTI access for single core parts

richardeoin:
See RM0433 19.6.24 onwards


aurelj:
You forgot to commit the h7_exti_dualcore.yaml file in the PR. But anyway, it seems your next PR also contains the changes from this PR, so you should probably close this PR.

richardeoin:
@aurelj Thanks for spotting this. This intention is that this PR has the fix for single core (only). So I should have referenced `h7_exti_singlecore.yaml` instead. Fixing now

PR #319, opened 2019-12-14
stm32h7: more progress on dual core

richardeoin:
- Add *cm4 target. This has identical registers to the *cm7
  equivalent, but different interrupts
- Fix error in RCC GCR relating to C-M4 core
- Add dualcore parts of DBGMCU peripheral
- Add dualcore parts of EXTI peripheral
- Fix EXTI register access for dual clore

Still TODO

- Add dualcore parts of SYSCFG


devices/common_patches/h7_dbgmcu_dualcore.yaml
```diff
@@ -0,0 +1,170 @@
...
+        bitWidth: 1
+  APB4FZ1:
+    _add:
+      IWDG2:
+        description: Independent watchdog for D2 stop when Cortex-M7 in debug mode
+        bitOffset: 19
+        bitWidth: 1
+  APB1LFZ1:
+    _add:
+      DBG_WWDG2:
```
adamgreig: The other fields in this register are all named like `WWDG1` without the `DBG_` prefix, but I see that the RM says `DBG_WWDG2`. Which did you want here?

devices/common_patches/h7_dbgmcu_dualcore.yaml
```diff
@@ -0,0 +1,170 @@
...
+          bitWidth: 1
+        I2C2:
+          description: I2C2 SMBUS timeout stop when Cortex-M4 in debug mode
+          bitOffset: 22
+          bitWidth: 1
+        I2C1:
+          description: I2C1 SMBUS timeout stop when Cortex-M4 in debug mode
+          bitOffset: 21
+          bitWidth: 1
+        DBG_WWDG2:
```
adamgreig: As with previous comment, should this be `WWDG2` without `DBG_`?

devices/common_patches/h7_dbgmcu_dualcore.yaml
```diff
@@ -0,0 +1,170 @@
...
+          description: TIM1 stop when Cortex-M4 in debug mode
+          bitOffset: 0
+          bitWidth: 1
+    APB4FZ2:
+      description: APB4 peripheral freeze register CPU2
+      addressOffset: 0x058
+      access: read-write
+      resetValue: 0x00000000
+      fields:
+        LSDWDG2:
```
adamgreig: The RM calls this `WDGLSD2` instead of `LSDWDG2`

devices/common_patches/h7_dbgmcu_dualcore.yaml
```diff
@@ -0,0 +1,170 @@
...
+          description: LS watchdog for D2 stop when Cortex-M4 in debug mode
+          bitOffset: 19
+          bitWidth: 1
+        LSDWDG1:
```
adamgreig: The RM calls this `WDGLSD1` instead of `LSDWDG1`

devices/common_patches/h7_exti_dualcore.yaml
```diff
@@ -0,0 +1,853 @@
...
+          description: CPU2 rising trigger event configuration bit of Configurable Event input
+          bitOffset: 07
+          bitWidth: 1
+        MR8:
+          description: CPU2 rising trigger event configuration bit of Configurable Event input
+          bitOffset: 08
+          bitWidth: 1
+        MR9:
+          description: CPU2 rising trigger event configuration bit of Configurable Event input
+          bitOffset: 09
```
adamgreig: To avoid YAML confusion probably best to specify these numbers as just `9` instead of `09` etc (I believe by default this will just cause it to load as a string instead of an integer, rather than processing as octal, but that's still not ideal).

devices/common_patches/h7_exti_dualcore.yaml
```diff
@@ -0,0 +1,853 @@
...
+      access: read-write
+
+  _add:
+    C2IMR1:
+      description: EXTI interrupt mask register
+      addressOffset: 0x00C0
+      resetValue: 0xFFC00000
+      fields:
+        MR0:
+          description: CPU2 rising trigger event configuration bit of Configurable Event input
```
adamgreig: I think all these descriptions are wrong, RM says "CPU2 interrupt mask on [direct|configurable] event input x" (direct for 22-31, configurable for 0-21)

devices/common_patches/h7_exti_dualcore.yaml
```diff
@@ -0,0 +1,853 @@
...
+          description: CPU2 event mask on Event input x
+          bitOffset: 07
+          bitWidth: 1
+        MR8:
+          description: CPU2 event mask on Event input x
+          bitOffset: 08
+          bitWidth: 1
+        MR9:
+          description: CPU2 event mask on Event input x
+          bitOffset: 09
```
adamgreig: Same comment as above re using `9` instead of `09` etc

devices/common_patches/h7_exti_dualcore.yaml
```diff
@@ -0,0 +1,853 @@
...
+          description: CPU2 event mask on Event input x
+          bitOffset: 31
+          bitWidth: 1
+      C1PR1:
+      description: EXTI pending register
+      addressOffset: 0x00C8
+      resetValue: 0x00000000
+      fields:
+        PR0:
+          description: CPU2 event mask on Event input x
```
adamgreig: I think these descriptions are also wrong, RM says "Configurable event inputs X pending bit"

devices/common_patches/h7_exti_dualcore.yaml
```diff
@@ -0,0 +1,853 @@
...
+          bitOffset: 30
+          bitWidth: 1
+        MR31:
+          description: CPU2 event mask on Event input x
+          bitOffset: 31
+          bitWidth: 1
+      C1PR1:
+      description: EXTI pending register
+      addressOffset: 0x00C8
```
adamgreig: This is the address offset for C2PR1 but named C1PR1, guessing you meant to put the name as C2PR1?

devices/common_patches/h7_exti_dualcore.yaml
```diff
@@ -0,0 +1,853 @@
...
+        MR31:
+          description: CPU2 rising trigger event configuration bit of Configurable Event input
+          bitOffset: 31
+          bitWidth: 1
+      C1EMR1:
+        description: EXTI event mask register
+        addressOffset: 0x00C4
+        resetValue: 0x00000000
+        fields:
+        MR0:
```
adamgreig: Indentation is wrong here so I don't think it will do what you want...

devices/common_patches/h7_exti_dualcore.yaml
```diff
@@ -0,0 +1,853 @@
...
+          bitWidth: 1
+        MR30:
+          description: CPU2 event mask on Event input x
+          bitOffset: 30
+          bitWidth: 1
+        MR31:
+          description: CPU2 event mask on Event input x
+          bitOffset: 31
+          bitWidth: 1
+      C1PR1:
```
adamgreig: Indentation is wrong here, dedent C1PR1?

devices/common_patches/h7_exti_dualcore.yaml
```diff
@@ -0,0 +1,853 @@
...
+        PR21:
+          description: CPU2 event mask on Event input x
+          bitOffset: 21
+          bitWidth: 1
+    C2IMR2:
+      description: EXTI interrupt mask register
+      addressOffset: 0x00D0
+      resetValue: 0xFFF5FFFF
+      fields:
+        MR0:
```
adamgreig: RM calls this `MR32` instead of `MR0`, etc for the rest

devices/common_patches/h7_exti_dualcore.yaml
```diff
@@ -0,0 +1,853 @@
...
+          bitWidth: 1
+        MR30:
+          description: CPU2 interrupt Mask on Direct Event input x+32
+          bitOffset: 30
+          bitWidth: 1
+        MR31:
+          description: CPU2 interrupt Mask on Direct Event input x+32
+          bitOffset: 31
+          bitWidth: 1
+      C1EMR2:
```
adamgreig: Indentation wrong, dedent C1EMR2?

adamgreig: Guessing this should be C2EMR2

devices/common_patches/h7_exti_dualcore.yaml
```diff
@@ -0,0 +1,853 @@
...
+          bitWidth: 1
+        MR87:
+          description: CPU2 interrupt Mask on Direct Event input x+64
+          bitOffset: 23
+          bitWidth: 1
+        MR88:
+          description: CPU2 interrupt Mask on Direct Event input x+64
+          bitOffset: 24
+          bitWidth: 1
+      C1EMR3:
```
adamgreig: Wrong indentation, and should be C2EMR3?

devices/common_patches/h7_exti_dualcore.yaml
```diff
@@ -0,0 +1,853 @@
...
+          bitWidth: 1
+        MR87:
+          description: CPU2 event mask on Event input x+64
+          bitOffset: 23
+          bitWidth: 1
+        MR88:
+          description: CPU2 event mask on Event input x+64
+          bitOffset: 24
+          bitWidth: 1
+      C1PR3:
```
adamgreig: Wrong indentation, and should be C2PR3?

devices/common_patches/h7_dbgmcu_dualcore.yaml
richardeoin: Following the RM is the default, but useless prefixes are bad too. For minimal changes, I say strip the prefix

devices/common_patches/h7_dbgmcu_dualcore.yaml
richardeoin: Following the RM is the default, but useless prefixes are bad too. For minimal changes, I say strip the prefix

devices/common_patches/h7_dbgmcu_dualcore.yaml
richardeoin: Resolved to the RM `WDGLSD2`

devices/common_patches/h7_dbgmcu_dualcore.yaml
richardeoin: Resolved to the RM `WDGLSD1`

devices/common_patches/h7_exti_dualcore.yaml
richardeoin: Tidied all leading zeros, it was an error from the regex I used to write it all this

devices/common_patches/h7_exti_dualcore.yaml
richardeoin: Tidied all leading zeros, it was an error from the regex I used to write it all this

devices/common_patches/h7_exti_dualcore.yaml
richardeoin: Fixed

devices/common_patches/h7_exti_dualcore.yaml
richardeoin: Fixed

devices/common_patches/h7_exti_dualcore.yaml
richardeoin: Yes - there were a bunch of these mixed in with the indentation issues. All resolved to C2xxx

devices/common_patches/h7_exti_dualcore.yaml
richardeoin: Fixed, along with register names

devices/common_patches/h7_exti_dualcore.yaml
richardeoin: Fixed, along with register names

devices/common_patches/h7_exti_dualcore.yaml
richardeoin: Fixed, along with register names

richardeoin: Fixed, along with register names

devices/common_patches/h7_exti_dualcore.yaml
richardeoin: Fixed, along with register names

devices/common_patches/h7_exti_dualcore.yaml
richardeoin: Fixed, along with register names

devices/common_patches/h7_exti_dualcore.yaml
richardeoin: Resolved, only made the mistake here for C2xxx or C3xxx I believe

richardeoin:
> We already combine e.g. STM32F405 and STM32F415 in the stm32f405 module where the only difference is CRYP, and the same goes for most existing crypt/no-crypt families I believe

I would support combining too, with hindsight. ST does split the SVDs in this case, but the only difference really is the `HASH` and `CRYP` peripherals. We should combine H74xx and H75xx across the family at the same time. 

I'd suggest naming the combination H747 etc, even though H757 is the superset. Mostly to help new users find it, as the H747 parts are more popular and common - eg. the program-it-in-javascript thing Arduino are touting at CES

adamgreig:
> ST does split the SVDs in this case

Aren't we copying the same source file `stm32h7x5_cm{4,7}.svd` into two different files for cryp/no-cryp?

> I'd suggest naming the combination H747 etc, even though H757 is the superset. Mostly to help new users find it, as the H747 parts are more popular and common - eg. the program-it-in-javascript thing Arduino are touting at CES

This sounds good. Do you want to do it in this PR? It looks like recombining 747/757 should be ok?

richardeoin:
> This sounds good. Do you want to do it in this PR? It looks like recombining 747/757 should be ok?

Sure, combining the dual core parts here ⬇️

adamgreig:
Thanks, this all looks good and ready to merge. Please could you fix the merge conflict?

richardeoin:
Done! The conflict was from #312

adamgreig:
Thanks!

bors r+

PR #323, opened 2019-12-23
[RFC] rcc: Use largest integer instead of 0 when possible

Sh3Rm4n:
Often times, PLL values are used to calculate the clock frequency in HAL
implementation. See:

https://github.com/stm32-rs/stm32f1xx-hal/blob/bd54507245dcff38319a055e35ac70e22e1724ad/src/rcc.rs#L191-L211

or

https://github.com/stm32-rs/stm32f3xx-hal/blob/668bae491d13a4e31d1a309ad9c40b180247724a/src/rcc.rs#L252-L266

This is done, with subtracting the offset first, and then using the resulting PLL as the divisor.

Porting this, using the variants, introduced with version `0.9.0`, could look
like this:

https://github.com/Sh3Rm4n/stm32f3xx-hal/blob/717b54dd0a2b4787d85fd6d09cdffd3caea88f3b/src/rcc.rs#L293-L313

With this change, it could be written like this:

```rust
let hpre_div_value = u8::from(hpre_div) - 7;
```

And without having to track the offset, this would also be possible:

```rust
let hpre_div_value = u8::from(hpre_div) - u8::from(rcc::cfgr::HPRE_A::DIV1);
```

This makes the enumeration and conversion to `u8` more consistent, imo.

### DIFF
--- a/peripherals/rcc/rcc_common.yaml
+++ b/peripherals/rcc/rcc_common.yaml
@@ -19,13 +19,13 @@ RCC:
       "On": [1, "Clock On"]
   CFGR:
     PPRE*:
-      Div1: [0, "HCLK not divided"] # Same for [0, 8]
+      Div1: [3, "HCLK not divided"] # Same for [0, 8]
       Div2: [4, "HCLK divided by 2"]
       Div4: [5, "HCLK divided by 4"]
       Div8: [6, "HCLK divided by 8"]
       Div16: [7, "HCLK divided by 16"]
     HPRE:
-      Div1: [0, "SYSCLK not divided"] # Same for [0, 7]
+      Div1: [7, "SYSCLK not divided"] # Same for [0, 7]
       Div2: [8, "SYSCLK divided by 2"]
       Div4: [9, "SYSCLK divided by 4"]
       Div8: [10, "SYSCLK divided by 8"]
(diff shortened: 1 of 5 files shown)


adamgreig:
I see what you mean and it would make numerical manipulation of the field easier. It feels a bit weird to use the non-zero values for Div1 but the RM does promise it works OK.

I guess my main question is whether the HALs could really trust doing this? It seems like a dubious behaviour to rely on, compared to hardcoding the offset (and I guess also special-casing div1). We'd want to be sure we did this everywhere it's applicable, but there would still be some divider registers that don't have this behaviour or where the intermediate values are reserved, I'm sure, so those would still require special casing. What do you think?

Sh3Rm4n:
Yeah, it feels a bit hacky to rely on the numerical manipulation.

I find the incremental steps between different divisions more natural. I think
of it more like an offset. I hoped that `rcc::cfgr::HPRE_A::DIV1` could encode
the offset information for the division, so that special casing is not needed
for the `0`-case. But that's only my personal taste.

Either way, with or without using the variant enums to calculate the clock
registers, most HALs do use an offset subtraction.

What I've seen so far is that, even when intermediate values are reserved the
bit value is still consistent as the division factor could still be calculated
by subtracting the offset.

However, I'm fine leaving this as is and not merging this change. I agree the
HALs should not trust on this behavior, and when including this it could give a
false sense of trust.

It's just convenient for my use case.

adamgreig:
Thanks for the PR and the idea. On balance I think it's least surprising to keep using 0 for the fields where this applies, even if it means the HALs have to do a little more work, so I'm closing this issue.

Issue #327, opened 2020-01-02
Buggy SVD files for STM32F4 USB OTG peripherals


FLamparski:
I was trying to do some USB OTG programming. However it looks like not all relevant registers are exposed in this crate, or at least they are mislabelled, owing to some dodgy SVD files. For instance, `FS_DIEPTXF0` appears to be missing from the crate but it looks like it might appear under the name `FS_GNPTXFSIZ`. This might not be the only inconsistency, if I come across any more I'll report them here as well.

FLamparski:
`FS_DIEPTXF4` and `FS_DIEPTXF5` appear to be missing completely; only `FS_DIEPTXF{1,2,3}` are exposed and their doc comments are wrong:

```xml
          <name>FS_DIEPTXF1</name>
          <displayName>FS_DIEPTXF1</displayName>
          <description>OTG_FS device IN endpoint transmit FIFO size
          register (OTG_FS_DIEPTXF2)</description>
```

From the manual:

![image](https://user-images.githubusercontent.com/912935/71665841-3e119b00-2d56-11ea-90b3-d1411fa1ca6b.png)

(RM0390 pp 1143)

FLamparski:
`FS_GRXSTSP_Device` - OTG receive status read and pop register - is missing from the SVD. Only `FS_GRXSTSR_Device` is present, and that one does not pop the RX FIFO.

![image](https://user-images.githubusercontent.com/912935/71666597-7ff01080-2d59-11ea-9f2d-b3a1271230d1.png)

(RM0390 pp 1132)

Windfisch:
There are more bugs. I have fixed those which i found in https://github.com/stm32-rs/stm32-rs/pull/697, and there are probably even more.

Should we close this issue though?

dbristow-otc:
I'd like to request that we keep this issue open as a meta-issue for all the bugs in the SVD until we get device/host modes worked out. However, this will be an uphill battle as even the reference manuals seem to be wrong:

![image](https://github.com/stm32-rs/stm32-rs/assets/154448426/10c4d51f-c8c7-430b-bde4-5a73641419e1)
(Page 1291 REM0090 Rev 19) 

Yet ST's official HAL treats this register as writable:

![image](https://github.com/stm32-rs/stm32-rs/assets/154448426/e7241523-5de3-482e-98be-eff43c5f79ad)
[...]

PR #328, opened 2020-01-02
Apply USB OTG FS fixes on STM32F446

FLamparski:
Attempts to address some of #327 for STM32F446 by applying existing USB OTG patches. These result in a more correct SVD file being used, fixing the `DIEPTXF0`/`GNPTXFSIZ` misnomer and adding `GRXSTSP`.


Disasm:
STM32F446 requires another set of fixes. `otg_{fs|hs}_fixes.yaml` applies to F429 and similar chips with Core ID `0x0000_1200/0x0000_1100`. F446 chip has Core ID `0x0000_2000/0x0000_2100` and slightly different registers/fields.

thousandeyes-fwieland:
That might be the case but I've looked through the reference manual and the modifications from the fixes files do match the layout given there. Maybe the way forward would be to split up the `otg_{hs|fs}_fixes.yaml` by the range of parts they best apply to? Given that already there seem to be 3 different versions of that peripheral in the f4 range alone in this library.

Disasm:
I think it's better to make something like `otg_{hs|fs}_v2_fixes.yaml`, complete them, and then start thinking about moving common fixes into a separate file.

FLamparski:
Hmm, having had a look at the general breakdown of USB features in [AN4879](https://www.st.com/content/ccc/resource/technical/document/application_note/group0/0b/10/63/76/87/7a/47/4b/DM00296349/files/DM00296349.pdf/jcr:content/translations/en.DM00296349.pdf) I'd probably go with creating a fixes file for each column in Table 5, which is what I'm going to do for the F446 part in this PR

FLamparski:
image

I should probably fix the descriptions but basically the register-level `_derive` is working and some of the missing registers can be added back - the SVD is missing endpoints 4 and 5 in basically all endpoint registers in the FS peripheral, and it looks like some of the registers are given on the wrong address compared to the datasheet. I'll probably poke some more on this still.

FLamparski:
It looks like most of the memory map in the SVD is wrong for the 446, if the datasheet is taken as the truth 😢

Disasm:
Yes, if you compare original F429 and F446 SVDs, you can see they have almost identical USB peripherals, but this is not true. Reference Manual is more reliable source of information.

FLamparski:
Oh never mind - I didn't realise that the address offsets given in the SVD are relative to OTG_FS_DEVICE rather than OTG_FS as a whole, as is given in the datasheet. Grrr!

FLamparski:
@Disasm this should be it - I've added registers for the missing endpoints and it looks like the addresses do match the reference manual now. I don't think I'll be able to check whether this works with the synopsys-otg driver for a while though

adamgreig:
I've merged this manually to resolve the build conflicts, thanks for the PR! I appreciate there are some TODOs left; it seemed useful to get what you've got so far merged but please feel free to add more later.

PR #331, opened 2020-01-14
Replace scripts/svdpatch.py with svdtools

nickray:
So `svdtools `now has at least `esp32` and `lpc55-pacs` as users. I noticed two patches since I forked svdtools off `svdpatch.py` here, pending https://github.com/stm32-rs/svdtools/pull/4 I will do a v0.1.0 release this could semver-rely on.

These patches make a hopefully minimal intrusive switch to using `svdtools` here too (Python packaging being a bit of a mess compared to Rust...).

Thoughts?

In particular perhaps on: I'd like to add some regression tests in `svdtools`, perhaps patching all or a substantial subset of stm32-rs SVDs and comparing against a known previously good outputs.


nickray:
I think the build error is due to Travis not being on Python >=3.6?

jessebraham:
IIRC Travis currently defaults to Xenial (16.04), whose `python3` package is version **3.5.1**. Bionic (18.04) uses version **3.6.5**, so adding `dist: bionic` to the top of `.travis.yml` may help.

adamgreig:
> IIRC Travis currently defaults to Xenial (16.04), whose `python3` package is version **3.5.1**. Bionic (18.04) uses version **3.6.5**, so adding `dist: bionic` to the top of `.travis.yml` may help.

Ah, that could explain it. @nickray I believe if you change travis.yml in this PR it will run the build with the updated dist, so would you like to try updating the PR?

adamgreig:
Looks like that did it. I'll do a quick regression tonight/this weekend to check all generated SVDs are identical to master and then merge this. Thanks for your work on svdtools!

> I noticed two patches since I forked svdtools off svdpatch.py here, pending stm32-rs/svdtools#4 I will do a v0.1.0 release this could semver-rely on.

Are those patches already in the PyPI version of svdtools?

scripts/timer_hierarchy.py
```diff
@@ -4,23 +4,23 @@
 
 from collections import defaultdict
 
-import svdpatch
+import from svdtools import patch
```
jordens: delete that first `import `

nickray:
Nice! Yes, PyPI v0.1.0 should be equivalent to what you have.

adamgreig:
Sorry, didn't find time over the weekend but I'm still looking to merge this ASAP.

I noticed #328 also includes some changes to `svdpatch.py`. Any thoughts on that? We could merge this first and try to PR 328's changes to svdtools, or update svdtools first?

cc @FLamparski

nickray:
Oops, missed that one, sorry.. Can add to svdtools soonish.

nickray:
svdtools v0.1.1 is out with said patch, thanks to @jessebraham.

adamgreig:
Everything looks good, thank you for all the work separating this out!

bors r+

PR #336, opened 2020-02-12
Split SOF in DMAMUX.CSR. Closes #317.

### LINKED ISSUE #317: STM32H7 - dmamux1::CSR is lacking methods to access channel flags
Right now, to access a flag of a specific channel you have to do this
```rust
use stm32h7::stm32h743::dmamux1::CSR;

fn is_overrun_channel_3(csr: &CSR) -> bool {
    const CHANNEL_ID: u16 = 3;
    const MASK: u16 = 1 << CHANNEL_ID;

    csr.read().sof().bits() & MASK != 0
}
```
which isn't as straight forward as
```rust
fn is_overrun_channel_3(csr: &CSR) -> bool {
    csr.read().sof().sof3().bit_is_set()
}
```
So I would suggest to add methods for `sof0`, ..., `sof15`.

adamgreig:
@richardeoin, would you mind having a very quick look at this? #317 noticed that SOF is not split, and we already split a bunch of similar fields, so it seems OK.

### DIFF
--- a/devices/common_patches/h7_dmamux.yaml
+++ b/devices/common_patches/h7_dmamux.yaml
@@ -10,3 +10,5 @@
     _split: [OF]
   RGCFR:
     _split: [COF]
+  CSR:
+    _split: [SOF]


richardeoin:
Sure - This part of the DMAMUX documentation actually comes from https://github.com/stm32-rs/stm32-rs/commit/07980b801b26d2c4d3230d5901f39f6f761b3ac7  (I only moved it). But splitting `SOF` looks correct to me, since `CSOF` and others are already split.

PR #339, opened 2020-02-22
Fix DMA CR CHSEL bitWidth

aurelj:
Some F4 and F7 have a 3 bits CHSEL and others have a 4 bits CHSEL.
This provides separate patches for the different bitWidth.

### DIFF
--- a/devices/common_patches/dma/dma_v21.yaml
+++ b/devices/common_patches/dma/dma_v21.yaml
@@ -5,4 +5,4 @@ _include:
   S?CR:
     _modify:
       CHSEL:
-        bitWidth: 3
+        bitWidth: 4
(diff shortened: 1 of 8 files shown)


aurelj:
BTW, Travis is failing to build due to a bug in svdtools 0.1.3 (see https://github.com/stm32-rs/svdtools/issues/14). Installing svdtools 0.1.2 instead solves the issue.

aurelj:
> BTW, Travis is failing to build due to a bug in svdtools 0.1.3 (see [stm32-rs/svdtools#14](https://github.com/stm32-rs/svdtools/issues/14)). Installing svdtools 0.1.2 instead solves the issue.

Issue is now fixed in svdtools 0.1.4, so I've triggered Travis again, and it is now OK.

PR #347, opened 2020-03-05
sai: sai_v1 patch and sai peripheral apply to H7. Fix FSDEF

richardeoin:
SAI peripheral applies to the H7 also. Note that we introduce the very slightest difference with RM0433 (single core 743/753) by setting `[AB]CR1.NOMCK` -> `[AB]CR1.NODIV`. Whilst RM0433 uses `NOMCK`, both RM0399 (later dual core parts) and `sai/sai.py` use `NODIV`.
The description in `sai/sai.py` has text from both the old and new definitions, and it appears the functionality of the bit is unchanged.

[AB]FRCR.FSDEF is marked as read only in the SVD and the register diagram in the RM. However, in the prose text in RM0399 it is clearly noted as read-write "This bit is set and cleared by software."

### DIFF
--- a/devices/common_patches/h7_sai.yaml
+++ b/devices/common_patches/h7_sai.yaml
@@ -3,14 +3,29 @@
     _modify:
       MCKDIV:
         bitWidth: 6
+      NOMCK:
+        name: NODIV
     _add:
       MCKEN:
         description: Master clock generation enable
         bitOffset: 27
         bitWidth: 1
+  AFRCR:
+    _modify:
+      FSDEF:
+        access: read-write
   BCR1:
+    _modify:
+      MCKDIV:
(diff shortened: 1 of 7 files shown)


adamgreig:
Ugh, a pain that they renamed the bit. I'm a little hesitant to rename valid bits in the older devices just to match the existing peripheral file, but if this also makes life easier for HALs it might be worth doing. Any thoughts?

richardeoin:
Pragmatically the answer is to rename it for consistency, but it does break the rule of matching  the best interpretation of the RM. My opinion is the RM isn't invariant either (ST release new versions!), nothing will ever be perfect, practicality wins in a trivial case like this..

PR #348, opened 2020-03-06
enumerate values of rng cr register for F405, F407, F427, F429

MarcoIeni:
I extracted `rng-cr-ie` in a separate file, since it is common to `rng_v1` and the new `rng_v2`.


adamgreig:
Isn't `RNGEN` the same as well? If the only difference is `CED` it might make more sense to factor it as rng_v1 and rng_v1_ced, and include the latter where appropriate.

MarcoIeni:
The difference between the two `RNGEN` is in the data sheet description.
I will delete "True random number generator is disabled/enabled" and I will keep "Random number generator is disabled/enabled".

PR #353, opened 2020-03-14
Add RAMECC registers across the H7 family

richardeoin:
RAMECC was 'added' to the single core parts in RM0433 Rev 7. For the dual core parts there are definitions in the SVD, but they are so terrible it's worth starting over (there are registers with the wrong address, and the wrong fields)

For a use case, there is at least one set of commercial spaceflight hardware using the H7:
(no affiliation, no endorsement intended)

- [Nanoavionics OBC](https://nanoavionics.com/cubesat-components/cubesat-on-board-computer-main-bus-unit-satbus-3c2/)

### DIFF
--- a/devices/common_patches/h7_common_dualcore.yaml
+++ b/devices/common_patches/h7_common_dualcore.yaml
@@ -321,6 +321,7 @@ EXTI:
 
 _delete:
   - DMA2
+  - RAMECC?                         # RAMECC definitions are so terrible
 
 _add:
   DMA2:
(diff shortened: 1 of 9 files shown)


adamgreig:
Can we just define RAMECC1 and have 2 and 3 derive from that, rather than redefining the registers for each one?

richardeoin:
Yup, done. RAMECC3 only has two monitors connected, but I assume its safe to access the remaining monitor registers..

richardeoin:
On a real target (H747), RAMECC3.M[3-5]CR locations are not really registers (they don't read what was just written to them). I think the easiest solution to only keep RAMECC3.M[1-2]* is to keep the separate `RAMECC3` definition.

adamgreig:
bors retry

Issue #352, opened 2020-03-13
FPU inconsistencies


MarcoIeni:
I searched for "FPU" on all svd. These are the inconsistencies I've found.
Let me know if you confirm these inconsistencies or not.
Since this is a long issue, we could do different pull requests to solve it. I would check the related entries once they are solved.

## stm32l4x3

-  [ ]  peripheral USB has FPU interrupt

## stm32l4x2

-   [ ]  peripheral VREFBUF has FPU interrupt

## stm32l4s9, stm32l4s5, stm32l4r5, stm32l4s7, stm32l4r9, stm32l4r7

-   [ ]  peripheral FMC has the FPU interrupt, too
-   [ ]  peripheral FPU has duplicated interrupts

## stm32l4x2

-   [ ]  peripheral VREFBUF has the FPU interrupt
-   [ ]  peripheral FPU has duplicated interrupts

## stm32l4x1, stm32f469, stm32f446, stm32f412, stm32f410, stm32f303

-   [ ]  peripheral FPU has duplicated interrupts

## stm32h750x

-   [ ]  peripheral PF has FPU interrupt

## stm32f427, stm32f429, stm32f405, stm32f407

-   [ ]  peripheral RNG has FPU interrupts
-   [ ]  peripheral FPU has duplicated interrupts

## stm32f413

-   [ ]  peripheral SDIO has FPU (and RCC) interrupts
-   [ ]  peripheral ADC_Common has FPU interrupt
-   [ ]  peripheral FPU has duplicated interrupts

## stm32f411, stm32f401

-   [ ]  peripheral ADC_Common has FPU interrupt
-   [ ]  peripheral FPU has duplicated interrupts

# Note

The problem "peripheral FPU has duplicated interrupt" has been solved in
stm32f3x4.yaml with the following code:

```
FPU:
  _delete:
    _interrupts:
      - FPU
  _add:
    _interrupts:
      FPU:
        description: Floating point unit
        value: 81
```

adamgreig:
Thanks for the detailed checking! A lot of interrupts are definitely in the wrong place; some have been fixed but since in general it doesn't cause any actual problems they haven't all been fixed. I'm very happy to take PRs fixing the interrupts though.

You might find the `scripts/interrupts.py` script useful if you've not already found it.

PR #357, opened 2020-03-15
fix fpu double interrupt

MarcoIeni:
As detailed in issue #352, a lot of microcontrollers svd have a duplicated FPU interrupt entry.

This problem has already been fixed for stm32f3x4.
This commit apply this fix to all the other microcontrollers that need it.

### DIFF
--- a/devices/common_patches/fpu_interrupt.yaml
+++ b/devices/common_patches/fpu_interrupt.yaml
@@ -1,0 +1,11 @@
+# different microcontrollers have duplicated FPU interrupt
+
+FPU:
+  _delete:
+    _interrupts:
+      - FPU
+  _add:
+    _interrupts:
+      FPU:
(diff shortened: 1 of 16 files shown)


adamgreig:
Thanks, this looks good for the ones you've addressed. However, it looks like you're adding several new devices (the L4R/L4S ones) in this PR. They're probably worth having, but could those go in a new PR please?

MarcoIeni:
Should I close this pull request and open 2 new pull requests?

adamgreig:
If you add a new commit (or just edit the current commit) so this PR only has fixing the interrupts for existing MCUs, we can merge this one, and then make one new PR for the new devices.

MarcoIeni:
I edited the current commit, I hope that in this way the history will be cleaner

adamgreig:
bors retry

PR #361, opened 2020-03-29
fix fpu double interrupt for new devices

MarcoIeni:
As detailed in issue #352, a lot of microcontrollers svd have a duplicated FPU interrupt entry.

This problem has already been fixed for all the existing microcontrollers in #357.
This commit apply the same fix to all the **new** microcontrollers that need it.

### DIFF
--- a/devices/stm32l4r5.yaml
+++ b/devices/stm32l4r5.yaml
@@ -1,0 +1,4 @@
+_svd: ../svd/stm32l4r5.svd
+
+_include:
+ - common_patches/fpu_interrupt.yaml
--- a/devices/stm32l4r7.yaml
+++ b/devices/stm32l4r7.yaml
@@ -1,0 +1,4 @@
+_svd: ../svd/stm32l4r7.svd
+
+_include:
+ - common_patches/fpu_interrupt.yaml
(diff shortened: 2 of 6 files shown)


adamgreig:
Thanks for splitting the previous PR out and doing this one separately.

I think it's important to note that this PR is really creating six new devices, not just fixing the FPU interrupt on them -- they don't exist at all yet. Do you know anything about these parts? We currently have e.g. `stm32l4x5` which one might hope would cover `stm32l4r5` and `stm32l4s5`, but I see ST provide a separate SVD file. If it's possible to cover them using the `l4x5` SVD, that would be better.

If not, and we really do need to add six new devices to the L4 family, we can definitely do that - but this PR should be called "Add new STM32L4 devices", they'd probably want a number of extra patches and peripherals (perhaps starting on what we've already got in the L4x5 devices), and they'd need adding to `stm32_part_table.yaml`.

MarcoIeni:
> Do you know anything about these parts? 

No, I just noticed this double interrupts with a simple grep command on all the svd files.

> If it's possible to cover them using the l4x5 SVD, that would be better.

I ran `svd mmap` with  `stm32l4x5.svd` vs `stm32l4r5.svd` and `stm32l4x5.svd` vs `stm32l4s5.svd` and they look very different, so maybe it is better to use the svd files provided by ST. If we decide to go in this direction I will rename this PR and I will update the `stm32_part_table.yaml`.

> they'd probably want a number of extra patches and peripherals 

Ok, but I have to figure out those in this PR or they could be submitted in the future in another PR?

PR #364, opened 2020-04-05
h7: Apply flash bank grouping to all parts

richardeoin:
The goal here is to unify the naming of the common flash registers accross all h7 parts. Sadly this solution still leaves the register fields themselves named with a trailing 1, unlike the 747cm7/4 parts where we delete the flash peripheral and start over.


adamgreig:
You might be interested in the [undocumented `_strip_end` feature](https://github.com/stm32-rs/svdtools/blob/master/svdtools/patch.py#L406-L407) of svd patch for this.

richardeoin:
Indeed, that is very helpful

PR #368, opened 2020-04-22
Document flash for f3 devices

Sh3Rm4n:
- Added a common flash file, which is applicable to any stm32f3 device. 
- Split out register documentation, which are only applicable for some devices.
- Renaming `Flash` to `FLASH`, as it is done for different devices too


peripherals/flash/flash_f3_common.yaml
```diff
@@ -0,0 +1,108 @@
...
+    PRFTBS:
+      Disabled: [0, "Prefetch buffer is disabled"]
+      Enabled: [1, "Prefetch buffer is enabled"]
+    PRFTBE:
+      Disabled: [0, "Prefetch is disabled"]
+      Enabled: [1, "Prefetch is enabled"]
+    LATENCY:
+      WS0: [0, "0 wait states, if 0 < HCLK <= 24 MHz"]
+      WS1: [1, "1 wait state, if 24 < HCLK <= 48 MHz"]
+      WS1: [2, "1 wait states, if 48 < HCLK <= 72 MHz"]
```
adamgreig: Should be `WS2` and `2 wait states`

PR #373, opened 2020-04-28
Document rtc for stm32f3 devices

Sh3Rm4n:
This was pretty straight forward, as most of the registers were already documented for the stm32l0 devices. 

I split out the unique l0 registers, and renamed the `rtc_l0.yaml` to `rtc_common.yaml`. 
The moved registers were unchanged, except for one description typo `216` -> `2^16`. 

One patch in the `rtc_l0.yaml` peripheral description moved to the `common_patches` patch file. 

All fields of the stm32f3 devices should now be covered.


Sh3Rm4n:
Via `matchperipherals.py` I noticed, that the `f4` and `f7` devices almost match the `rtc_common` documentation. 

I had to add some missing flags to the `RTC CR` register of the f4 devices and added the `WCKSEL` -> `WUCKSEL` patch to the `f4` and `f7` devices. I confirmed that both of these patches apply to the reference manuals. 

The rest of the `rtc_common.yaml` documentation was added semi-automatic. I have not yet rechecked manually, if the documentation fits the rtc registers. A quick glance at the html output looks good though.

adamgreig:
Oh, and if it's possible to resolve the merge conflicts at the same time that would be great, should be straightforward I think.

Sh3Rm4n:
Resolved naming conventions and rebased onto master. 

These naming conventions affect [`rtc_l0`](https://github.com/stm32-rs/stm32-rs/pull/373/files#diff-c2440cc3c6e73ac76a79b8bfa2dc9b99L243) too, so I applied them there as well.

adamgreig:
Thanks! Sorry to be picky but could we use `Samples2` instead of `Sample2`? I know the existing UART stuff says `Sample1` but I think it probably should be plural, which is what we do for `Clocks` and `Cycles` and `Bits`.

Sh3Rm4n:
No worries :) I'll get to that later today!

PR #374, opened 2020-05-01
Add ACR, KEYR, OPTKEYR, and CR definitions for STM32F401 FLASH

yusefkarim:
Hi, I have added several FLASH register definitions, which I have tested on my STM32F401. This is my first time trying to add something to a PAC, please let me know if I have done anything incorrectly. I am also not sure if the peripheral .yaml file I created applies to all F4 boards, the name may need to be changed if not.

If this is accepted and there is value in adding all the registers definitions for FLASH, please let me know and I wouldn't mind reading the datasheet and adding them for everyone.

Thanks.


adamgreig:
Hi, thanks for the PR!

I think most of these registers are already documented by [flash_v1.yaml](https://github.com/stm32-rs/stm32-rs/blob/master/peripherals/flash/flash_v1.yaml) which is included in STM32F405 and F407 but apparently not in F401. Sorry that this wasn't easier to find! Maybe you could see if there's any difference between your new flash_f4 and the existing flash_v1 and then include flash_v1 on the F401?

yusefkarim:
Yikes, the file I made is almost identical to flash_v1. Thank you for taking the time to point that out. I will make sure to check similar devices when updating mine in the future!

I have pushed an update removing the flash_f4 file and including flash_v1 instead.

PR #375, opened 2020-05-02
Document syscfg for stm32f3 devices

Sh3Rm4n:
This should cover 100 % of the syscfg register from the stm32f3 devices. 

- I left quite a view TODOs regarding naming and doubled values, as I'm unsure how to go about this
- Some files could be split to smaller chunks, to reduce the duplication (however some devices slightly differ in their description strings and syscfg control register are sometimes quiet different between the devices)
- I split up the FPU_IE into it's five flags, as it seems, that they are independent of each other (correct me if I'm wrong about this)
- This PR is split into many small commits intentionally. This should make it easier to sqaush it into larger more reasonable commits after all review comments are addressed.

### DIFF
--- a/peripherals/syscfg/syscfg_f301.yaml
+++ b/peripherals/syscfg/syscfg_f301.yaml
@@ -1,0 +1,90 @@
+SYSCFG:
+  CFGR1:
+    I2C3_FMP:
+      Standard: [0, "FM+ mode is controlled by I2C_Pxx_FMP bits only"]
+      FMP: [1, "FM+ mode is enabled on all I2C3 pins selected through selection trhough IOPORT control registers AF selection bits"]
+    ENCODER_MODE:
+      NoRedirection: [0, "No redirection"]
+      MapTim2Tim15: [1, "TIM2 IC1 and TIM2 IC2 are connected to TIM15 IC1 and TIM15 IC2 respectively"]
+    TIM6_DAC1_DMA_RMP:
+      NotRemapped: [0, "TIM6_UP and DAC_CH1 DMA requests mapped on DMA2 channel 3"]
+      Remapped: [1, "TIM6_UP and DAC_CH1 DMA requests mapped on DMA1 channel 3"]
+    TIM1_ITR3_RMP:
+      NotRemapped: [0, "Not remapped"]
+      Remapped: [1, "TIM1_ITR3 = TIM17_OC"]
+    MEM_MODE:
+      MainFlash: [0, "Main Flash memory mapped at 0x0000_0000"]
+      SystemFlash: [1, "System Flash memory mapped at 0x0000_0000"]
+      MainFlash2: [2, "Main Flash memory mapped at 0x0000_0000"]
+      SRAM: [3, "Embedded SRAM mapped at 0x0000_0000"]
+  EXTICR1:
+    EXTI0:
+      PA0: [0, "Select PA0 as the source input for the EXTI0 external interrupt"]
+      PB0: [1, "Select PB0 as the source input for the EXTI0 external interrupt"]
+      PC0: [2, "Select PC0 as the source input for the EXTI0 external interrupt"]
+      PF0: [5, "Select PF0 as the source input for the EXTI0 external interrupt"]
+    EXTI1:
+      PA1: [0, "Select PA1 as the source input for the EXTI1 external interrupt"]
+      PB1: [1, "Select PB1 as the source input for the EXTI1 external interrupt"]
+      PC1: [2, "Select PC1 as the source input for the EXTI1 external interrupt"]
+      PF1: [5, "Select PF1 as the source input for the EXTI1 external interrupt"]
+    EXTI2:
+      PA2: [0, "Select PA2 as the source input for the EXTI2 external interrupt"]
+      PB2: [1, "Select PB2 as the source input for the EXTI2 external interrupt"]
+      PC2: [2, "Select PC2 as the source input for the EXTI2 external interrupt"]
+      PD2: [3, "Select PD2 as the source input for the EXTI2 external interrupt"]
+    EXTI3:
+      PA3: [0, "Select PA3 as the source input for the EXTI3 external interrupt"]
+      PB3: [1, "Select PB3 as the source input for the EXTI3 external interrupt"]
+      PC3: [2, "Select PC3 as the source input for the EXTI3 external interrupt"]
+  EXTICR2:
+    EXTI4:
+      PA4: [0, "Select PA4 as the source input for the EXTI4 external interrupt"]
+      PB4: [1, "Select PB4 as the source input for the EXTI4 external interrupt"]
+      PC4: [2, "Select PC4 as the source input for the EXTI4 external interrupt"]
+    EXTI5:
+      PA5: [0, "Select PA5 as the source input for the EXTI5 external interrupt"]
+      PB5: [1, "Select PB5 as the source input for the EXTI5 external interrupt"]
+      PC5: [2, "Select PC5 as the source input for the EXTI5 external interrupt"]
+    EXTI6:
+      PA6: [0, "Select PA6 as the source input for the EXTI6 external interrupt"]
+      PB6: [1, "Select PB6 as the source input for the EXTI6 external interrupt"]
+      PC6: [2, "Select PC6 as the source input for the EXTI6 external interrupt"]
+    EXTI7:
+      PA7: [0, "Select PA7 as the source input for the EXTI7 external interrupt"]
+      PB7: [1, "Select PB7 as the source input for the EXTI7 external interrupt"]
+      PC7: [2, "Select PC7 as the source input for the EXTI7 external interrupt"]
+  EXTICR3:
+    EXTI8:
+      PA8: [0, "Select PA8 as the source input for the EXTI8 external interrupt"]
+      PB8: [1, "Select PB8 as the source input for the EXTI8 external interrupt"]
+      PC8: [2, "Select PC8 as the source input for the EXTI8 external interrupt"]
+    EXTI9:
+      PA9: [0, "Select PA9 as the source input for the EXTI9 external interrupt"]
+      PB9: [1, "Select PB9 as the source input for the EXTI9 external interrupt"]
+      PC9: [2, "Select PC9 as the source input for the EXTI9 external interrupt"]
+    EXTI10:
+      PA10: [0, "Select PA10 as the source input for the EXTI10 external interrupt"]
+      PB10: [1, "Select PB10 as the source input for the EXTI10 external interrupt"]
+      PC10: [2, "Select PC10 as the source input for the EXTI10 external interrupt"]
+    EXTI11:
+      PA11: [0, "Select PA11 as the source input for the EXTI11 external interrupt"]
+      PB11: [1, "Select PB11 as the source input for the EXTI11 external interrupt"]
+      PC11: [2, "Select PC11 as the source input for the EXTI11 external interrupt"]
+  EXTICR4:
+    EXTI12:
+      PA12: [0, "Select PA12 as the source input for the EXTI12 external interrupt"]
+      PB12: [1, "Select PB12 as the source input for the EXTI12 external interrupt"]
+      PC12: [2, "Select PC12 as the source input for the EXTI12 external interrupt"]
+    EXTI13:
+      PA13: [0, "Select PA13 as the source input for the EXTI13 external interrupt"]
+      PB13: [1, "Select PB13 as the source input for the EXTI13 external interrupt"]
+      PC13: [2, "Select PC13 as the source input for the EXTI13 external interrupt"]
+    EXTI14:
+      PA14: [0, "Select PA14 as the source input for the EXTI14 external interrupt"]
+      PB14: [1, "Select PB14 as the source input for the EXTI14 external interrupt"]
+      PC14: [2, "Select PC14 as the source input for the EXTI14 external interrupt"]
+    EXTI15:
+      PA15: [0, "Select PA15 as the source input for the EXTI15 external interrupt"]
+      PB15: [1, "Select PB15 as the source input for the EXTI15 external interrupt"]
+      PC15: [2, "Select PC15 as the source input for the EXTI15 external interrupt"]
--- a/peripherals/syscfg/syscfg_f302.yaml
+++ b/peripherals/syscfg/syscfg_f302.yaml
@@ -11,10 +11,10 @@ SYSCFG:
     TIM6_DAC1_DMA_RMP:
       NotRemapped: [0, "TIM6_UP and DAC_CH1 DMA requests mapped on DMA2 channel 3"]
       Remapped: [1, "TIM6_UP and DAC_CH1 DMA requests mapped on DMA1 channel 3"]
-    ADC24_DMA_RMP:
+    ADC2_DMA_RMP:
       NotRemapped: [0, "ADC24 DMA requests mapped on DMA2 channels 1 and 2"]
       Remapped: [1, "ADC24 DMA requests mapped on DMA2 channels 3 and 4"]
-    DAC_TRIG_RMP:
+    DAC1_TRIG_RMP:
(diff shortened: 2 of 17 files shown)


peripherals/syscfg/syscfg_f302.yaml
```diff
@@ -0,0 +1,158 @@
+SYSCFG:
+  CFGR1:
+    I2C3_FMP:
+      Standard: [0, "FM+ mode is controlled by I2C_Pxx_FMP bits only"]
+      FMP: [1, "FM+ mode is enabled on all I2C3 pins selected through selection trhough IOPORT control registers AF selection bits"]
+    ENCODER_MODE:
+      NoRedirection: [0, "No redirection"]
+      # TODO: Naming?
```
adamgreig, line 8: `MapTim2Tim15` etc seems good enough to me.

peripherals/syscfg/syscfg_f302.yaml
```diff
@@ -0,0 +1,158 @@
...
+    ADC24_DMA_RMP:
+      NoRemap: [0, "ADC24 DMA requests mapped on DMA2 channels 1 and 2"]
+      Remap: [1, "ADC24 DMA requests mapped on DMA2 channels 3 and 4"]
+    DAC_TRIG_RMP:
+      NoRemap: [0, "DAC trigger is TIM8_TRGO in STM32F303xB/C and STM32F358xC devices"]
+      Remap: [1, "DAC trigger is TIM3_TRGO"]
+    TIM1_ITR3_RMP:
+      NoRemap: [0, "TIM1_ITR3 = TIM4_TRGO in STM32F303xB/C and STM32F358xC devices"]
+      Remap: [1, "TIM1_ITR3 = TIM17_OC"]
+    # TODO: Is documented as reserverd but appears in register table (RM0365 p.172) and register map
```
adamgreig, line 24: Yea, this looks like an error in the reference manual's longer explanation, I'd leave it in.

peripherals/syscfg/syscfg_f302.yaml
```diff
@@ -0,0 +1,158 @@
...
+    USB_IT_RMP:
+      NoRemap: [0, "USB_HP, USB_LP and USB_WAKEUP interrupts are mapped on interrupt lines 19, 20 and 42 respectively"]
+      Remap: [1, "USB_HP, USB_LP and USB_WAKEUP interrupts are mapped on interrupt lines 74, 75 and 76 respectively"]
+    MEM_MODE:
+      MainFlash: [0, "Main Flash memory mapped at 0x0000_0000"]
+      SystemFlash: [1, "System Flash memory mapped at 0x0000_0000"]
+      SRAM: [3, "Embedded SRAM mapped at 0x0000_0000"]
+      # TODO: How to go about these specialties
```
adamgreig, line 32: If you mean "how to handle FMC only being available on D/E parts": I'd just include it here with the note as you have. We don't have the granularity to distinguish those parts.

peripherals/syscfg/syscfg_f303_3x8.yaml
```diff
@@ -0,0 +1,276 @@
...
+    ADC2_DMA_RMP:
+      # TODO: Naming
+      MapDma2: [0, "ADC2 mapped on DMA2"]
+      MapDma1Ch2: [3, "ADC2 mapped on DMA1 channel 2"]
+      MapDma1Ch4: [4, "DC2 mapped on DMA1 channel 4"]
+    I2C1_TX_DMA_RMP:
+      MapDma1Ch6: [0, "I2C1_TX mapped on DMA1 CH6"]
+      MapDma1Ch2: [1, "I2C1_TX mapped on DMA1 CH2"]
+      MapDma1Ch4: [2, "I2C1_TX mapped on DMA1 CH4"]
+      # TODO: Name doubled value??
```
adamgreig, line 213: We can't have duplicate names, so depending on the situation we've either omitted the duplicate variant or named it something to distinguish it. For this I'd probably just omit it unless there's any reason to use one over the other?

adamgreig:
> I split up the FPU_IE into it's five flags, as it seems, that they are independent of each other (correct me if I'm wrong about this)

I agree, this seems like the sensible thing to do.

peripherals/syscfg/syscfg_f303_3x8.yaml
Sh3Rm4n: There is no reason to add the duplicated, so I do as you suggested.

peripherals/syscfg/syscfg_f302.yaml
Sh3Rm4n: Yes, that's what i meant.

peripherals/syscfg/syscfg_f301.yaml
```diff
@@ -0,0 +1,95 @@
...
+    ENCODER_MODE:
+      NoRedirection: [0, "No redirection"]
+      # TODO: Naming?
+      MapTim2Tim15: [1, "TIM2 IC1 and TIM2 IC2 are connected to TIM15 IC1 and TIM15 IC2 respectively"]
+    TIM6_DAC1_DMA_RMP:
+      NoRemap: [0, "TIM6_UP and DAC_CH1 DMA requests mapped on DMA2 channel 3"]
+      Remap: [1, "TIM6_UP and DAC_CH1 DMA requests mapped on DMA1 channel 3"]
+    TIM1_ITR3_RMP:
+      NoRemap: [0, "No remap"]
+      Remap: [1, "TIM1_ITR3 = TIM17_OC"]
```
Sh3Rm4n, line 14: What about the naming convention in this case, should'nt it technically be:

```suggestion
      NotRemapped: [0, "No remap"]
      Remapped: [1, "TIM1_ITR3 = TIM17_OC"]
```

Just trying to figure out the naming convention here. I don't really mind, which to prefer, as they both convey the message similarly, though in this case, the flag is used for configuration (user doing it actively) as opposed to reading out the current state of the hardware (in a general sense). But it is still a read-write flag.

peripherals/syscfg/syscfg_f303_3x8.yaml
```diff
@@ -0,0 +1,276 @@
...
+      # TODO: Naming?
+      MapTim2Tim15: [1, "TIM2 IC1 and TIM2 IC2 are connected to TIM15 IC1 and TIM15 IC2 respectively"]
+      MapTim3Tim15: [2, "TIM3 IC1 and TIM3 IC2 are connected to TIM15 IC1 and TIM15 IC2 respectively"]
+      MapTim4Tim15: [3, "TIM4 IC1 and TIM4 IC2 are connected to TIM15 IC1 and TIM15 IC2 respectively (STM32F303xB/C and STM32F358xC devices only)"]
+    DAC2_CH1_DMA_RMP:
+      # Note: In STM32F303x6/8 and STM32F328x8, this bit must be set
+      NoRemap: [0, "No remap"]
+      Remap: [1, "DAC2_CH1 DMA requests mapped on DMA1 channel 5"]
+    TIM7_DAC1_CH2_DMA_RMP:
+      # Note:    In STM32F303x6/8 and STM32F328x8, this bit must be set as there is no DMA2 in these products.
```
Sh3Rm4n, line 18: Any way, to get this information into the generated code / documentation. Or is it up to the HAL implementer to get this right. I feel like, this bit is too hidden for this important(?) information. Though this is not really a topic we have to address, it's more on the side of STMicroelectronics.

peripherals/syscfg/syscfg_f303_3x8.yaml
```diff
@@ -0,0 +1,276 @@
...
+      NoRemap: [0, "TIM7_UP and DAC_CH2 DMA requests mapped on DMA2 channel 4 in STM32F303xB/C and STM32F358xC devices"]
+      Remap: [1, "TIM7_UP and DAC_CH2 DMA requests mapped on DMA1 channel 4"]
+    # TODO: Check if this should be renamed or not. Conflicting with register map
+    TIM6_DAC1_CH1_DMA_RMP:
```
Sh3Rm4n, line 22: This is a case, where documentation name and the name in the register map are not the same.

As of RM0316 p. 261, this should be named:

```suggestion
    TIM6_DAC1_DMA_RMP:
```

I don't know which one to follow. When following the naming in the register map the patch [`devices/common_patches/f3_syscfg_cfgr1_timx_dacx_chx_dma_rmp.yaml`](https://github.com/stm32-rs/stm32-rs/pull/375/files#diff-d3ed5e32b8b8c96527fd5dc0597adb2b) wouldn't be necessary. 

As of now, I have a mixed case of applying patches to conform the documentation name (this example) but later still use names, that are used in the register map. IMO I should tend towards the register map names, as they also conform to the svd file and thus there is no need to patch all these names. But it would be harder to find the relevant documentation, if someone would search for it in the reference manual.

Sh3Rm4n:
> I haven't had time to check all the external interrupt remaps yet. It would be lovely if there was some way to combine them across different devices - perhaps they could be factored out into an exticr.yaml which the individual devices include, if possible?

I initially created a branch dedicated to this, as I too saw this as an obvious improvement. But I discarded  it. 

I really did not see a pattern. This is really dependent on the package and therefor on the available gpio pins. E.g. sometimes PFx might exist, sometimes not. I initially split out EXTIx A...D, because I thought every chip must have these. IIRC this was not the case, and the build process will happily accept every documentation I'll throw at it, as enumerated values are not checked against the svd file (as they do not exist initially). 

Also, I might have made mistakes, as these flags are documented in descending order in the reference manual (EXTI3, EXTI3, ..., EXT0, EXTI7, ..., EXTI4) and in the yaml file I've written them down as in ascending order.

peripherals/syscfg/syscfg_f301.yaml
adamgreig: Good point, yes they should be `NotRemapped`/`Remapped` please.

peripherals/syscfg/syscfg_f303_3x8.yaml
adamgreig: We don't have any way to provide this information really; it's up to the HAL writer or end user to read the RM and set this appropriately.

peripherals/syscfg/syscfg_f303_3x8.yaml
adamgreig: That is frustrating. I think it's not too important which one we follow:

* either way might cause confusion compared to the RM
* an extra patch is not a big overhead
* users (and HAL writers) will probably be checking the PAC documentation or autocompletion anyway

If you'd rather use the register map names that's fine with me.

peripherals/syscfg/syscfg_f37x.yaml
```diff
@@ -0,0 +1,144 @@
...
+      PD10: [3, "Select PD10 as the source input for the EXTI10 external interrupt"]
+      PE10: [4, "Select PE10 as the source input for the EXTI10 external interrupt"]
+      PF10: [5, "Select PF10 as the source input for the EXTI10 external interrupt"]
+    EXTI11:
+      PA11: [0, "Select PA11 as the source input for the EXTI11 external interrupt"]
+      PB11: [1, "Select PB11 as the source input for the EXTI11 external interrupt"]
+      PC11: [2, "Select PC11 as the source input for the EXTI11 external interrupt"]
+      PD11: [3, "Select PD11 as the source input for the EXTI11 external interrupt"]
+      PE11: [4, "Select PE11 as the source input for the EXTI11 external interrupt"]
+      PF11: [5, "Select PF11 as the source input for the EXTI11 external interrupt"]
```
adamgreig, line 106: No `PF11` on `EXTI11` (RM0313 9.1.4)

peripherals/syscfg/syscfg_f37x.yaml
```diff
@@ -0,0 +1,144 @@
...
+  EXTICR4:
+    EXTI12:
+      PA12: [0, "Select PA12 as the source input for the EXTI12 external interrupt"]
+      PB12: [1, "Select PB12 as the source input for the EXTI12 external interrupt"]
```
adamgreig, line 110: No `PB12` on `EXTI12` (RM0313 9.1.5)

peripherals/syscfg/syscfg_f37x.yaml
```diff
@@ -0,0 +1,144 @@
...
+      PC12: [2, "Select PC12 as the source input for the EXTI12 external interrupt"]
+      PD12: [3, "Select PD12 as the source input for the EXTI12 external interrupt"]
+      PE12: [4, "Select PE12 as the source input for the EXTI12 external interrupt"]
+    EXTI13:
+      PA13: [0, "Select PA13 as the source input for the EXTI13 external interrupt"]
+      PB13: [1, "Select PB13 as the source input for the EXTI13 external interrupt"]
```
adamgreig, line 116: No `PB13` on `EXTI13` (RM0313 9.1.5)

peripherals/syscfg/syscfg_f3x4.yaml
```diff
@@ -0,0 +1,187 @@
...
+      PC4: [2, "Select PC4 as the source input for the EXTI4 external interrupt"]
+      PD4: [3, "Select PD4 as the source input for the EXTI4 external interrupt"]
+      PE4: [4, "Select PE4 as the source input for the EXTI4 external interrupt"]
+      PF4: [5, "Select PF4 as the source input for the EXTI4 external interrupt"]
+    EXTI5:
+      PA5: [0, "Select PA5 as the source input for the EXTI5 external interrupt"]
+      PB5: [1, "Select PB5 as the source input for the EXTI5 external interrupt"]
+      PC5: [2, "Select PC5 as the source input for the EXTI5 external interrupt"]
+      PD5: [3, "Select PD5 as the source input for the EXTI5 external interrupt"]
+      PE5: [4, "Select PE4 as the source input for the EXTI4 external interrupt"]
```
adamgreig, line 77:
```suggestion
      PE5: [4, "Select PE5 as the source input for the EXTI5 external interrupt"]
```

peripherals/syscfg/syscfg_f3x4.yaml
```diff
@@ -0,0 +1,187 @@
...
+      PD7: [3, "Select PD7 as the source input for the EXTI7 external interrupt"]
+      PE7: [4, "Select PE7 as the source input for the EXTI7 external interrupt"]
+  EXTICR3:
+    EXTI8:
+      PA8: [0, "Select PA8 as the source input for the EXTI8 external interrupt"]
+      PB8: [1, "Select PB8 as the source input for the EXTI8 external interrupt"]
+      PC8: [2, "Select PC8 as the source input for the EXTI8 external interrupt"]
+      PD8: [3, "Select PD8 as the source input for the EXTI8 external interrupt"]
+      PE8: [4, "Select PE8 as the source input for the EXTI8 external interrupt"]
+      PF8: [5, "Select PF8 as the source input for the EXTI8 external interrupt"]
```
adamgreig, line 99: No `PF8` in `EXTI8` (RM0364 Rev 2 10.1.5)

Sh3Rm4n:
Thanks for the thorough review. Neat Idea with the script. Should've used something like this in the first place. :sweat_smile: 
I've gone through all the view notes and rechecked them against the reference manual again, just to be sure! :) 

I'll cleanup the commits and remove left TODO notes to make this PR mergable.

peripherals/syscfg/syscfg_f303_3x8.yaml
Sh3Rm4n: I'll keep it as is.  I prefer it, when a search in the reference manual matches the flag description rather than the register map.

adamgreig:
Thanks! Everything looks good to me now, let me know when it's ready to merge.

Sh3Rm4n:
I rebased to clean up the commit history. The only thing changed is [`c5eaa4b`](https://github.com/stm32-rs/stm32-rs/pull/375/commits/c5eaa4b8075acbe5d436eea36e1e801095c8ade7), which does apply naming-patches, as discussed in https://github.com/stm32-rs/stm32-rs/pull/375#discussion_r460446094, to different occasions. 

This is ready to merge, now! :)

PR #376, opened 2020-05-03
Document CRC

Sh3Rm4n:
Documenting advanced CRC for f3, f7,h7,l0 and l4. 

Added a patch to the value name of the CRC INIT register. With that most of the existing crc documentation where applicable for the `{h7,f7,l0,l4}` devices. Also the existing `crc_l0.yaml` was not necessary anymore.


peripherals/crc/crc_pol.yaml
```diff
@@ -0,0 +1,3 @@
+CRC:
+  POL:
+    POL: [0, 4294967295]
```
adamgreig, line 3:
```suggestion
    POL: [0, 0xFFFFFFFF]
```

PR #378, opened 2020-05-05
Add basic support for H7B3 "high memory density" line

richardeoin:
ST doesn't appear to have a coherent name for these parts, but "high memory density" seems to be a good description. The Reference Manual is RM0455

`h7_common_highmemory.yaml` has parts from both the existing `_singlecore.yaml` and `_dualcore.yaml` (and some special parts, of course)

`../peripherals/dma/dmamux_v1.yaml` applies, but throws an error.

### DIFF
--- a/svd/extract.sh
+++ b/svd/extract.sh
@@ -9,3 +9,5 @@ cp stm32h753x.svd stm32h753.svd
 cp stm32h753x.svd stm32h753v.svd
 cp stm32h7x5_cm4.svd stm32h747cm4.svd
 cp stm32h7x5_cm7.svd stm32h747cm7.svd
+cp stm32h7a3x.svd stm32h7a3.svd
+cp stm32h7b3x.svd stm32h7b3.svd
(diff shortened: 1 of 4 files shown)


svd/extract.sh
```diff
@@ -9,3 +9,5 @@ cp stm32h753x.svd stm32h753.svd
 cp stm32h753x.svd stm32h753v.svd
 cp stm32h7x5_cm4.svd stm32h747cm4.svd
 cp stm32h7x5_cm7.svd stm32h747cm7.svd
+cp stm32h7a3x.svd stm32h7a3.svd
```
adamgreig, line 12: Do we need to extract stm32h7a3.svd? Looks like only b3 is being used.

adamgreig:
Besides the query about the a3 svd file, everything else looks fine to me.

richardeoin:
Great! Agreed the a3 svd is unused, so can be removed.

PR #380, opened 2020-05-08
stm32f107: Ethernet and USB OTG support

torkeldanielsson:


torkeldanielsson:
*friendly nudge*? :)

There are some seemingly duplicate stuff in this PR: eth_mmc/mac/dma. But there are slight differences in the specification of ethernet registers in the f107 vs the other mcu:s with ethernet, so this is my attempt at how to fix that. Is there a better way to do this?

(I have an implementation of Ethernet support for the stm32f107, and it depends on these changes before I can make a PR of that in the other repos (eth and 1xx).)

adamgreig:
Hi! Thanks for the nudge, sorry I haven't got to this sooner. I'll try and get it reviewed in the next few days, it's certainly next in the queue.

> Is there a better way to do this?

In other peripherals, where things are mostly the same we have a _common.yaml containing the parts that are the same, and then e.g. _f4.yaml and _f107.yaml that include the _common.yaml and extend it with the part-specific features. It depends how much is similar or different compared to the existing ones. If it's only a few fields then it's a shame to duplicate everything else, but if lots of fields are different it might be simplest to just have totally separate files.

torkeldanielsson:
No worries! To let you know: I am working on de-duplicating the eth_xxx.yml things now into common and addition now. The 107 have a few less registers than the other ones, so it seems like it will work.

torkeldanielsson:
Ok, I think this ended up better.

PR #381, opened 2020-05-09
Use wildcard in bors.toml instead of specifying each build

therealprof:
According to the bors documentation SQL-like wildcards are allowed

### DIFF
--- a/.github/bors.toml
+++ b/.github/bors.toml
@@ -1,6 +1,5 @@
 required_approvals = 1
 block_labels = ["wip"]
 delete_merged_branches = true
-status = [
-    "Check (%)",
-]
+status = ["Check (%)"]
+status_wait_success = ["Check (%)"]
--- a/.github/bors.toml
+++ b/.github/bors.toml
@@ -2,17 +2,5 @@ required_approvals = 1
 block_labels = ["wip"]
 delete_merged_branches = true
 status = [
-    "Check (stm32f0)",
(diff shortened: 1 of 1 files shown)


adamgreig:
Aah, this would be much better.

bors try

adamgreig:
Looks like the whole build now succeeds after any one success, without knowing it has to wait for the rest of them...

therealprof:
Weird.

therealprof:
Hm, there's an RFC for a new variable to wait until those checks have been completed: https://bors.tech/rfcs/0360-add-a-new-configuration-option-status-wait-success.html

Not sure whether this is available already, though.

therealprof:
bors try

adamgreig:
bors try

adamgreig:
It looks like that bors RFC addresses a slightly different need, when there are checks that start out failing and then turn green within the timeout period. For this GHA case, the extra checks are simply not reported to Travis at all, it doesn't know they will occur until they complete. We don't want it to always wait 1h to see what might appear, so I think we just have to enumerate all the checks.

therealprof:
The way I'm reading it the "status" is used for quick failing and the "status_wait_success" for the unblocking.

That's super-weird; why would you have wildcard support if it means "any" of the matches instead of "all" of the matches?

therealprof:
Anyway, let's close this then.

Issue #386, opened 2020-06-07
Detailed alternate function names for GPIOs


Past9:
Is there any interest in having enumerated values for GPIO alt functions that describe the actual function? For example, instead of setting AFRL/AFRH fields to AF1-AF16, values like TIM2_CH1 could be used instead. This info is available in the datasheets, like this one starting on page 45: https://www.st.com/resource/en/datasheet/stm32f303cb.pdf

I'm not sure this isn't already being done, but I don't see any examples of it. Thought I might make a PR for a couple chips I'm using if there's interest.

salkinium:
Alternate Function Mapping can be extracted from CubeMX data. See https://github.com/modm-io/modm-devices for a shameless plug ;-P

Past9:
Wow, thank you. It looks like the CubeMX data may be incomplete, given that there are some alt funcs missing from those files. For example all but one of the port D alt funcs is missing from the file for the device I linked above: https://github.com/modm-io/modm-devices/blob/develop/devices/stm32/stm32f3-03-6_8.xml

Still, this could save a big chunk of a pretty tedious manual task. Will definitely try to make use of it.

salkinium:
The CubeMX data also has some conditions encoded in them, that are resolved with their Java engine. I didn't manage to reverse-engineer that at all, so there's probably more missing. It would be nice to be able to parse the PDFs if only for comparison. I'm flying pretty blind with only manual spot checks here and there.

Past9:
I've noticed that some chips in the same family have alt funcs that aren't implemented in others. Like the F303xD/E (https://www.st.com/resource/en/datasheet/stm32f303zd.pdf) have an AF2 for PA9, but the F303xB/C chips don't. So the conditions could be related to that.

Unfortunately I think you'd need OCR software to get anything meaningful from the PDFs. I'm not very familar with that format but I believe the tables aren't actually in a tabular format in the file. It's more like "draw a square here with this text in it, another square there with that text in it, etc." I guess that's what makes it "portable". No ambiguity in how things should be rendered.

adamgreig:
The CubeMX data is also available here: https://github.com/stm32-rs/cube-MX-db

I think it would be a _lot_ of work to get a meaningful amount of AF entries encoded compared to what we do right now, but at least in principle it would be nice to encode both what the AF does and also not emit a variant for disconnected AFs.

[...]

PR #388, opened 2020-06-09
Add detailed alternate function names to STM32F303

Past9:
This PR moves the generic alt function definition from gpio_v2_common.yaml to gpio_v2_common_altfuncs.yaml. Chips that don't have detailed alt function definitions can import this new file, while those that do can instead import the file that defines them. In this case the F303 series imports gpio_v2_f303.yaml.


Past9:
Related: https://github.com/stm32-rs/stm32-rs/issues/386

adamgreig:
Thanks for trying this out!

Your PR highlighted something I didn't realise in #386 which is that you of course have to copy the peripherals instead of having them be `derivedFrom`, which has the kind of annoying side effect of meaning they're no longer the same type, so there's much more code generated which will impact crate sizes and compile times.

Really this is a problem that needs to be solved better somewhere else, because already some GPIOs are not identical to others even though they should be (due to SVD file weirdness), but also because some GPIOs are not identical to others in hardware (e.g. not having an LCKR register). It sucks to not have the GPIO peripherals all be the same type as it complicates writing generic code and increases build times a lot. Not sure what the best answer is here really. I'll try and do some measurements with this PR to see what the actual impact is.

Past9:
I see your point about the code size. I hadn't thought about that either. Maybe this is better addressed in the HAL crates.

Anyway, if you decide this approach will work, let me know and I can knock out the F407 as well. If not, no worries. Thanks for looking into it.

PR #390, opened 2020-06-14
STM32F4: rename, fix and document FMPI2C peripheral

aurelj:


adamgreig (approved):
Sorry this took so long to review, and thanks for the big PR!

I'm a little wary of renaming some RCC fields to no longer match the RM but in this case it seems to give the best consistency so let's go with it.

bors merge

PR #383, opened 2020-05-21
Update stm32f4 SDIO register fields and values

jkristell:
Documentation and limits for the sdio register values.

PR #393, opened 2020-06-30
STM32F446 SDIO

cyberillithid:
For some reason SVD file is significantly different from what the reference manual says. 

Since some of the fields are presumably common across different F4 implementations of this peripheral, I've tried to DRY the descriptions where applicable.

### DIFF
--- a/peripherals/sdio/sdio_f446.yaml
+++ b/peripherals/sdio/sdio_f446.yaml
@@ -1,0 +1,258 @@
+_modify:
+  SDMMC:   # renaming block to be RM0390-compliant
+    name: SDIO
+    description: 'Secure digital input/output SD/SDIO MMC card host interface'
+    groupName: SDIO
+    addressBlock:
+      size: "0x400"     
+
+SDIO:
+  _strip: SDMMC_
+  _delete:
+    - '*ACKTIMER'
+    - '*IDMA*'
+    - '*VER'
+    - '*ID'
+  _modify:
+    SDMMC_POWER:
+      description: 'SDIO power control register'
+    SDMMC_CLKCR:
+      description: 'CLKCR register controls the SDIO_CK output clock.'
+    SDMMC_CMDR:
+      name: "SDMMC_CMD"  # NOTE: something weird happens on leaving just 'CMD' -- 'MD' -- ?
+    SDMMC_RESP4R:
+      addressOffset: "0x20"
+      name: RESP4
+    SDMMC_RESP3R:
+      addressOffset: "0x1C"
+      name: RESP3
+    SDMMC_RESP2R:
+      addressOffset: "0x18"
+      name: RESP2
+    SDMMC_RESP1R:
+      addressOffset: "0x14"
+      name: RESP1
+    SDMMC_RESPCMD:
+      addressOffset: "0x10"
+    SDMMC_ARGR:
+      name: ARG
+    SDMMC_RESPCMDR:
+      name: RESPCMD
+    SDMMC_DLENR:
+      name: "SDMMC_DLEN"
+    SDMMC_DCNTR:
+      name: 'SDMMC_DCOUNT'
+    SDMMC_STAR:
+      name: SDMMC_STA
+    SDMMC_MASKR:
+      name: SDMMC_MASK
+    SDMMC_FIFOR:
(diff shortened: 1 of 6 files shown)


peripherals/sdio/sdio_f446.yaml
```diff
@@ -0,0 +1,258 @@
...
+    SDMMC_DLENR:
+      name: "SDMMC_DLEN"
+    SDMMC_DCNTR:
+      name: 'SDMMC_DCOUNT'
+    SDMMC_STAR:
+      name: SDMMC_STA
+    SDMMC_MASKR:
+      name: SDMMC_MASK
+    SDMMC_FIFOR:
+      addressOffset: "0x80" # acc. to CMSIS headers (??)
```
adamgreig, line 50: 0x80 seems correct in RM0390 too, where SDIO_FIFO is listed at address offset 0x80

peripherals/sdio/sdio_f446.yaml
```diff
@@ -0,0 +1,258 @@
...
+      name: SDMMC_FIFO
+  # What is the exact logic of [svdpatch.py]? Why is stripping AFTER delete & modify?
```
adamgreig, line 52: That's just the order of operations in patch.py: it deletes, then modifies, then strips. Not saying it's necessarily the most useful or clear thing but that's the order it's written in.

peripherals/sdio/sdio_f446.yaml
```diff
@@ -0,0 +1,258 @@
...
+    - '*IDMA*'
+    - '*VER'
+    - '*ID'
+  _modify:
+    SDMMC_POWER:
+      description: 'SDIO power control register'
+    SDMMC_CLKCR:
+      description: 'CLKCR register controls the SDIO_CK output clock.'
+    SDMMC_CMDR:
+      name: "SDMMC_CMD"  # NOTE: something weird happens on leaving just 'CMD' -- 'MD' -- ?
```
adamgreig, line 22: Not sure what you mean here exactly.. is something weird still happening with the patch as-is?

peripherals/sdio/sdio_f446.yaml
```diff
@@ -0,0 +1,258 @@
...
+#          bitOffset: 0
+#          bitWidth: 32
+#          access: read-write
+#  _array:
+#    FIFOARR*: {}
+#    "RESP?R":
+#      name: "RESP%s"
+#      _modify:
+#        CARDSTATUS1:
+#          name: CARDSTATUS
```
adamgreig, line 87: How come these are all commented out?

adamgreig (commented):
Sorry it's taken so long to review this, and thanks for the PR! It's wild how different the SVD was to the reference manual, I wonder what was going on there.

I've left a few comments where you had some outstanding comments in the files which might be worth discussing or removing from the yaml files before merging. Otherwise this change looks OK to me.

peripherals/sdio/sdio_f446.yaml
cyberillithid: I'm not sure what exactly am I doing wrong (or is this just as designed, and I've missed that part of svdtools documentation), but the `_strip: SDMMC_` directive strips the first letter of any register starting with C; e.g., if this line 22 is `name: "CMD"`, the register name becomes "MD" (and the patch fails).

peripherals/sdio/sdio_f446.yaml
cyberillithid: The CARDSTATUS bits were already implemented in sdio_f4.yaml; and I'm not sure how to implement FIFO array and whether it's even needed, since most of C code I've seen sticked with just the first 32-bit word (at 0x80).

cyberillithid:
Thanks for the review! I've fiddled with this F446 fix a while ago, before I decided to attempt to pullrequest these changes, so some of the lines you've questioned are just ones I've missed while refactoring in changes from #383 pull request.

(Moreover, this commit fixes another oversight of mine: the 'RESPCMD' offset did not affect 'RESPCMDR')

adamgreig (approved):
Thanks again for the PR! I've moved the sdio_f446 file to common_patches as that's where we put yaml files that just fix the SVDs rather than add enum variants, and I noticed the bitWidth for DCOUNT was off by one.

bors merge

Issue #394, opened 2020-07-02
SVD files for STM32G0 series are incorrect


ijager:
The G0 series includes 2 lines and a total of only 6 devices:

- value line `stm32g0x0`: G030 , G070
- access line `stm32g0x1`: G031 , G041, G071 , G081

Each of these parts have different sets of peripherals. However that is not reflected in the current set of SVD files

```bash
stm32-rs/svd$ ls -hl stm32g0*.svd                                                                                                                                                                                                                                                  master
742K Aug 29  2019 stm32g030.svd
723K Aug 29  2019 stm32g031.svd
742K Aug 29  2019 stm32g041.svd
897K Aug 29  2019 stm32g07x.svd
897K Aug 29  2019 stm32g081.svd
```

Also see https://github.com/stm32-rs/stm32g0xx-hal/issues/18

I cannot find SVD files for these parts on [st.com](https://www.st.com/en/microcontrollers-microprocessors/stm32g0-series.html#overview). Does anyone have access to them if they exist?

ijager:
This was fixed with the svd yaml patches for these parts.

PR #395, opened 2020-07-03
Add stm32g070 and stm32g071 to replace stm32g07x

ijager:
The main goal of this PR is to create separate crates for `stm32g070` and `stm32g071` since the `g070` belongs to the value line that is missing a lot of features compared to the 'access line'. 

I extracted the latest SVD files from STM32CubeIDE. They're a bit better than previous version, still quite some bugs in the files for parts other than G081 (since it is the part with all features). However the bugs are mostly peripherals that actually don't exist on that specific part. I manually edited the SVD files for G071 and G070 to remove some of those. 

This addresses https://github.com/stm32-rs/stm32-rs/issues/394 and is related to https://github.com/stm32-rs/stm32g0xx-hal/issues/18


adamgreig:
Thanks for the update! I think it does make sense to split the 70/71 apart.

> I manually edited the SVD files for G071 and G070 to remove some of those.

We only put completely stock unmodified SVDs in the zip files; ideally these zip files should be directly downloaded from ST and not modified at all. Would it be possible to redo your modifications as patches in the device YAML files or in `common_patches/` please?

ijager:
Ok I see. It is gonna take way more time then to do. I can revert to the stock files, but the stock ST version literally has typos:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: In peripheral `ADC`

Caused by:
    0: In register `HWCFGR6`
    1: 0x1F1F1F11F invalid
    2: number too large to fit in target type', /Users/ingmar/.cargo/registry/src/github.com-1ecc6299db9ec823/svd2rust-0.17.0/src/main.rs:96:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
make: *** [stm32g0/src/stm32g070/mod.rs] Error 101
```

So I fix that now with a patch and then in the next revision ST also fixes it and then we undo the patch again? Or can svdtools/svd2rust ignore those cases then?

ijager:
Now it contains the stock SVD files v1.3 with a patch to make it build. The other changes I might add back later with patches, it is a different issue anyway.

ijager:
@adamgreig, anything else I should change?

PR #396, opened 2020-07-03
Fix Makefile and gitignore for MacOS

ijager:


.gitignore
```diff
@@ -10,3 +10,6 @@ html/
 
 # automatically generated crates
 /stm32*/
+
+# MacOS
+.DS_Store
```
adamgreig, line 15: I don't think platform- and tool-specific ignores should go in repository `.gitignores`; otherwise we should start adding ignores for vim temporary files, Windows thumbnail databases, MacOS's `._`, and so on. Would it work for you to put `.DS_Store` in your global git ignore (`.config/git/ignore`)?

Makefile
```diff
@@ -52,7 +52,7 @@ $(1)/src/%/mod.rs: svd/%.svd.patched $(1)/Cargo.toml
 	rustfmt --config-path="rustfmt.toml" $$(@D)/lib.rs
 	sed "1,20d;23,28d" $$(@D)/lib.rs > $$@
 	rm $$(@D)/build.rs $$(@D)/lib.rs
-	mv -f -t $$(@D)/.. $$(@D)/generic.rs
+	mv -f $$(@D)/generic.rs $$(@D)/..
```
adamgreig, line 55:
```suggestion
	mv -f $$(@D)/generic.rs $$(@D)/../
```

.gitignore
ijager: yes sure, makes sense

PR #398, opened 2020-07-09
Fixes for G070, G071, G081

ijager:
- removal of peripherals that are not present on G070 and G071
- fix register offset bug in EXTI

### DIFF
--- a/devices/stm32g070.yaml
+++ b/devices/stm32g070.yaml
@@ -20,8 +20,11 @@ _delete:
   - SYSCFG_VREFBUF
   - RNG
   - HDMI_CEC
+# deleted so we can re-add them without interrupts
+  - TIM7
+  - USART3
 
-# Remove timer 3 (above) and rename TIM2 to TIM3. Because TIM3 was derive from TIM2
+# Remove TIM3 (above) and rename TIM2 to TIM3. Because TIM3 was derive from TIM2
 _modify:
   TIM2:
     name: TIM3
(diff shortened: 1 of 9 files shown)


ijager:
I have just succesfully tested this PR for each part/feature with the hal (https://github.com/stm32-rs/stm32g0xx-hal/pull/23) and in a real firmware project

devices/stm32g070.yaml
```diff
@@ -8,6 +8,126 @@ DMA:
...
+        description: TIM7 global interrupt
+
+# TODO: crashes svdtools.patch.MissingPeripheralError: Could not find USART3.
+# maybe bug in svdtools v1.6
+# USART3:
+#   _modify:
+#     _interrupts:
+#       USART3_USART4_LPUART1:
+#         name: USART3_USART4
+#         description: USART3 + USART4 interrupt
```
andresv, line 127: I guess this needs to be addressed before merging.

ijager: I believe this is related to https://github.com/stm32-rs/svdtools/pull/13 which was reverted. If a peripheral is `derivedFrom` another one, then the patching tool cannot seem to match on the peripheral name. Maybe there is a workaround but I couldn't figure it out yet.

```xml
<peripheral derivedFrom="USART1">
    <name>USART3</name>
    <baseAddress>0x40004800</baseAddress>
    <interrupt>
      <name>USART3_USART4_LPUART1</name>
      <description>USART3 + USART4 + LPUART1</description>
      <value>29</value>
    </interrupt>
  </peripheral>
```

So the result is that the STM32G070 crate will still have `USART3` erroneously available through its PAC. It is not a big deal I think. When svdtools supports `derivedFrom` we can still remove it.

andresv: @adamgreig do you know if it is possible to do it differently or should we just let it be as it is?

ijager: Wait this was only about the interrupt _name_. Which should be `USART3_USART4` instead of `USART3_USART4_LPUART1`since there is no `LPUART1` peripheral on this part.

So the issue here is: "How to modify properties of a _`derivedFrom` peripheral_ "

adamgreig: Interrupts don't actually need to be associated with the correct peripheral so in this case it might be simplest to just make sure the base peripheral has all relevant interrupts and the derived peripheral has none, which should be doable with current svdtools without needing to modify the derived peripheral.

ijager: Removing them and re-adding seems to work!

PR #400, opened 2020-07-15
Do not add writeConstraints to DMA address registers

adamgreig:
Currently we mark these fields with a writeConstraint of any 32-bit value, which means svd2rust generates a safe `bits()` method to write to them. In turn, this allows entirely safe code to use the DMA to read or write any address, which can cause memory unsafety.

Ideally perhaps svd2rust shouldn't act like this, but by removing these writeConstraints we make the relevant `bits()` methods unsafe, so unsafe code is required to use the DMA engine. A HAL with a DMA driver can implement a safe abstraction using the upcoming DMA traits to ensure the addresses being written do not lead to unsafety.

What do people think about this change? Especially it might impact @thalesfragoso's DMA and stm32-eth work.

### DIFF
--- a/peripherals/dma/dma_v1.yaml
+++ b/peripherals/dma/dma_v1.yaml
@@ -64,7 +64,3 @@
       Enabled: [1, "Channel enabled"]
   "CNDTR*":
     NDT: [0, 65535]
-  "CPAR*":
-    PA: [0, 4294967295]
-  "CMAR*":
-    MA: [0, 4294967295]
--- a/peripherals/dma/dma_v3.yaml
+++ b/peripherals/dma/dma_v3.yaml
@@ -82,10 +82,6 @@
       Enabled: [1, "Stream enabled"]
   "S?NDTR":
     NDT: [0, 65535]
-  "S?PAR":
-    PA: [0, 4294967295]
-  "S?M?AR":
-    "M?A": [0, 4294967295]
   "S?FCR":
     FEIE:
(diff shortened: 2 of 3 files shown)


jordens:
I support this change. Handing an address (which can be obtained safely) to the DMA peripheral is sufficiently equivalent in unsafety as dereferencing such an address.

It's unfortunately going to be far from complete: merely starting the DMA peripheral with the default reset value (0?) in the address register is just as unsafe. In general the memory safety of peripherals is not easily determined by their register settings, contrary what a safe svd2rust api might suggest.

therealprof:
I agree with the reasoning and support the change.

adamgreig:
> It's unfortunately going to be far from complete: merely starting the DMA peripheral with the default reset value (0?) in the address register is just as unsafe. In general the memory safety of peripherals is not easily determined by their register settings, contrary what a safe svd2rust api might suggest.

At least by default 0x0 is mapped to flash instead of RAM, but yes, it's definitely not perfect. I am more and more convinced that svd2rust should not be exporting safe APIs at all, but that's probably a discussion for another time and place. Anyway I think let's get this merged. I'm hoping to finish reviewing the other outstanding PRs and get a new release in fairly soon too.

therealprof:
> I am more and more convinced that svd2rust should not be exporting safe APIs at all, but that's probably a discussion for another time and place.

I agree. But most likely for different reasons. ;)

PR #401, opened 2020-07-22
h7: Rename USBxOTGHSEN -> USBxOTGEN, USBxOTGULPIHSEN -> USBxULPIEN

richardeoin:
The effect of this PR is different between the H7 groupings:

**highmemory (RM0455):** This change brings us in line with the RM, fixing a mistake on our part

**singlecore/dualcore (RM0433/RM0399):** Breaks with the RM, instead preferring a more consistent scheme that matches RM0455. All other peripherals enable/resets are in the form name{en,rst} , so having the inconsistent name{hsen,rst} is a pain.

I'm not sure the existance of USB2ULPIEN makes any sense (the USB2 peripheral has no ULPI), but it is RM0433/RM0399.

### DIFF
--- a/devices/common_patches/h7_common_dualcore.yaml
+++ b/devices/common_patches/h7_common_dualcore.yaml
@@ -558,15 +558,6 @@ RCC:
         description: ART Clock Enable
         bitOffset: 14
         bitWidth: 1
-    _modify:
-      USB1OTGEN:
-        name: USB1OTGHSEN
-      USB2OTGEN:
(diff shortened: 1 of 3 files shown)


richardeoin:
Yes, we're changing it to match `USB2OTGRST` and RM0455. It seems an unlikely typo, but I doubt there's anything to be gained in second-guessing what that actual reason was.

adamgreig:
bors merge

PR #409, opened 2020-08-06
Add corrections to STM32G474's TIM1

timblakely:
Noticed some more errors in the SVD for the STM32G474's TIM1. These also likely apply to TIM8 and TIM20 too, but I haven't had a chance to check those for error yet.

Let me know if there's any style guides specific to this repo and/or if you'd like me to move the file around to a different location.


adamgreig:
Thanks for the PR! Generally the style seems OK to me. I agree that it probably applies exactly to TIM8 and TIM20, since in the SVD you'll find both of those are marked as 'derivedFrom=TIM1', which means their registers will be identical. Fixing TIM1 should therefore automatically fix TIM8/20 too.

What's the motivation around deleting CCMR1_Input and renaming CCMR1_Output to just CCMR1? It looks like the SVD is correct to have both here, since you need to access this register in either mode, and svd2rust should handle creating accessors for the two modes. I think in other devices we keep CCMR1_Input and CCMR1_Output as-is, more or less.

burrbull:
Outdated and not planned.

PR #412, opened 2020-08-07
Add virtual DR8 and DR16 registers to the advanced CRC peripheral

Pagten:
The advanced CRC peripheral, present on e.g. the stm32f0 series, can process input data by byte, half-word or word. Currently, the CRC's DR register is only defined as a word-sized register, making it impossible to supply input data by half-word or byte.

This patch adds two "virtual" DR8 and DR16 registers to the advanced CRC peripheral. They are both mapped to the same base address as the DR register, but come with a width of 8 and 16 bits respectively. These virtual registers allow writing data per byte or half-word to the DR register.

From RM0091:
Screenshot from 2020-08-07 21-21-05

### DIFF
--- a/peripherals/crc/crc_advanced.yaml
+++ b/peripherals/crc/crc_advanced.yaml
@@ -12,6 +12,7 @@ CRC:
       size: 0x8
       access: read-write
       resetValue: 0xFF
+      alternateRegister: DR
       fields:
         DR8:
           description: Data register bits
--- a/peripherals/crc/crc_advanced.yaml
+++ b/peripherals/crc/crc_advanced.yaml
@@ -5,6 +5,34 @@ _include:
  - ./crc_basic.yaml
 
 CRC:
+  _add:
+    DR8:
+      description: Data register - byte sized
+      addressOffset: 0x0
+      size: 0x8
+      access: read-write
+      resetValue: 0xFF
+      fields:
+        DR8:
+          description: Data register bits
+          bitOffset: 0
+          bitWidth: 8
+    DR16:
+      description: Data register - half-word sized
+      addressOffset: 0x0
+      size: 0x10
+      access: read-write
+      resetValue: 0xFFFF
+      fields:
+        DR16:
+          description: Data register bits
+          bitOffset: 0
+          bitWidth: 16
+
+  DR8:
+    DR8: [0, 255]
+  DR16:
+    DR16: [0, 65535]
   CR:
     REV_IN:
       Normal: [0, "Bit order not affected"]
--- a/peripherals/crc/crc_advanced.yaml
+++ b/peripherals/crc/crc_advanced.yaml
@@ -24,6 +24,7 @@ CRC:
       size: 0x10
       access: read-write
       resetValue: 0xFFFF
+      alternateRegister: DR
       fields:
         DR16:
           description: Data register bits


adamgreig:
Hmm, interesting idea. When you test this, what does the resulting API look like? I think it would cause the `CRC` struct to no longer contain a `DR` member, only methods that allow access to the various widths, so it would be useful to see what exactly it looks like. Have you tested that it does let you generate 8-bit and 16-bit writes?

Right now it's still _possible_ to manually do an 8 bit write, but only by using an unsafe volatile_write call directly and passing it something like `&CRC.DR as *const u16` etc, which definitely isn't very ergonomic.

Pagten:
The register block now looks like this, as you described:
```
///Register block
#[repr(C)]
pub struct RegisterBlock {
    _reserved_0_dr: [u8; 4usize],
    ///0x04 - Independent data register
    pub idr: IDR,
    ///0x08 - Control register
    pub cr: CR,
    _reserved3: [u8; 4usize],
    ///0x10 - Initial CRC value
    pub init: INIT,
}
impl RegisterBlock {
    ///0x00 - Data register - half-word sized
    #[inline(always)]
    pub fn dr16(&self) -> &DR16 {
        unsafe { &*(((self as *const Self) as *const u8).add(0usize) as *const DR16) }
    }
    ///0x00 - Data register - half-word sized
    #[inline(always)]
    pub fn dr16_mut(&self) -> &mut DR16 {
        unsafe { &mut *(((self as *const Self) as *mut u8).add(0usize) as *mut DR16) }
    }
    ///0x00 - Data register - byte sized
    #[inline(always)]
    pub fn dr8(&self) -> &DR8 {
        unsafe { &*(((self as *const Self) as *const u8).add(0usize) as *const DR8) }
    }
    ///0x00 - Data register - byte sized
    #[inline(always)]
    pub fn dr8_mut(&self) -> &mut DR8 {
        unsafe { &mut *(((self as *const Self) as *mut u8).add(0usize) as *mut DR8) }
    }
    ///0x00 - Data register
    #[inline(always)]
    pub fn dr(&self) -> &DR {
        unsafe { &*(((self as *const Self) as *const u8).add(0usize) as *const DR) }
    }
    ///0x00 - Data register
    #[inline(always)]
    pub fn dr_mut(&self) -> &mut DR {
        unsafe { &mut *(((self as *const Self) as *mut u8).add(0usize) as *mut DR) }
    }
}
```

The DR16 type looks like this (similar for DR8):
```
///Data register - half-word sized
///
///This register you can [`read`](crate::generic::Reg::read), [`reset`](crate::generic::Reg::reset), [`write`](crate::generic::Reg::write), [`write_with_zero`](crate::generic::Reg::write_with_zero), [`modify`](crate::generic::Reg::modify). See [API](https://docs.rs/svd2rust/#read--modify--write-api).
///
///For information about available fields see [dr16](dr16) module
pub type DR16 = crate::Reg<u16, _DR16>;
#[allow(missing_docs)]
#[doc(hidden)]
pub struct _DR16;
///`read()` method returns [dr16::R](dr16::R) reader structure
impl crate::Readable for DR16 {}
///`write(|w| ..)` method takes [dr16::W](dr16::W) writer structure
impl crate::Writable for DR16 {}
///Data register - half-word sized
pub mod dr16 {
    ///Reader of register DR16
    pub type R = crate::R<u16, super::DR16>;
    ///Writer for register DR16
    pub type W = crate::W<u16, super::DR16>;
    ///Register DR16 `reset()`'s with value 0xffff
    impl crate::ResetValue for super::DR16 {
        type Type = u16;
        #[inline(always)]
        fn reset_value() -> Self::Type {
            0xffff
        }
    }
    ///Reader of field `DR16`
    pub type DR16_R = crate::R<u16, u16>;
    ///Write proxy for field `DR16`
    pub struct DR16_W<'a> {
        w: &'a mut W,
    }
    impl<'a> DR16_W<'a> {
        ///Writes raw bits to the field
        #[inline(always)]
        pub fn bits(self, value: u16) -> &'a mut W {
            self.w.bits = (self.w.bits & !0xffff) | ((value as u16) & 0xffff);
            self.w
        }
    }
    impl R {
        ///Bits 0:15 - Data register bits
        #[inline(always)]
        pub fn dr16(&self) -> DR16_R {
            DR16_R::new((self.bits & 0xffff) as u16)
        }
    }
    impl W {
        ///Bits 0:15 - Data register bits
        #[inline(always)]
        pub fn dr16(&mut self) -> DR16_W {
            DR16_W { w: self }
        }
    }
}
```

This is an example of how I'd use these registers:
```
pub fn update(crc: &mut stm32::CRC, bytes: &[u8]) {
    for &val in bytes {
        crc.dr8().write(|w| w.dr8().bits(val));
    }
}

pub fn result(crc: &stm32::CRC) {
    crc.dr().read().bits();
}
```

When I compile these functions:
```
#[inline(never)]
pub fn update_u8(crc: &mut stm32::CRC, byte: u8) {
    crc.dr8().write(|w| w.dr8().bits(byte));
}

#[inline(never)]
pub fn update_u16(crc: &mut stm32::CRC, byte: u16) {
    crc.dr16().write(|w| w.dr16().bits(byte));
}

#[inline(never)]
pub fn update_u32(crc: &mut stm32::CRC, byte: u32) {
    crc.dr().write(|w| w.dr().bits(byte));
}
```

I end up with this assembly code:
```
08001238 <update_u8>:
 8001238:	4901      	ldr	r1, [pc, #4]
 800123a:	7008      	strb	r0, [r1, #0]
 800123c:	4770      	bx	lr
 800123e:	46c0      	nop			; (mov r8, r8)
 8001240:	40023000 	.word	0x40023000

08001244 <update_u16>:
 8001244:	4901      	ldr	r1, [pc, #4]
 8001246:	8008      	strh	r0, [r1, #0]
 8001248:	4770      	bx	lr
 800124a:	46c0      	nop			; (mov r8, r8)
 800124c:	40023000 	.word	0x40023000

08001250 <update_u32>:
 8001250:	4901      	ldr	r1, [pc, #4]
 8001252:	6008      	str	r0, [r1, #0]
 8001254:	4770      	bx	lr
 8001256:	46c0      	nop			; (mov r8, r8)
 8001258:	40023000 	.word	0x40023000
```
So DR8, DR16 and DR are behaving as intended.

peripherals/crc/crc_advanced.yaml
```diff
@@ -5,6 +5,34 @@ _include:
...
+        DR8:
+          description: Data register bits
+          bitOffset: 0
+          bitWidth: 8
+    DR16:
+      description: Data register - half-word sized
+      addressOffset: 0x0
+      size: 0x10
+      access: read-write
+      resetValue: 0xFFFF
```
adamgreig, line 25:
```suggestion
      resetValue: 0xFFFF
      alternateRegister: DR
```

peripherals/crc/crc_advanced.yaml
```diff
@@ -5,6 +5,34 @@ _include:
  - ./crc_basic.yaml
 
 CRC:
+  _add:
+    DR8:
+      description: Data register - byte sized
+      addressOffset: 0x0
+      size: 0x8
+      access: read-write
+      resetValue: 0xFF
```
adamgreig, line 14:
```suggestion
      resetValue: 0xFF
      alternateRegister: DR
```

adamgreig (changes requested):
Thanks for the explanation! I think we should include this. I had a couple of reservations:

* Breaking change is pretty minor, not really a concern
* It's unusual to put new registers/fields into the `peripherals/` descriptions, which generally only contain enums, but in this case it applies to every SVD for this peripheral, and isn't fixing a mistake in the SVD so much as enhancing the description to make it easier to do 8-bit and 16-bit writes. I don't think there's any better way to express this in the SVD spec.
* We don't currently do this anywhere else, but there are other fields like SPI DR where the same logic applies and I believe currently HALs just take the address of the field and do a manual correct-sized write. Maybe we could use this pattern there too.

The only thing I think we should change is adding [`alternateRegister` tags](https://www.keil.com/pack/doc/CMSIS/SVD/html/elem_registers.html#elem_register) to mark in the SVD file that the redefinition of this address is intentional. I think it should suffice to just write `alternateRegister: DR` in each of these new registers, as per my suggested changes.

burrbull:
You reminded me to remove `_mut` version of alternate register access functions.

Pagten:
Thanks for the comments @adamgreig. The `alternateRegister` tag indeed applies here, so I've committed your suggested changes.

adamgreig (approved):
Thanks!

bors merge

PR #415, opened 2020-08-13
TIM21: Fix ccm missing options

almusil:
I have found some missing field options for `TIM2x`. This patch is probably not the best approach and I will be happy to fix it if anyone can please point me to the right direction why the `tim_ccm_v1.yaml` doesn't work for `TIM2x`.


almusil:
@adamgreig Can you please take a look?

adamgreig (approved):
Thanks, sorry for the wait!

bors merge

PR #416, opened 2020-08-17
stm32h7: rcc ccip patches and sdmmc prefix for b3

richardeoin:
Quick coverage of two points from https://github.com/stm32-rs/stm32h7xx-hal/pull/122

* Rename d3ccip -> srdccip (ref https://github.com/stm32-rs/stm32h7xx-hal/issues/95 ), fix field naming in that register
* Strip `SDMMC_` prefix from sdmmc registers


adamgreig (approved):
bors merge

PR #423, opened 2020-08-31
Add support for family with more than one letter

maximeborges:
We are trying to implement the support for the STM32MP1 family.
The `makecrates.py` script is currently retrieving the chip name by slicing the 7 first chars from the full chip name.
This pull request propose to match the chip name with a regex instead.

### DIFF
--- a/scripts/makecrates.py
+++ b/scripts/makecrates.py
@@ -13,6 +13,7 @@ import os
 import glob
 import os.path
 import argparse
+import re
 import yaml
 
 VERSION = "0.11.0"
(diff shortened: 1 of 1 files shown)


adamgreig (approved):
Thanks, this looks fine. At some point I guess we might need to get even cleverer (`stm32l4+`?) but this should work for now.

bors merge

maximeborges:
I checked quickly the L4P serie, I guess replacing the regex with
`stm32[a-z]+[0-9][pqrs]?`
should be enough ?

PR #424, opened 2020-08-31
Add optional `families` parameter to makecrates.py

maximeborges:
When running `make`, the `makecrates.py` scripts will retrieve every yaml files and run on every single one of them.
When making changes on a single family (or not all of them at least), one can edit the `CRATES` variable in `MakeFile` to specify the families that needs to be generated.
When running the script manually without specifying this new parameter, the behaviour will stay the same.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ svd/%.svd.formatted: svd/%.svd.patched
 
 # Generates the common crate files: Cargo.toml, build.rs, src/lib.rs, README.md
 crates:
-	python3 scripts/makecrates.py devices/ -y
+	python3 scripts/makecrates.py devices/ -y --families $(CRATES)
 
 define crate_template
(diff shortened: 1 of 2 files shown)


adamgreig:
Instead of editing the Makefile, you can also specify it as an environment variable, for example:

```
CRATES=stm32f0 make patch
```

That's intentional in how it's defined in the Makefile, to allow building for a single crate.

Your patch is good, but I don't understand why it's needed: `makecrates.py` should run very quickly, is only required very infrequently (before publishing, basically), and only writes to files that are git-ignored, so what is the benefit in having it only run for some crates?

maximeborges:
If the Makefile is done this way, I think it's better to actually use it that way: you want to work on a specific family, don't touch the other files.
I would more ask why it's actually doing more than what is asked by the Makefile.

adamgreig (approved):
Fair enough!

bors merge

PR #425, opened 2020-08-31
Add preliminary support for the STM32MP1 family

maximeborges:
Need #423

------------------------

Add minimal support for the **STM32MP157**, and documentation for the 151 and 153.

SVD file version **1.6** from Imen Ezzine from ST in this [ST Community thread](https://community.st.com/s/question/0D73W000000UX9k/your-stm32mp157-svd-file-is-highly-incomplete-and-plain-wrong-at-times-could-you-please-update-it)
This version has the proper interrupts compared to the **v1.2** available on the ST website.

**Discussion**: what would be the best way to add support for the 151 and 153 variant, since the only differences are some peripheral that are not available?
I think having a way to add `cfg(feature)`'able peripheral/registers/fields in the YAML patches could be a good way to do this kind of thing without duplicating all the code.
What do you think?


adamgreig:
> Discussion: what would be the best way to add support for the 151 and 153 variant, since the only differences are some peripheral that are not available?
I think having a way to add cfg(feature)'able peripheral/registers/fields in the YAML patches could be a good way to do this kind of thing without duplicating all the code.

We don't have any good way to do a `cfg(feature)` sort of thing; for that to end up in the Rust crates we'd need to express it in the YAML but also in the SVD format and have svd2rust support it.

I think you should be able to accomplish the same thing quite easily using the existing YAML syntax: put all the common YAML into a separate file, make a device YAML file for both the 151 and the 153, and have each device file only contain the relevant includes for the peripherals that it has. That way the only thing being copy-pasted are the include lines.

maximeborges:
At the moment I only have the full-featured SVD for the 157. For the 153 and 151 I need to remove some peripherals. What I understand from your proposition would be to use the SVD, make a common yaml that would basically be the 151, and then make two yaml for the 153 and 157 adding the peripherals that we removed for making the 151 ?
That would still duplicate the generated rust code right ? (I'm kinda new to embedded rust and don't have much experience with svd2rust)

adamgreig:
It would still duplicate the Rust code, yes, right now there's no way around that. If the devices are identical except some have a few more peripherals, then we've often just only generated a single device, for example we have `stm32f405` but not `stm32f415`; the only difference is that the 415 has the crypto peripherals and the 405 does not, so the `stm32f405` feature includes all those peripherals. Perhaps for the 153 and 151 you could do the same and only generate a single device, or perhaps even just have one device for all three if it's only some peripherals not present.

Do you want to merge this PR and then work on the 151/153 afterwards?

maximeborges:
Generating only the 157 is what is done at the moment, so I think this can be merged yes.

stm32_part_table.yaml
```diff
@@ -583,3 +583,26 @@ stm32g4:
...
+    rm_url: https://www.st.com/resource/en/reference_manual/dm00366349-stm32mp151-advanced-armbased-32bit-mpus-stmicroelectronics.pdf
+    members:
+      - STM32MP151
+  stm32mp153:
+    url: https://www.st.com/en/microcontrollers-microprocessors/stm32mp153.html
+    rm: RM0442
+    rm_title: STM32MP153
+    rm_url: https://www.st.com/resource/en/reference_manual/dm00366355-stm32mp153-advanced-armbased-32bit-mpus-stmicroelectronics.pdf
+    members:
+      - STM32MP153
```
adamgreig, line 601: We shouldn't add these to the parts table until they're included in the PAC; could you delete them for now and we'll re-add when device files are added for them?

svd/extract.sh
```diff
@@ -10,3 +10,4 @@ cp stm32h753x.svd stm32h753v.svd
 cp stm32h7x5_cm4.svd stm32h747cm4.svd
 cp stm32h7x5_cm7.svd stm32h747cm7.svd
 cp stm32h7b3x.svd stm32h7b3.svd
+cp stm32mp157x.svd stm32mp157.svd
```
adamgreig, line 13: There's no need to copy this, you can just put `stm32mp157x.svd` in the device file. If the 151/153 end up using the same SVD, then it would be appropriate to copy the file to a 151.svd and 153.svd.

adamgreig (approved):
Great, thanks!

bors merge

PR #426, opened 2020-09-01
Fix duplicate enums

Rahix:
As reported in https://github.com/stm32-rs/svdtools/pull/31, some patches here produce multiple overlapping `<enumeratedValues>` tags in the patched SVD file.  This can be fixed by ensuring the two patch statements in question don't overlap.

I'll leave this PR a draft until all discussions in the above mentioned PR to `svdtools` are resolved, in case anything else comes up.


Rahix:
I'd say this PR is ready now with the outstanding problem in the other PR solved.

adamgreig (approved):
Thanks, I'm glad it was just these two cases in the end!

bors merge

PR #440, opened 2020-09-21
h7: Add TIM3/4/12/13/14 interrupts not preserved through _copy

richardeoin:
This PR fixes a regression caused by the changes to `copy_peripheral` in https://github.com/stm32-rs/svdtools/pull/35. 

Previously interrupts from both the template peripheral and the target were generated in the patched SVD, resulting in two `<interrupt></interrupt>` blocks. Now only the interrupts from the template peripheral are preserved, so those from the target are lost. 

Confirmed that this is the only breakage for the h7 family by comparing mmaps for SVDs patched with svdtools 0.1.7 vs 0.1.8.

BTW - Thanks for all your hard work with the latest release  🎉 
Ref https://github.com/stm32-rs/stm32h7xx-hal/pull/134

### DIFF
--- a/devices/common_patches/h7_common_dualcore.yaml
+++ b/devices/common_patches/h7_common_dualcore.yaml
@@ -749,6 +749,20 @@ _copy:
   TIM14:
     from: TIM2
 
+# Add interrupts for TIM3 / TIM4
+TIM3:
+  _add:
+    _interrupts:
+      TIM3:
+        description: TIM3 global interrupt
+        value: 29
+TIM4:
+  _add:
+    _interrupts:
+      TIM4:
+        description: TIM4 global interrupt
+        value: 30
+
 SYSCFG:
   _add:
     PWRCR:
--- a/devices/common_patches/h7_common_singlecore.yaml
+++ b/devices/common_patches/h7_common_singlecore.yaml
@@ -539,3 +539,32 @@ _copy:
     from: TIM2
   TIM14:
     from: TIM2
+
+# Add interrupt for TIM3
+TIM3:
+  _add:
+    _interrupts:
+      TIM3:
+        description: TIM3 global interrupt
+        value: 29
+
+# Add interrupts for TIM12/TIM13/TIM14. For other SVDs these may be in TIM8, but
+# here they are in TIM12/TIM13/TIM14 and thus need to be added again
+TIM12:
+  _add:
+    _interrupts:
+      TIM8_BRK_TIM12:
+        description: TIM8 and 12 break global
+        value: 43
+TIM13:
+  _add:
+    _interrupts:
+      TIM8_UP_TIM13:
+        description: TIM8 and 13 update global
+        value: 44
+TIM14:
+  _add:
+    _interrupts:
+      TIM8_TRG_COM_TIM14:
+        description: TIM8 and 14 trigger/commutation and global
+        value: 45


richardeoin:
Here's the full list of differences to patched mmaps for the [stm32-rs master](https://github.com/stm32-rs/stm32-rs/commit/f43e5e582d702b2fe4b40180047f6c18226d8974), svdtools 0.1.7 vs 0.1.8. (I ignored stm32l4 and stm32l5)

Looks like f0 and f3 are affected also.

```diff
diff mmap17/stm32f0x1.svd.patched.mmap mmap18/stm32f0x1.svd.patched.mmap
6905,6906c6905
< INTERRUPT 015: TIM2 (TIM2): TIM2 global interrupt
< INTERRUPT 016: TIM3 (TIM3): TIM3 global interrupt
---
> INTERRUPT 015: TIM2 (TIM3): TIM2 global interrupt
diff mmap17/stm32f0x2.svd.patched.mmap mmap18/stm32f0x2.svd.patched.mmap
6248,6249c6248
< INTERRUPT 015: TIM2 (TIM2): TIM2 global interrupt
< INTERRUPT 016: TIM3 (TIM3): TIM3 global interrupt
---
> INTERRUPT 015: TIM2 (TIM3): TIM2 global interrupt
diff mmap17/stm32f0x8.svd.patched.mmap mmap18/stm32f0x8.svd.patched.mmap
6915,6916c6915
< INTERRUPT 015: TIM2 (TIM2): TIM2 global interrupt
< INTERRUPT 016: TIM3 (TIM3): TIM3 global interrupt
---
> INTERRUPT 015: TIM2 (TIM3): TIM2 global interrupt
diff mmap17/stm32f101.svd.patched.mmap mmap18/stm32f101.svd.patched.mmap
6361a6362
> INTERRUPT 018: ADC1_2 (ADC3): ADC1 and ADC2 global interrupt
6390d6390
< INTERRUPT 047: ADC3 (ADC3): ADC3 global interrupt
diff mmap17/stm32f102.svd.patched.mmap mmap18/stm32f102.svd.patched.mmap
5737a5738
> INTERRUPT 018: ADC1_2 (ADC3): ADC1 and ADC2 global interrupt
5766d5766
< INTERRUPT 047: ADC3 (ADC3): ADC3 global interrupt
diff mmap17/stm32f107.svd.patched.mmap mmap18/stm32f107.svd.patched.mmap
6172a6173
> INTERRUPT 018: ADC1_2 (ADC3): ADC1 and ADC2 global interrupt
6201d6201
< INTERRUPT 047: ADC3 (ADC3): ADC3 global interrupt
diff mmap17/stm32f302.svd.patched.mmap mmap18/stm32f302.svd.patched.mmap
8778,8780c8778
< INTERRUPT 028: TIM2 (TIM2): TIM2 global interrupt
< INTERRUPT 029: TIM3 (TIM3): TIM3 global interrupt
< INTERRUPT 030: TIM4 (TIM4): TIM4 global interrupt
---
> INTERRUPT 028: TIM2 (TIM4): TIM2 global interrupt
diff mmap17/stm32f303.svd.patched.mmap mmap18/stm32f303.svd.patched.mmap
9334,9336c9334
< INTERRUPT 028: TIM2 (TIM2): TIM2 global interrupt
< INTERRUPT 029: TIM3 (TIM3): TIM3 global interrupt
< INTERRUPT 030: TIM4 (TIM4): TIM4 global interrupt
---
> INTERRUPT 028: TIM2 (TIM4): TIM2 global interrupt
diff mmap17/stm32f410.svd.patched.mmap mmap18/stm32f410.svd.patched.mmap
3718c3718
< INTERRUPT 085: SPI5 (SPI1): SPI 5 global interrupt
---
> INTERRUPT 085: SPI5 (SPI5): SPI 5 global interrupt
diff mmap17/stm32h743.svd.patched.mmap mmap18/stm32h743.svd.patched.mmap
21884,21885c21884
< INTERRUPT 028: TIM2 (TIM4): TIM2 global interrupt
< INTERRUPT 029: TIM3 (TIM3): TIM3 global interrupt
---
> INTERRUPT 028: TIM2 (TIM14): TIM2 global interrupt
21898,21900d21896
< INTERRUPT 043: TIM8_BRK_TIM12 (TIM12): TIM8 and 12 break global
< INTERRUPT 044: TIM8_UP_TIM13 (TIM13): TIM8 and 13 update global
< INTERRUPT 045: TIM8_TRG_COM_TIM14 (TIM14): TIM8 and 14 trigger /commutation and global
diff mmap17/stm32h743v.svd.patched.mmap mmap18/stm32h743v.svd.patched.mmap
21890,21891c21890
< INTERRUPT 028: TIM2 (TIM4): TIM2 global interrupt
< INTERRUPT 029: TIM3 (TIM3): TIM3 global interrupt
---
> INTERRUPT 028: TIM2 (TIM14): TIM2 global interrupt
21904,21906d21902
< INTERRUPT 043: TIM8_BRK_TIM12 (TIM12): TIM8 and 12 break global
< INTERRUPT 044: TIM8_UP_TIM13 (TIM13): TIM8 and 13 update global
< INTERRUPT 045: TIM8_TRG_COM_TIM14 (TIM14): TIM8 and 14 trigger /commutation and global
diff mmap17/stm32h747cm4.svd.patched.mmap mmap18/stm32h747cm4.svd.patched.mmap
22705,22706d22704
< INTERRUPT 029: TIM3 (TIM3): TIM3 global interrupt
< INTERRUPT 030: TIM4 (TIM4): TIM4 global interrupt
diff mmap17/stm32h747cm7.svd.patched.mmap mmap18/stm32h747cm7.svd.patched.mmap
22741,22742d22740
< INTERRUPT 029: TIM3 (TIM3): TIM3 global interrupt
< INTERRUPT 030: TIM4 (TIM4): TIM4 global interrupt
diff mmap17/stm32h753.svd.patched.mmap mmap18/stm32h753.svd.patched.mmap
22131,22132c22131
< INTERRUPT 028: TIM2 (TIM4): TIM2 global interrupt
< INTERRUPT 029: TIM3 (TIM3): TIM3 global interrupt
---
> INTERRUPT 028: TIM2 (TIM14): TIM2 global interrupt
22145,22147d22143
< INTERRUPT 043: TIM8_BRK_TIM12 (TIM12): TIM8 and 12 break global
< INTERRUPT 044: TIM8_UP_TIM13 (TIM13): TIM8 and 13 update global
< INTERRUPT 045: TIM8_TRG_COM_TIM14 (TIM14): TIM8 and 14 trigger /commutation and global
diff mmap17/stm32h753v.svd.patched.mmap mmap18/stm32h753v.svd.patched.mmap
22137,22138c22137
< INTERRUPT 028: TIM2 (TIM4): TIM2 global interrupt
< INTERRUPT 029: TIM3 (TIM3): TIM3 global interrupt
---
> INTERRUPT 028: TIM2 (TIM14): TIM2 global interrupt
22151,22153d22149
< INTERRUPT 043: TIM8_BRK_TIM12 (TIM12): TIM8 and 12 break global
< INTERRUPT 044: TIM8_UP_TIM13 (TIM13): TIM8 and 13 update global
< INTERRUPT 045: TIM8_TRG_COM_TIM14 (TIM14): TIM8 and 14 trigger /commutation and global
```

richardeoin:
Oh, the newest h7b3 svds weren't affected by this at all

adamgreig:
Thanks for catching this! Evidently I really should have checked the stm32-rs mmaps after the change. I'll release a 0.12.1 ASAP once this is fixed.

I think `copy_peripheral` contains a couple of bugs actually; the documentation specifically says 'interrupts are not copied', but that's not true for a few reasons; the old code has a bug where for making new peripherals it checks for `value.tag in ("interrupt")` but should be a list or tuple so is instead checking if `value.tag` is `"i"`, `"n"`, `"t"`, etc and so will copy all interrupts, while for copying into an existing peripheral it would remove as many interrupts from the copy as the original-to-be-replaced had but no more. After my change it makes even less sense for interrupts and the behaviour is a weird mix of how many interrupts each tag had.

It seems to me like the correct behaviour should be to never copy interrupts from the template and always preserve those found in the target, what do you think?

>  (I ignored stm32l4 and stm32l5)

Were they especially bad or was there some other reason to ignore them?

richardeoin:
No problem! I'd agree that using the target interrupts only would be a much more reasonable behaviour. Currently the _copy behaviour for interrupts is not ideal as you've described, and additionally because the interrupts from the template peripheral end up in the patched SVD multiple times. 

I initially ignored stm32l4 and stm32l5 because since https://github.com/stm32-rs/stm32-rs/pull/435 it's not possible to patch them with svdtools 0.17.0. Actually it appears there's changes associated with every use of `_copy`, and there are four of those in stm32l4 (for timers again). So going back to ddc0f65 to make the comparison:

```diff
diff mmap17/stm32l4x1.svd.patched.mmap mmap18/stm32l4x1.svd.patched.mmap
7210,7211c7210
< INTERRUPT 028: TIM2 (TIM2): TIM2 global interrupt
< INTERRUPT 029: TIM3 (TIM3): TIM3 global interrupt
---
> INTERRUPT 028: TIM2 (TIM3): TIM2 global interrupt
diff mmap17/stm32l4x2.svd.patched.mmap mmap18/stm32l4x2.svd.patched.mmap
7068,7069c7068
< INTERRUPT 028: TIM2 (TIM2): TIM2 global interrupt
< INTERRUPT 029: TIM3 (TIM3): TIM3 global interrupt
---
> INTERRUPT 028: TIM2 (TIM3): TIM2 global interrupt
diff mmap17/stm32l4x5.svd.patched.mmap mmap18/stm32l4x5.svd.patched.mmap
8564,8566c8564
< INTERRUPT 028: TIM2 (TIM2): TIM2 global interrupt
< INTERRUPT 029: TIM3 (TIM3): TIM3 global interrupt
< INTERRUPT 030: TIM4 (TIM4): TIM4 global interrupt
---
> INTERRUPT 028: TIM2 (TIM4): TIM2 global interrupt
diff mmap17/stm32l4x6.svd.patched.mmap mmap18/stm32l4x6.svd.patched.mmap
10527,10529c10527
< INTERRUPT 028: TIM2 (TIM2): TIM2 global interrupt
< INTERRUPT 029: TIM3 (TIM3): TIM3 global interrupt
< INTERRUPT 030: TIM4 (TIM4): TIM4 global interrupt
---
> INTERRUPT 028: TIM2 (TIM4): TIM2 global interrupt
```
(there's none in stm32l5)

adamgreig:
I've updated svdtools in https://github.com/stm32-rs/svdtools/pull/38 to always preserve interrupts from the target. The difference on stm32-rs ddc0f65 between svdtools 0.1.7 and the new fix in 38 is below. It looks like a few interrupts that had gone missing are restored (correctly), and other interrupts are now associated with the correct peripheral instead a wrong one, but no interrupts are missing. Hopefully this means this PR is no longer required; I'll release svdtools 0.9.0 and stm32-rs 0.12.1 with the changes soon.

```diff
diff mmaps_017/stm32f101.svd.patched.mmap mmaps_fix/stm32f101.svd.patched.mmap
6361a6362
> INTERRUPT 018: ADC1_2 (ADC3): ADC1 and ADC2 global interrupt
diff mmaps_017/stm32f102.svd.patched.mmap mmaps_fix/stm32f102.svd.patched.mmap
5737a5738
> INTERRUPT 018: ADC1_2 (ADC3): ADC1 and ADC2 global interrupt
diff mmaps_017/stm32f107.svd.patched.mmap mmaps_fix/stm32f107.svd.patched.mmap
6172a6173
> INTERRUPT 018: ADC1_2 (ADC3): ADC1 and ADC2 global interrupt
diff mmaps_017/stm32f410.svd.patched.mmap mmaps_fix/stm32f410.svd.patched.mmap
3718c3718
< INTERRUPT 085: SPI5 (SPI1): SPI 5 global interrupt
---
> INTERRUPT 085: SPI5 (SPI5): SPI 5 global interrupt
diff mmaps_017/stm32h743.svd.patched.mmap mmaps_fix/stm32h743.svd.patched.mmap
21884c21884
< INTERRUPT 028: TIM2 (TIM4): TIM2 global interrupt
---
> INTERRUPT 028: TIM2 (TIM2): TIM2 global interrupt
21964c21964
< INTERRUPT 115: SWPMI1 (TIM14): SWPMI global interrupt
---
> INTERRUPT 115: SWPMI1 (TIM2): SWPMI global interrupt
diff mmaps_017/stm32h743v.svd.patched.mmap mmaps_fix/stm32h743v.svd.patched.mmap
21890c21890
< INTERRUPT 028: TIM2 (TIM4): TIM2 global interrupt
---
> INTERRUPT 028: TIM2 (TIM2): TIM2 global interrupt
21970c21970
< INTERRUPT 115: SWPMI1 (TIM14): SWPMI global interrupt
---
> INTERRUPT 115: SWPMI1 (TIM2): SWPMI global interrupt
diff mmaps_017/stm32h747cm4.svd.patched.mmap mmaps_fix/stm32h747cm4.svd.patched.mmap
22704c22704
< INTERRUPT 028: TIM2 (TIM14): TIM2 global interrupt
---
> INTERRUPT 028: TIM2 (TIM2): TIM2 global interrupt
diff mmaps_017/stm32h747cm7.svd.patched.mmap mmaps_fix/stm32h747cm7.svd.patched.mmap
22740c22740
< INTERRUPT 028: TIM2 (TIM14): TIM2 global interrupt
---
> INTERRUPT 028: TIM2 (TIM2): TIM2 global interrupt
diff mmaps_017/stm32h753.svd.patched.mmap mmaps_fix/stm32h753.svd.patched.mmap
22131c22131
< INTERRUPT 028: TIM2 (TIM4): TIM2 global interrupt
---
> INTERRUPT 028: TIM2 (TIM2): TIM2 global interrupt
22213c22213
< INTERRUPT 115: SWPMI1 (TIM14): SWPMI global interrupt
---
> INTERRUPT 115: SWPMI1 (TIM2): SWPMI global interrupt
diff mmaps_017/stm32h753v.svd.patched.mmap mmaps_fix/stm32h753v.svd.patched.mmap
22137c22137
< INTERRUPT 028: TIM2 (TIM4): TIM2 global interrupt
---
> INTERRUPT 028: TIM2 (TIM2): TIM2 global interrupt
22219c22219
< INTERRUPT 115: SWPMI1 (TIM14): SWPMI global interrupt
---
> INTERRUPT 115: SWPMI1 (TIM2): SWPMI global interrupt
```

adamgreig:
All done, fixed by #441.

PR #442, opened 2020-09-24
Fix cargo feature list for stm32g0 series

ijager:


adamgreig:
Thanks for fixing this. Is there much difference between the 070 and 071 that warrants including both? I'm worried that having all those features will increase memory usage beyond the current 3GB limit, which would cause builds to fail. If 070 is a subset of 071, perhaps we could only include 071 in the docs?

ijager:
Yes there are a lot of differences between G070 and G071. There are less differences between G081 and G071. The G081 and G041 versions add AES stuff. Here is a nice overview of the differences: https://github.com/stm32-rs/stm32g0xx-hal#g0-hardware-overview

So perhaps we could remove G071 and G031 as those are subsets of G081 and G041 respectively.

ijager:
And I guess G030 is a subset of G070

adamgreig:
It looks like we need to request some memory limit increases for other crates anyway (#443) so I'll test out how much RAM this PR requires and see if we can just build it as-is first.

adamgreig (approved):
Sorry for the wait on this. I'm going to merge your PR and then remove the subset devices (030, 031, 071 perhaps) until we get under the memory limit, since I plan to publish a new version within a few days.

bors merge

adamgreig:
I measured 2.7GB for the current features, hopefully this will build OK as-is.

PR #444, opened 2020-09-24
Add STM32WL family with stm32wle5 chip

ijager:
This patch adds the stm32wl family. I hope I followed all the naming conventions correctly since it is a bit different from the rest of the stm32 series.


devices/stm32wle5.yaml
```diff
@@ -0,0 +1,4 @@
+_svd: ../svd/STM32WLE5.svd
```
adamgreig, line 1:
```suggestion
_svd: ../svd/stm32wle5.svd
```

The extraction script makes all the SVD files lowercase.

Makefile
```diff
@@ -6,7 +6,7 @@ all: patch svd2rust
 SHELL := /usr/bin/env bash
 
 CRATES ?= stm32f0 stm32f1 stm32f2 stm32f3 stm32f4 stm32f7 stm32h7 \
-          stm32l0 stm32l1 stm32l4 stm32l5 stm32g0 stm32g4 stm32mp1
+          stm32l0 stm32l1 stm32l4 stm32l5 stm32g0 stm32g4 stm32mp1 stm32wl
```
adamgreig, line 9:
```suggestion
          stm32l0 stm32l1 stm32l4 stm32l5 stm32g0 stm32g4 stm32mp1 \
          stm32wl
```

stm32_part_table.yaml
```diff
@@ -591,4 +591,13 @@ stm32mp1:
     rm_title: STM32MP157
     rm_url: https://www.st.com/resource/en/reference_manual/dm00366355-stm32mp157-advanced-armbased-32bit-mpus-stmicroelectronics.pdf
     members:
-      - STM32MP157
\ No newline at end of file
+      - STM32MP157
+
+stm32wl:
+  stm32wlex:
```
adamgreig, line 597: These have to match the module name inside the crate, so either we could rename the YAML file to stm32wlex.yaml or rename this line to stm32wle5. Given we don't know what ST will release next, probably stm32wle5 is safest.

```suggestion
  stm32wle5:
```

scripts/makecrates.py
```diff
@@ -220,6 +222,8 @@ def main(devices_path, yes, families):
         yamlfile = os.path.basename(path)
         family = re.match(r'stm32[a-z]+[0-9]', yamlfile)[0]
         device = os.path.splitext(yamlfile)[0].lower()
+        if family == 'stm32wle5':
```
adamgreig, line 225:
```suggestion
        if family.startswith('stm32wl'):
```

This should be more future-proof.

adamgreig (changes requested):
Thanks for this PR! The WL family looks really neat. I've identified a couple of changes to make this a bit more future-proof for additional WL devices, let me know what you think.

stm32_part_table.yaml
ijager: For now we can change it to `stm32wle5` like you suggested. In the future we might see a STM32WLE**6** or something, maybe then it should be stm32wle**x** again? I believe they are planning to release a dual core version with M0 and M4 cores, because they accidentally sent me SVD files for that part which is called `STM32WL5x`, so a 5 instead of the E...

adamgreig (approved):
Thanks!

bors merge

adamgreig:
This is now published :tada: https://crates.io/crates/stm32wl

PR #447, opened 2020-09-30
Create common patch for missing CRC fields in F7 SVDs

chengsun:


adamgreig (approved):
Thanks for this patch, and sorry for the delay reviewing it!

bors merge

PR #454, opened 2020-10-20
stm32h7: add cluster for MDMA registers

richardeoin:
This applies ok with the current svdtools (`svd patch`), but https://github.com/stm32-rs/svdtools/pull/42 is needed to correctly identify the dimIndex. 

For current svdtools, because `spec_ind()` prefers `?` to `[..]`, placing the `?` pattern after the comma means that `rname[li : len(rname) - ri]` is always `""`. At least that's reasonably neat.


adamgreig:
bors retry

adamgreig:
bors try

adamgreig:
This looks good and I've now released svdtools 0.1.10 that includes your fix there too. The only thing I noticed is that previously the ISR and IFCR registers had fields like `TEIF4` where now they're all `TIEF0`. Maybe we should rename them to just `TIEF` so the ones for e.g. channel 4 don't say 0?

richardeoin:
Thanks for the release!

adamgreig (approved):
Thanks!

bors merge

PR #458, opened 2020-11-27
Fix H7 CRC INIT and POL register addresses

AlyoshaVasilieva:
According to RM0433, RM0399, and RM0455 `INIT` is at 0x10 and `POL` is at 0x14. (See for example [page 798/section 21.4.4 of RM0433](https://www.st.com/resource/en/reference_manual/dm00314099-stm32h742-stm32h743-753-and-stm32h750-value-line-advanced-arm-based-32-bit-mcus-stmicroelectronics.pdf#page=798).) SVDs have them both four bytes earlier.

Tested on an STM32H750VBT6.


adamgreig (approved):
Thanks!

bors merge

PR #461, opened 2020-12-02
Define CRC IDR as 8-bit or 32-bit as appropriate

AlyoshaVasilieva:
The CRC peripheral's independent data register only allows 8 bits to be used on most chips. On all H7 chips (RM0399, RM0433, RM0455, RM0468) it allows all 32 bits to be used. [As an example, RM0433 page 797](https://www.st.com/resource/en/reference_manual/dm00314099-stm32h742-stm32h743753-and-stm32h750-value-line-advanced-armbased-32bit-mcus-stmicroelectronics.pdf#page=797).

H7 chips are not the only chips that have a 32-bit IDR, but I don't have any of the others: the G4 series, G0 series, L5 series, and the STM32F334. Probably others, I didn't check every reference manual. Since I can't test them at all I haven't included them here, but I'll add this patch to the G0/G4 chips if desired.

I can't determine any pattern for which chips support it, as chips with the advanced peripheral + configurable polynomials do not necessarily have a 32-bit IDR (eg F373).

Tested on an STM32H743VIT6.

### DIFF
--- a/devices/common_patches/crc/crc_32bit_idr.yaml
+++ b/devices/common_patches/crc/crc_32bit_idr.yaml
@@ -1,8 +1,0 @@
-# Some CRC peripherals allow all 32-bits of their independent data register to be used.
-# Whether it's safe to use all 32-bits appears independent of other CRC features.
-
-CRC:
-  IDR:
-    _modify:
-      IDR:
-        _write_constraint: [0, 0xFFFFFFFF]
--- a/devices/common_patches/crc/crc_32bit_idr.yaml
+++ b/devices/common_patches/crc/crc_32bit_idr.yaml
@@ -5,4 +5,4 @@ CRC:
   IDR:
     _modify:
       IDR:
-        _write_constraint: [0, 4294967295]
+        _write_constraint: [0, 0xFFFFFFFF]
--- a/devices/common_patches/crc/crc_32bit_idr.yaml
+++ b/devices/common_patches/crc/crc_32bit_idr.yaml
@@ -1,0 +1,8 @@
+# Some CRC peripherals allow all 32-bits of their independent data register to be used.
+# Whether it's safe to use all 32-bits appears independent of other CRC features.
+
+CRC:
+  IDR:
+    _modify:
+      IDR:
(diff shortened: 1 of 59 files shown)


AlyoshaVasilieva:
Apparently this fails the linting step, which I didn't realize existed. It builds successfully, though. Beyond splitting out the definition of IDR from `crc_basic.yaml` and manually setting it to 8-bit or 32-bit on each chip I'm not sure how to write this patch properly. Will attempt to figure it out.

adamgreig:
It looks like your modify is running first, and then the IDR write constraint from `crc_basic.yaml` is being added afterwards, resulting in two write constraints in the file (which is what the linter rejects). Not sure what the best way to fix that is. By the way, you can use 0xFFFFFFFF as the constant, it's a bit more obvious/easier to check.

AlyoshaVasilieva:
I think removing the IDR from `crc_basic` and defining it separately per chip as 8-bit or 32-bit might work, I'll try that tomorrow or so.

AlyoshaVasilieva:
I've split IDR out of `crc_basic.yaml` and given it two separate definitions, 8-bit or 32-bit. I then added this to every device definition that included `crc_basic` or `crc_advanced` and checked the reference manuals to ensure the selection was correct.

The definitions override the `bitWidth` because:

* The `stm32f3x4` SVD gives the width as 8-bit but [RM0364 states it is 32-bit](https://www.st.com/resource/en/reference_manual/dm00093941-stm32f334xx-advanced-armbased-32bit-mcus-stmicroelectronics.pdf#page=79).
* The STM32L100, L151, L162 SVDs give the width as 7 bits but [RM0038 states it is 8-bit](https://www.st.com/resource/en/reference_manual/cd00240193-stm32l100xx-stm32l151xx-stm32l152xx-and-stm32l162xx-advanced-armbased-32bit-mcus-stmicroelectronics.pdf#page=96).

There's probably more that could be done here to promote interoperability of this register across chips (like an "IDR8" register or something) but I don't really think that's worth it. This register is just temporary data storage.

adamgreig (approved):
Great work, thanks!

bors merge

PR #463, opened 2020-12-09
Added ADC Common register for L4 series

korken89:


devices/common_patches/l4_adc_common.yaml
```diff
@@ -0,0 +1,166 @@
...
+      size: 0xC
+      usage: registers
+    registers:
+      CSR:
+        description: ADC common status register
+        addressOffset: 0x0
+        access: read-only
+        resetValue: 0x00000000
+        fields:
+          JQOVF_SLV:
```
adamgreig, line 20: As far as I can tell, none of the L4x1/2/3 chips (i.e. any covered by RM0394) actually contain any of the `_SLV` fields, it looks like they're only present on the L4x5?

devices/common_patches/l4_adc_common.yaml
```diff
@@ -0,0 +1,166 @@
...
+            bitWidth: 1
+          PRESC:
+            description: ADC prescaler
+            bitOffset: 18
+            bitWidth: 4
+          CKMODE:
+            description: ADC clock mode
+            bitOffset: 16
+            bitWidth: 2
+          MDMA:
```
adamgreig, line 136: Similarly it looks like MDMA, DMACFG, DELAY, and DUAL are not in L4x1/2/3.

devices/common_patches/l4_adc_common.yaml
```diff
@@ -0,0 +1,166 @@
...
+          DELAY:
+            description: Delay between 2 sampling phases
+            bitOffset: 8
+            bitWidth: 4
+          DUAL:
+            description: Dual ADC mode selection
+            bitOffset: 0
+            bitWidth: 4
+
+      CDR:
```
adamgreig, line 153: Also looks like CDR is only in L4x5?

adamgreig (commented):
From looking at RM0934 it seems like a bunch of the fields to do with using multiple ADCs are not available in the L4x1/2/3 parts covered by RM0394, and we don't use this patch for L4x5/6 which already have the ADC_Common peripheral in their SVDs, so are they needed here? I might just be misunderstanding how the L4 family is split up though.

korken89:
The extra field should be available for all parts that have 2 ADCs as I understand it.
This should be l41x, not sure on the rest. 
Having a look in RM0394 I am not clear on which field do not exist in which MCU, as from what I can find it is not stated.
The only info I can find is `Table 54. Main ADC features`.

Should I remove the slave bits for the non l41x chips?

devices/common_patches/l4_adc_common.yaml
korken89: From what I find, only the L41x should be like this? See `Table 54`.

devices/common_patches/l4_adc_common.yaml
```diff
@@ -0,0 +1,166 @@
...
+            bitOffset: 13
+            bitWidth: 1
+          DELAY:
+            description: Delay between 2 sampling phases
+            bitOffset: 8
+            bitWidth: 4
+          DUAL:
+            description: Dual ADC mode selection
+            bitOffset: 0
+            bitWidth: 4
```
adamgreig, line 151: I think this should be 5 bits?
```suggestion
            bitWidth: 5
```

devices/common_patches/l4_adc_common.yaml
```diff
@@ -0,0 +1,166 @@
...
+  # This SVD is missing the ADC_Common peripheral that most other parts with
+  # this ADC contain; consequently it's missing the CSR and CCR and CDR regs
+  # from RM0394.
+  ADC_Common:
+    description: ADC common registers
+    groupName: ADC
+    baseAddress: 0x50040300
+    addressBlock:
+      offset: 0
+      size: 0xC
```
adamgreig, line 11: I think this actually needs to be 0x10 because there's 4 reserved bytes in the block; the ADR_CDR register starts at 0xC then continues for 4 bytes.
```suggestion
      size: 0x10
```

devices/common_patches/l4_adc_common.yaml
adamgreig: Oh, I see what's happened; I have an older copy of RM0394 from when it covered L43/L43/L45/L46 and at the time L4x1 was covered by a separate RM0392; the latest revision of RM0394 covers L41x/L42x/L43x/L44x/L45x and RM0351 covers the L47x/L48x/L49x/L4Ax. Annoyingly it means the division of SVD files doesn't actually match the division of datasheets, go figure.

Anyway now that I'm using the latest RM0394 everything makes sense and I think we can go ahead with all the _SLV bits too.

adamgreig (changes requested):

korken89:
Ahh, I see - funny division of datasheets xD
Thanks for the comments and fixes are commited!

adamgreig (approved):
Thanks!

bors merge

Issue #226, opened 2019-05-31
Support for STM32F750


ondono:
I was thinking in getting started in embedded rust using a STM32F7508-DK board, that uses a STM32F50N8 Microcontroller.

I've seen there's apparently no support yet, not seeing it in the Device Coverage list either. I just wanted to make sure that there's no major roadblock that I'm not yet able to see, it's okay if it's just that nobody has started to work on it, I'm willing to do it.

Thank you!

adamgreig:
You're the second person to ask about F7x0 support recently! I'm on a work
trip at the moment so can't work on it yet, but it shouldn't be too bad if
you want to open a PR.

You'd need to update to the latest f7 SVD zip (in the svd/vendor folder)
and then add new 730 and 750 device yaml files. Ideally also include
appropriate patches and peripherals (see the other f7 device files for
inspiration).

Shout if you run into any issues or need help with anything.

ondono:
I went for it and I already have the board. I have quite a lot to figure out since I'm new to rust and I'm still trying to understand all the tooling and libraries to work with embedded devices.

I've found an issue though, and I'm not sure what's the procedure:
- I've loaded the latest .zip file from ST for the f7 family.
- The STM32F7X5.svd file is no longer present on the .zip file, I think it has been split into STM32F745.svd and STM32F765.svd
- I've also created STM32F745.yaml and STM32F765.yaml by just copy pasting the previous STM32F7X5.yaml file

This seems to at least let me build the modified stm32f7 crate, though I'm still modifying it away to add the 750. My main worry is that I have no way to test that I have not broken the 745 & 765 libs by accident since I have no way to test them on real hardware. What is the proper way forward?

ogoffart:
Hi @ondono   
I also have a STM32F7508-DK board and wanted to try rust on it. So thanks for your work.
I found your fork of this repository by clicking on your name and gave it a try

I get this error:

```
python3 scripts/svdpatch.py devices/stm32f7x9.yaml
[...]
  File "scripts/svdpatch.py", line 614, in process_field_enum
    .format(pname, rname, fspec))
__main__.MissingFieldError: Could not find FLASH:CR.MER
make: *** [Makefile:38: svd/stm32f7x9.svd.patched] Error 1
```

Did you also get that? 
Anyway since it is not related to the board i have, i just went ahead and tried to build an hello world. (a simple empty loop). And it worked.

[...]

PR #464, opened 2020-12-15
Add STM32F750 device

chengsun:
This addresses #226.

I tested this on a STM32F7508-DK Discovery kit, and I can say that at least the most important peripherals work. 

The peripherals were matched automatically using `matchperipherals.py`; the `patch` process succeeds, but the majority remain untested on the device itself.


adamgreig (approved):
Looks great, thank you!

bors merge

PR #465, opened 2020-12-17
ADC -> ADC1 fix for STM32L4xx series

korken89:
Hi, I renamed the ADC to ADC1 to follow the datasheet.
It was not needed for the rest of the series, they had the correct name.

One question, ADC2 is missing and it shouly be on L412 and L422, but the feature flags do not have this granularity.
How should I add this?

Table for overview:

image
image

### DIFF
--- a/devices/stm32l4x1.yaml
+++ b/devices/stm32l4x1.yaml
@@ -18,6 +18,11 @@ RCC:
         name: USART3RST
         description: USART3 reset
 
+_modify:
+  # The SVD calls ADC1 ADC.
+  ADC:
+    name: ADC1
+
 # Merge the thousands of individal bit fields into a single field for each
 # CAN filter register. This is not only much easier to use but also saves
 # a huge amount of filespace and compilation time etc -- as much as 30% of all
--- a/devices/stm32l4x1.yaml
+++ b/devices/stm32l4x1.yaml
@@ -23,6 +23,12 @@ _modify:
   ADC:
     name: ADC1
 
+ADC1:
+  _modify:
+    _interrupts:
+      ADC1:
+        name: ADC1_2
+
 _add:
   ADC2:
     derivedFrom: ADC1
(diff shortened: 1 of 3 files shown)


korken89:
Huh, for some reason the change did not happen for `stm32l4x2`, not sure why.
The change is the same in all the YAMLs and checking the `stm32l4x2` memory map it still has the name `ADC`.

korken89:
Hmm, I have been playing around with the syntax, but nothing seems to stick on the `l4x2`.

Any ideas what I am doing wrong? Maybe @adamgreig ?

adamgreig:
Sorry for the wait. I just had a look, I think the problem here is that the L4x2 YAML file contains two `_modify` keys at the top level, and PyYAML just takes the second one. Possibly it should reject the file as invalid entirely. It's a bit confusing for stm32-rs, because our `_include` directive will merge the contents of identical keys if they occur in separate files, but it can't do anything when they're duplicated inside the same file.

In this case I think just combining the ADC rename with the USB rename should fix the issue.

korken89:
Thanks @adamgreig, that did the trick!
I'd say this is ready for merge unless we want to enable ADC2 somehow?

I had a look at the HAL discussions and the plan seems to be that the more granular distinction will me made there.

adamgreig:
Thanks, renames all look good. Perhaps the best bet is to add the ADC2 peripheral block to the L4x2 device here, and the HALs can have fine-grained features to only support it on the relevant variants. That way at least the registers will exist; someone using an L4x2 with only one ADC might try and use the non-existent ADC2 but that's the same situation we have with some other devices where some variants have extra features and isn't generally a big problem.

Assuming all the registers are the same, you could add it like:


```yaml
_add:
    ADC2:
        derivedFrom: ADC1
        baseAddress: "0x400XXXXX"
```

Otherwise, you could use `_copy` to copy all the ADC1 registers explicitly and then just modify the baseAddress.

korken89:
Thanks for the help @adamgreig !
I have added ADC2 (luckily ADC1 and 2 are the same) and I merged interrupt names for the entire L4 family.

I think this is ready now for the proper review and merge.

adamgreig (approved):
Looks good, thanks!

bors merge

PR #466, opened 2020-12-24
Fix GPIO OSPEEDR variant on f3

Piroro-hs:
F3 mcus does not support very high speed mode in OSPEEDR register.

image

This patch ignores the `0b10`.


adamgreig (approved):
Thanks!

bors merge

PR #467, opened 2020-12-26
Add STM32WB55 chip

eupn:
Let's start adding the STM32WBx family by adding the WB55 chip which I currently use.


adamgreig (approved):
Thanks, looks good!

We also need to update the README and bors.toml, but I'll do that after this merges.

bors merge

adamgreig:
This is now released as [stm32wb](https://crates.io/crates/stm32wb) :tada:

PR #468, opened 2020-12-30
Fix STM32L552 TIMx interrupts

kenbell:
The timer documentation also seems 'broken'.  Table from RM0438 below.  I'm assuming acronym column is incorrect in the doc, so fixing that as well as adding missing interrupts from TIM6 & TIM7.
image


adamgreig (approved):
Thanks!

bors merge

PR #472, opened 2021-01-04
stm32l5: add enum values for RCC

kenbell:
Instead of duplicating much of rcc_common.yaml, could instead look to factor rcc_common.yaml so that it applies cleanly to L5 MCUs.  Let me know if that approach would be preferred.


adamgreig (approved):
Sorry for the wait, looks good!

bors merge

PR #473, opened 2021-01-07
wb55: fix SYSCFG base address (and Makefile)

eupn:
This is a batch of two fixes:

- Uses the correct base address from the RM for `SYSCFG`. This fix was checked to work in the hardware
- Fixes `Makefile` to produce correct directory

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ SHELL := /usr/bin/env bash
 
 CRATES ?= stm32f0 stm32f1 stm32f2 stm32f3 stm32f4 stm32f7 stm32h7 \
           stm32l0 stm32l1 stm32l4 stm32l5 stm32g0 stm32g4 stm32mp1 \
-          stm32wl stm32wb55
+          stm32wl stm32wb
 
 # All yaml files in devices/ will be used to patch an SVD
 YAMLS := $(foreach crate, $(CRATES), \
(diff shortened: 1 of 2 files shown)


adamgreig:
Thanks!

It looks like the SYSCFG problems run deeper: the reason the SVD has the wrong address is because SYSCFG is split between 4001_0000-4001_002F and 4001_0100-4001_01FF with VREFBUF stuck in the middle. As well as updating the baseAddress as you've done, we should update the addressBlock size to 0x200. Unfortunately it seems all the registers in the 4001_0100 region are wrong in the SVD too (IMR1, IMR2, C2IMR1, C2IMR2, SIPCR) as their offsets ignore the jump over VREFBUF so need updating to 0x100, 0x104, etc.

eupn:
@adamgreig so the next steps should be joining those two together with the VREFBUF area marked as reserved/skipped to get a contiguous memory map for SYSCFG?

adamgreig:
> @adamgreig so the next steps should be joining those two together with the VREFBUF area marked as reserved/skipped to get a contiguous memory map for SYSCFG?

Yes, I think that's the best way to go.

jorgeig-space:
The same problem exists in the WL family. I'll try to get to it this week and perhaps you can use the same patches here.

adamgreig:
#493 cherry-picked the Makefile fix, and #624 fixes SYSCFG, so it looks like this PR can be closed now. Thanks for getting it started!

PR #478, opened 2021-01-10
Add basic patches for FDCAN

richardeoin:
FDCAN appears in the H7, G4, L5 and MP1 families. The peripheral varies between the families, with the G4/L5 versions being somewhat simpler. I don't see any variation within families, but haven't exhaustively checked for this.

On the H7, the FDCAN2 doesn't support TTCAN. However FDCAN2 is `derivedFrom="FDCAN1"` in the original ST SVDs (and remains so). The RM isn't very clear about which registers are missing from FDCAN2, but we could to remove the `TT*` registers from FDCAN2 if this is important.

G4 devices have either 1 or 3 FDCANs.

Checked against:
H7 RM0399
G4 RM0440

### DIFF
--- a/devices/common_patches/fdcan/fdcan_g4.yaml
+++ b/devices/common_patches/fdcan/fdcan_g4.yaml
@@ -1,0 +1,303 @@
+# Patches for the Flexible Data Rate CAN (FDCAN) on G4
+
+_include:
+  - ./fdcan_common.yaml
+
+"FDCAN,FDCAN?":
+  RXGFC:
+    _add:
+      LSE:
+        description: List size extended
+        bitOffset: 24
+        bitWidth: 4
+      LSS:
+        description: List size standard
+        bitOffset: 16
+        bitWidth: 5
+      F0OM:
+        description: FIFO 0 operation mode
+        bitOffset: 9
+        bitWidth: 1
+      F1OM:
+        description: FIFO 1 operation mode
+        bitOffset: 8
+        bitWidth: 1
+
+  # Interrupt registers are a disaster, start over
+  IR:
+    _delete:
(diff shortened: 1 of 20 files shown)


devices/common_patches/fdcan/fdcan_g4.yaml
```diff
@@ -0,0 +1,303 @@
...
+        bitWidth: 1
+      TEFN:
+        description: Tx even FIFO new entry
+        bitOffset: 10
+        bitWidth: 1
+      TEFF:
+        description: Tx event FIFO full
+        bitOffset: 11
+        bitWidth: 1
+      REFL:
```
adamgreig, line 79:
```suggestion
      TEFL:
```

devices/common_patches/fdcan/fdcan_g4.yaml
```diff
@@ -0,0 +1,303 @@
...
+        bitWidth: 1
+      TEFNE:
+        description: Tx even FIFO new entry enable
+        bitOffset: 10
+        bitWidth: 1
+      TEFFE:
+        description: Tx event FIFO full enable
+        bitOffset: 11
+        bitWidth: 1
+      REFLE:
```
adamgreig, line 179:
```suggestion
      TEFLE:
```

adamgreig (changes requested):
Thanks, this looks great and must have been a real pain. I just found three typos, everything else is ready to merge.

richardeoin:
Thanks for the thorough review as always! It was a little tedious, but not too bad

adamgreig (approved):

adamgreig:
Thanks!

bors merge

PR #482, opened 2021-01-14
Fix RCC SWS doc typo

AlyoshaVasilieva:
RM0008 says this is correct, see "7.3.2 Clock configuration register (RCC_CFGR)".


adamgreig (approved):
Good catch, thank you!

bors merge

PR #484, opened 2021-01-25
Fix for missing SMPR field

korken89:
Fixed SMP fields docs in L4 series, and missing SMP0 field


adamgreig (approved):
Thanks!

bors merge

PR #485, opened 2021-01-31
Add CORDIC math function accelerator support to stm32g4

pawelchcki:
Add CORDIC math accellerator enum definitions to stm32g4 family    
    - modified description on mislabeled registiers
    - added clarifying descriptions for two CORDICs CSR fields

### DIFF
--- a/peripherals/cordic/cordic_g4.yaml
+++ b/peripherals/cordic/cordic_g4.yaml
@@ -7,14 +7,12 @@ CORDIC:
       _read:
         NotReady: [0, "Results from computation are not read"]
         Ready: [1, "Results are ready, this flag will be automatically cleared once value is read"]
-      _write:
-        Clear: [0, "Clear the computation results"]
     ARGSIZE:
-      Size32: [0, "Use 32 bit input values"]
-      Size16: [1, "Use 16 bit input values"]
+      Bits32: [0, "Use 32 bit input values"]
+      Bits16: [1, "Use 16 bit input values"]
     RESSIZE:
-      Size32: [0, "Use 32 bit output values"]
-      Size16: [1, "Use 16 bit output values"]
+      Bits32: [0, "Use 32 bit output values"]
+      Bits16: [1, "Use 16 bit output values"]
     NARGS:
       Num1: [0, "Only single argument write is needed for next calculation"]
       Num2: [1, "Two argument writes need to be performed for next calculation"]
@@ -31,7 +29,7 @@ CORDIC:
       Disabled: [0, "Disable interrupt request generation"]
       Enabled: [1, "Enable intterrupt request generation"]
     PRECISION: [1, 15]
-    SCALE: [0, 3]
+    SCALE: [0, 7]
     FUNC:
       Cosine: [0, "Cosine funciton"]
       Sine: [1, "Sine function"]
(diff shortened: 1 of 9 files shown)


peripherals/cordic/cordic_g4.yaml
```diff
@@ -0,0 +1,50 @@
...
+# for computing math funcitons
+
+CORDIC:
+  CSR:
+    RRDY:
+      _read:
+        NotReady: [0, "Results from computation are not read"]
+        Ready: [1, "Results are ready, this flag will be automatically cleared once value is read"]
+      _write:
+        Clear: [0, "Clear the computation results"]
```
adamgreig, line 11:
```suggestion
```

From looking at RM0440, it seems like this field is read-only, and is cleared by hardware when you read RDATA enough, rather than being clearable from software?

peripherals/cordic/cordic_g4.yaml
```diff
@@ -0,0 +1,50 @@
...
+    ARGSIZE:
+      Size32: [0, "Use 32 bit input values"]
+      Size16: [1, "Use 16 bit input values"]
+    RESSIZE:
+      Size32: [0, "Use 32 bit output values"]
+      Size16: [1, "Use 16 bit output values"]
```
adamgreig, line 17:
```suggestion
      Bits32: [0, "Use 32 bit input values"]
      Bits16: [1, "Use 16 bit input values"]
    RESSIZE:
      Bits32: [0, "Use 32 bit output values"]
      Bits16: [1, "Use 16 bit output values"]
```

We've usually used `BitsN` for this sort of enum, what do you think?

peripherals/cordic/cordic_g4.yaml
```diff
@@ -0,0 +1,50 @@
...
+      Disabled: [0, "No DMA channel writes are generated"]
+      Enabled: [1, "Write requests are generated on the DMA channel when no operation is pending"]
+    DMAREN:
+      Disabled: [0, "No DMA channel reads are generated"]
+      Enabled: [1, "Read requests are generated on the DMA channel when RRDY flag is set"]
+    IEN:
+      Disabled: [0, "Disable interrupt request generation"]
+      Enabled: [1, "Enable intterrupt request generation"]
+    PRECISION: [1, 15]
+    SCALE: [0, 3]
```
adamgreig, line 34:
```suggestion
    SCALE: [0, 7]
```

As far as I could tell SCALE can use all 3 bits, so 0 to 7.

adamgreig (changes requested):
Thanks, this looks good! I just noticed a few small changes/comments, otherwise this is ready to merge.

peripherals/cordic/cordic_g4.yaml
pawelchcki: Good catch - just double checked. My brain forgot to do `2^n-1`

peripherals/cordic/cordic_g4.yaml
pawelchcki: Yes, thanks for catching it! I wrote it while still wrapping my head around the docs and the DSL and haven't checked again.

Just checked in the docs and RRDY is indeed marked as read only.

peripherals/cordic/cordic_g4.yaml
pawelchcki: Sounds good!

pawelchcki:
Thanks @adamgreig for fast and thorough review! :bowing_man: 
Applied all suggestions. Should be ready for another review round.

adamgreig (approved):

adamgreig:
Great, thanks!

bors merge

PR #486, opened 2021-02-01
Add OPAMP enums for STM32G4

pawelchcki:
Verified with RM0440 the OpAmp register count for each chip in the family.
+ stm32g491 was missing OPAMP6

### Table from RM0440:
| Feature | STM32G474/STM32G484 | STM32G473/STM32G483 | STM32G471 | STM32G431/STM32G441 | STM32G491/STM32G4A1 |
|---|---|---|---|---|---|
| OPAMP | 6 (OPAMP1..6) | 6 (OPAMP1..6) | 4 (OPAMP1.2,3,6) | 3 (OPAMP1..3) | 4 (OPAMP1.2,3,6) |

Note that actually in STM32Cube header files stm32g471 doesn't have 4th Opamp, so leaving it off from this change - until actual hardware can be checked

### DIFF
--- a/peripherals/opamp/opamp_g4_common.yaml
+++ b/peripherals/opamp/opamp_g4_common.yaml
@@ -44,19 +44,19 @@ OPAMP:
       OutputPin: [0, "Output is connected to the output Pin"]
       ADCChannel: [1, "Output is connected internally to ADC channel"]
     OPAHSM:
-      NormalMode: [0, "OpAmp in normal mode"]
-      HighSpeedMode: [1, "OpAmp in high speed mode"]
+      Normal: [0, "OpAmp in normal mode"]
+      HighSpeed: [1, "OpAmp in high speed mode"]
     VM_SEL:
       VINM0: [0, "VINM0 connected to VINM input"]
       VINM1: [1, "VINM1 connected to VINM input"]
       PGA: [2, "Feedback resistor connected to VINM (PGA mode)"]
       Output: [3, "OpAmp output connected to VINM (Follower mode)"]
     USERTRIM:
-      FactoryTrim: [0, "Factory trim used"]
-      UserTrim: [1, "User trim used"]
+      Factory: [0, "Factory trim used"]
+      User: [1, "User trim used"]
     FORCE_VP:
-      NormalMode: [0, "Non-inverting input connected configured inputs"]
-      CalibrationVerificationMode: [1, "Non-inverting input connected to calibration reference voltage"]
+      Normal: [0, "Non-inverting input connected configured inputs"]
+      CalibrationVerification: [1, "Non-inverting input connected to calibration reference voltage"]
     OPAEN:
       Disabled: [0, "OpAmp disabled"]
       Enabled: [1, "OpAmp enabled"]
@@ -65,14 +65,14 @@ OPAMP:
       ReadWrite: [0, "TCMR is read-write"]
       ReadOnly: [1, "TCMR is read-only, can only be cleared by system reset"]
     T20CM_EN:
-      AutoSwitchDisabled: [0, "Automatic input switch triggered by TIM20 disabled"]
-      AutoSwitchEnabled: [1, "Automatic input switch triggered by TIM20 enabled"]
+      Disabled: [0, "Automatic input switch triggered by TIM20 disabled"]
+      Enabled: [1, "Automatic input switch triggered by TIM20 enabled"]
     T8CM_EN:
-      AutoSwitchDisabled: [0, "Automatic input switch triggered by TIM8 disabled"]
-      AutoSwitchEnabled: [1, "Automatic input switch triggered by TIM8 enabled"]
+      Disabled: [0, "Automatic input switch triggered by TIM8 disabled"]
+      Enabled: [1, "Automatic input switch triggered by TIM8 enabled"]
     T1CM_EN:
-      AutoSwitchDisabled: [0, "Automatic input switch triggered by TIM1 disabled"]
-      AutoSwitchEnabled: [1, "Automatic input switch triggered by TIM1 enabled"]
+      Disabled: [0, "Automatic input switch triggered by TIM1 disabled"]
+      Enabled: [1, "Automatic input switch triggered by TIM1 enabled"]
   
   OPAMP1_CSR:
     _replace_enum:
--- a/peripherals/opamp/opamp_g4_common.yaml
+++ b/peripherals/opamp/opamp_g4_common.yaml
@@ -1,0 +1,133 @@
+
+OPAMP:
+  OPAMP?_CSR:
+    LOCK:
+      ReadWrite: [0, "CSR is read-write"]
+      ReadOnly: [1, "CSR is read-only, can only be cleared by system reset"]
+    CALOUT: [0, 1]
+    TRIMOFFSETN: [0, 31] 
+    TRIMOFFSETP: [0, 31]
+    PGA_GAIN:
+      Gain2: [0, "Gain 2"]
+      Gain4: [1, "Gain 4"]
+      Gain8: [2, "Gain 8"]
+      Gain16: [3, "Gain 16"]
+      Gain32: [4, "Gain 32"]
+      Gain64: [5, "Gain 64"]
+      Gain2_InputVINM0: [8, "Gain 2, input/bias connected to VINM0 or inverting gain"]
+      Gain4_InputVINM0: [9, "Gain 4, input/bias connected to VINM0 or inverting gain"]
+      Gain8_InputVINM0: [10, "Gain 8, input/bias connected to VINM0 or inverting gain"]
+      Gain16_InputVINM0: [11, "Gain 16, input/bias connected to VINM0 or inverting gain"]
+      Gain32_InputVINM0: [12, "Gain 32, input/bias connected to VINM0 or inverting gain"]
+      Gain64_InputVINM0: [13, "Gain 64, input/bias connected to VINM0 or inverting gain"]
+      Gain2_FilteringVINM0: [16, "Gain 2, with filtering on VINM0"]
+      Gain4_FilteringVINM0: [17, "Gain 4, with filtering on VINM0"]
+      Gain8_FilteringVINM0: [18, "Gain 8, with filtering on VINM0"]
+      Gain16_FilteringVINM0: [19, "Gain 16, with filtering on VINM0"]
+      Gain32_FilteringVINM0: [20, "Gain 32, with filtering on VINM0"]
+      Gain64_FilteringVINM0: [21, "Gain 64, with filtering on VINM0"]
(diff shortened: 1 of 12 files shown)


peripherals/opamp/opamp_g4.yaml
```diff
@@ -0,0 +1,161 @@
...
+      Gain16_InputVINM0: [11, "Gain 16, input/bias connected to VINM0 or inverting gain"]
+      Gain32_InputVINM0: [12, "Gain 32, input/bias connected to VINM0 or inverting gain"]
+      Gain64_InputVINM0: [13, "Gain 64, input/bias connected to VINM0 or inverting gain"]
+      Gain2_FilteringVINM0: [16, "Gain 2, with filtering on VINM0"]
+      Gain4_FilteringVINM0: [17, "Gain 4, with filtering on VINM0"]
+      Gain8_FilteringVINM0: [18, "Gain 8, with filtering on VINM0"]
+      Gain16_FilteringVINM0: [19, "Gain 16, with filtering on VINM0"]
+      Gain32_FilteringVINM0: [20, "Gain 32, with filtering on VINM0"]
+      Gain64_FilteringVINM0: [21, "Gain 64, with filtering on VINM0"]
+      Gain2_InputVINM0FilteringVINM1: [24, "Gain 2, input/bias connected to VINM0 with filtering on VINM1 or inverting gain"]
```
pawelchcki, line 29: These enums could be customized to Describe actual Pins where VINM0 VINM1 are connected. But IMO this would make this yml bit too big, especially if all 6 opamps were to be implemented

peripherals/opamp/opamp_g4.yaml
```diff
@@ -0,0 +1,161 @@
...
+# OPAMP1 PA3   PC5   PA1   PA3   PA7   NONE  PA2   DAC3_CH1  ADC1_IN3, ADC1_IN13
+# OPAMP2 PA5   PC5   PA7   PB14  PB0   PD13  PA6   NONE      ADC2_IN3, ADC2_IN16
+# OPAMP3 PB2   PB10  PB0   PB13  PA1   NONE  PB1   DAC3_CH2  ADC3_IN1, ADC1_IN12, ADC2_IN18, ADC2_IN13
+# OPAMP4 PB10  PD8   PB13  PD11  PB11  NONE  PB12  DAC4_CH1  ADC4_IN3, ADC1_IN11, ADC5_IN5
+# OPAMP5 PB15  PA3   PB14  PD12  PC3   NONE  PA8   DAC4_CH2  ADC5_IN1, ADC_IN3
+# OPAMP6 PA1   PB1   PB12  PD9   PB13  NONE  PB11  DAC3_CH1  ADC12_IN14, ADC4_IN17, ADC3_IN17
+  OPAMP1_CSR:
+    _replace_enum:
+      VM_SEL:
+        PA3_VINM0: [0, "PA3 connected to VINM input"]
```
pawelchcki, line 88: added _VINMx suffix to help people configuring PGA gain figure out what VINMx, etc mean in this context.

Plan to work later on the hal crate implementation, but it will be a week or two until I get around to it. 
I might revisit some of these then.  For now this could be a good start

pawelchcki: Removed the VINM0 suffix to be in line with stm32f3 opamp definitions

pawelchcki:
Was looking at CubeMX headers, and it seems we're missing g491 chip and g471 indeed only has 3 OPAMPs. will update the SVD files first before proceeding with this PR

pawelchcki:
Waiting for changes from https://github.com/stm32-rs/stm32-rs/pull/489 to be merged

pawelchcki:
Now that #492 was merged. I checked the g4a1 and it seems missing the OPAMP6 as well (cross checked it with STM cube generated headers) - added it aswel.

I think this PR is finally ready for review @adamgreig

adamgreig:
Thanks for this PR! On the whole this all seems correct, my only concern is about naming the VM_SEL/VP_SEL variants after pin names instead of their generic names from the reference manual.

On the one hand, I totally get that this is probably clearer when you're writing the register, even though it's a bit unusual for the pin mapping to be fixed at the peripheral level (but makes sense for analogue peripherals like this, of course). On the other hand, it doesn't match the reference manual's description of the variants, and it means the types for each register must be difference, which precludes combining them, for example using a cluster or array, which can make HAL use much simpler (as it can be generic over all OPAMPs). _Usually_ this sort of pin mapping knowledge belongs in the HAL, not the PAC/SVD, so we'd have "VIMP0" or "VINM1" etc names instead.

I also think that maybe `_replace_enum` is not doing what you want for the VM_SEL fields, since it throws away the feedback/follower variants and just leaves the pins, iirc.

This detail sort of caught me up and hence the delay, I'm not really sure what the best way to go is. We probably can't actually cluster/array these registers for most devices anyway because of the gaps in the layout where some opamps aren't implemented, so maybe it's not an especially important detail. Any thoughts?

peripherals/opamp/opamp_g4_common.yaml
```diff
@@ -0,0 +1,133 @@
...
+      Percent90: [3, "0.9*VDDA applied to OPAMP inputs during calibration"]
+    CALON:
+      Disabled: [0, "Calibration mode disabled"]
+      Enabled: [1, "Calibration mode enabled"]
+    OPAINTOEN:
+      OutputPin: [0, "Output is connected to the output Pin"]
+      ADCChannel: [1, "Output is connected internally to ADC channel"]
+    OPAHSM:
+      NormalMode: [0, "OpAmp in normal mode"]
+      HighSpeedMode: [1, "OpAmp in high speed mode"]
```
adamgreig, line 48:
```suggestion
      Normal: [0, "OpAmp in normal mode"]
      HighSpeed: [1, "OpAmp in high speed mode"]
```

peripherals/opamp/opamp_g4_common.yaml
```diff
@@ -0,0 +1,133 @@
...
+    VM_SEL:
+      VINM0: [0, "VINM0 connected to VINM input"]
+      VINM1: [1, "VINM1 connected to VINM input"]
+      PGA: [2, "Feedback resistor connected to VINM (PGA mode)"]
+      Output: [3, "OpAmp output connected to VINM (Follower mode)"]
+    USERTRIM:
+      FactoryTrim: [0, "Factory trim used"]
+      UserTrim: [1, "User trim used"]
```
adamgreig, line 56:
```suggestion
      Factory: [0, "Factory trim used"]
      User: [1, "User trim used"]
```

peripherals/opamp/opamp_g4_common.yaml
```diff
@@ -0,0 +1,133 @@
...
+    FORCE_VP:
+      NormalMode: [0, "Non-inverting input connected configured inputs"]
+      CalibrationVerificationMode: [1, "Non-inverting input connected to calibration reference voltage"]
```
adamgreig, line 59:
```suggestion
      Normal: [0, "Non-inverting input connected configured inputs"]
      CalibrationVerification: [1, "Non-inverting input connected to calibration reference voltage"]
```

peripherals/opamp/opamp_g4_common.yaml
```diff
@@ -0,0 +1,133 @@
...
+      ReadOnly: [1, "TCMR is read-only, can only be cleared by system reset"]
+    T20CM_EN:
+      AutoSwitchDisabled: [0, "Automatic input switch triggered by TIM20 disabled"]
+      AutoSwitchEnabled: [1, "Automatic input switch triggered by TIM20 enabled"]
+    T8CM_EN:
+      AutoSwitchDisabled: [0, "Automatic input switch triggered by TIM8 disabled"]
+      AutoSwitchEnabled: [1, "Automatic input switch triggered by TIM8 enabled"]
+    T1CM_EN:
+      AutoSwitchDisabled: [0, "Automatic input switch triggered by TIM1 disabled"]
+      AutoSwitchEnabled: [1, "Automatic input switch triggered by TIM1 enabled"]
```
adamgreig, line 75:
```suggestion
      Disabled: [0, "Automatic input switch triggered by TIM20 disabled"]
      Enabled: [1, "Automatic input switch triggered by TIM20 enabled"]
    T8CM_EN:
      Disabled: [0, "Automatic input switch triggered by TIM8 disabled"]
      Enabled: [1, "Automatic input switch triggered by TIM8 enabled"]
    T1CM_EN:
      Disabled: [0, "Automatic input switch triggered by TIM1 disabled"]
      Enabled: [1, "Automatic input switch triggered by TIM1 enabled"]
```

adamgreig (commented):
Beyond the VM_SEL concern mentioned above, I had a couple other suggestions for variant names, generally just making them a bit more concise where the type of the thing is already present in the field name. Open to thoughts on these too, if you think the longer names are useful for some of the variants.

pawelchcki:
@adamgreig I don't have any strong opinions about VM_SEL et.al. I only followed other - existing implementations for this kind of things - namely: https://github.com/stm32-rs/stm32-rs/blob/master/peripherals/opamp/opamp_f3.yaml#L10

I think I see both arguments:
One is predictability and generecity, allowing maybe writing universal drivers in the future - I've little experience with that but presume there is a easier path towards that if all enum's are more or less standard and predictable and not too hardware layout specific 

Second is being able to not look at the reference manual when writing code, though probably that lays on the shoulders of hal's - as you've pointed out.

There is also 3rd argument - where maintaining these yaml's are a bit of pain - so the more concise they are the better.

So seem's like I don't have a good argument for current enum's besides opamp_f3 did first. I can work on the renames. 

/cc @Geens who added PR to stm32g4 crate

adamgreig:
For the F3, weirdly, the RM only lists the enum names as pin names, which is probably why the f3 opamps are done like that:
image

I think unless there's a strong reason from the HAL point of view to use pin names, let's go with the reference manual's VINP2 etc style. Hopefully it's not too much work to swap over. Thanks!

Geens:
@pawelchcki, thanks for keeping me in the loop!  I had a look earlier today (using the nightlies build you recommended in the HAL crate PR) and noticed opamp1's VM_SEL_A enum is the only one that has the VINM0 naming while the rest is named after the pin. Also opamp2's VP_SEL_A should have PD14 as last value (instead of PD1). All other pin-inputs where named correctly. I have not had time to check any of the other registers yet.

I can see the arguments for switching to VINM0 naming since that's how it is done in the  datasheet. I don't have a whole lot of experience in this so no strong preference either, as long as it is consistent.

pawelchcki:
Thanks @adamgreig and @Geens for taking a look. I've pushed the changes Adam suggested. But also saved the previous version to another branch - if at any point we'd want to revisit.

adamgreig (approved):
Thanks! Yep, if it turns out the other way would have been much nicer we can change it in a later release. This looks good to me for now, should be released as soon as #499 lands.

bors merge

pawelchcki:
Thanks @adamgreig! 

For reference I pushed the alternative implementation here: https://github.com/pawelchcki/stm32-rs/tree/add_opamp_descriptive_version

PR #490, opened 2021-02-05
Fix misspelled MCJDIV->MCKDIV field in SAI CR1 register in several parts

### LINKED ISSUE #471: STM32F4: SAI CR1 MCKDIV misspelled
The master clock (MCK) divider field in SAI_xCR1 is spelled MCJDIV which is most likely a typo. According to a quick google search, multiple (all?) STM32F4 variants are affected. In the manual, the field is called "MCKDIV".

wallacejohn:
This PR resolves #471. 26 parts were affected:

stm32f413
stm32f427
stm32f429
stm32f446
stm32f469
stm32f745
stm32f750
stm32f765
stm32f7x6
stm32f7x7
stm32f7x9
stm32g431
stm32g441
stm32g471
stm32g473
stm32g474
stm32g483
stm32g484
stm32l4x1
stm32l4x2
stm32l4x3
stm32l4x5
stm32l4x6
stm32l552
stm32l562
stm32wb55

This is my first PR here; I'm happy to make any changes if necessary.


adamgreig:
Thanks for this PR!

It looks like the SAI peripheral is already put into a cluster in [sai_v1.yaml](https://github.com/stm32-rs/stm32-rs/blob/master/devices/common_patches/sai/sai_v1.yaml#L4), so there's no need to make a new file to re-create the cluster. The patching syntax is a bit confusing: the `_cluster` spec you've written out is an instruction to not only change that MCJDIV field, but also to group the specified registers together into a channel cluster. Which is a good outcome but we already do it in another patch file, so it doesn't make sense to duplicate it.

Instead, could you put the `_modify` part inside the existing sai_v1.yaml file? You could then also add that existing file to all the parts you've added your new file to, which didn't already have the SAI patches applied.

wallacejohn:
Thanks for your feedback! I incorporated your suggestions into a new commit.

adamgreig (approved):
Perfect, thanks!

bors merge

PR #491, opened 2021-02-05
Fixed ch17 and ch18 in ADC COMMON for STM32L4x5

korken89:
I noticed when extending the STM32L4xx HAL that this MCU had some bits with different names compared to the rest of the series.
This PR fixes this.

### DIFF
--- a/devices/stm32l4x5.yaml
+++ b/devices/stm32l4x5.yaml
@@ -6,6 +6,16 @@ _modify:
   ADC123_Common:
     name: ADC_Common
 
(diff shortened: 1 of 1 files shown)


korken89:
Ready for review!

adamgreig (approved):
Thanks! Annoying that ST aren't more consistent here, sometimes the fields are TSEN/VBATEN on other devices, but CH17SEL/CH18SEL clearly matches the RM and the other L4, so let's go with this.

bors merge

PR #489, opened 2021-02-03
Add devices/stm32g491.yaml

pawelchcki:
This is a companion PR to https://github.com/stm32-rs/stm32-rs/pull/488 which should be merged first.

Otherwise the CI mmap checker breaks, on missing vendor svd file for `stm32g491` in master

PR #488, opened 2021-02-02
Update stm32g4 svd add missing stm32g491

pawelchcki:
Update stm32g4 svd zip to newest download from STM site
    
New SVD ZIP contains fixes and missing svd file for stm32g491 chip
```    
    $ curl -s -L -o- https://www.st.com/resource/en/svd/stm32g4_svd.zip | sha1sum
    d9ad831ef1001a1760f90a8ec6456f824c25525f  -
```

PR #492, opened 2021-02-06
G4 SVD update and add G491, G4A1

adamgreig:
This PR includes #488 and #489's commits, plus a new more to add the 4A1 device and remove some now-redundant patch files that the updated SVDs address.

The only significant change with the new SVDs is that the ADC common register block (CCR etc) had its position corrected from offset 0x200 to offset 0x300 on some devices.


pawelchcki:
Thanks a ton @adamgreig ! And great eye for detail! I completely missed the 4A1 and opportunity to delete some code for the updated ADCs

PR #495, opened 2021-02-12
Merge f3x8 to f303

Piroro-hs:


adamgreig (approved):
Thanks for this PR!

For others' reference: f303 and f3x8 share the same RM0316 and thus in general the same registers, so it makes sense to only maintain one device. The naming could be clearer but our part table should help people find the right thing.

I've checked the detailed memory map differences between the new device and the old 3x8 and there are a number of differences, but almost all of them are fields missing from the 3x8 device or typos in its existing descriptions. A few other errors include extra timers in the 3x8 that don't exist in the reference manual (and so are presumably mistakes as well). Apologies in advance to anyone who discovers any further discrepancies...

It does seem like 303 is missing DAC2, though. I'll make any changes we need before releasing soon.

bors merge

PR #498, opened 2021-02-13
Add DSI host global interrupt for STM32F469

### LINKED ISSUE #487: STM32F469 seem to miss the DSI global IRQ
In the `stm32f469xx.h` header file it is defined as:
```
DSI_IRQn = 92      /*!< DSI global Interrupt */
```
But it does not seem to be included here. How would I go about including it?

wallacejohn:
This PR closes #487. F469/79 is the only device affected in the F4 family.


adamgreig (approved):
Thanks!

bors merge

PR #499, opened 2021-02-13
H7B3: Split DAC1/2 REC, remove LPTIM45

mattico:


mattico:
Hmm, removing lptims didn't work.

devices/stm32h7b3.yaml
```diff
@@ -18,6 +18,67 @@ _modify:
...
+      DAC12RST:
+        name: DAC1RST
+        description: DAC1 (containing two converters) reset
+  APB1LENR:
+    _modfy:
+      DAC12EN:
+        name: DAC1EN
+        description: DAC1 (containing two converters) peripheral clock enable
+  APB1LLPENR:
+    _modfy:
```
adamgreig, line 50:
```suggestion
    _modify:
      DAC12RST:
        name: DAC1RST
        description: DAC1 (containing two converters) reset
  APB1LENR:
    _modify:
      DAC12EN:
        name: DAC1EN
        description: DAC1 (containing two converters) peripheral clock enable
  APB1LLPENR:
    _modify:
```

adamgreig:
> Hmm, removing lptims didn't work.

This is an annoying problem due to using YAML, but basically you have duplicated the RCC key, and only the second instance survives (the merging we do for included files can't work inside a single file). You also can't just combine the two RCC elements as the APB4* registers would then also be defined twice and only the second would survive; you need to move the `_delete` blocks inside the existing APB4* register blocks instead.

adamgreig (approved):
Thanks! Looks good now. From a quick check it seems the EXTI TODO is just some enumerated values, which is a bit fiddly to remove and I don't think is a major concern for right now, so I'm going to get this merged and the release done.

I noticed the resetValues weren't taking effect because they have to be specified in a higher-level `_modify` block (not alongside fields), and also they are case sensitive (must be `resetValue`) and _also_ it turns out they need to be specified with the names prior to renaming those registers because of how the merges work. We could probably do with better error/warning diagnostics around this sort of thing...

bors merge

mattico:
Thanks for helping to get this in before the release!

PR #502, opened 2021-02-19
WWDG prescaler fix for H7

### LINKED ISSUE #501: STM32H753 has 3 bits of wwdg prescaler while only 2 are in the pac
Datasheet of the H742/H743/H753/H750:
image

Config: https://github.com/stm32-rs/stm32-rs/blob/master/peripherals/wwdg/wwdg.yaml

This has been mentioned in an H7 HAL PR comment, but is seems it wasn't fixed: https://github.com/stm32-rs/stm32h7xx-hal/pull/10/files#r294054557

I don't know how this crate works, so if somebody can instruct me with the rough steps, I could do it.

diondokter:
Fixes #501 

- Renamed the G4 wwdg to wwdg_v2
- Created patch for the bit width
- Added the patch to the H7 devices (I checked the datasheets and they all have the bigger prescaler)

### DIFF
--- a/devices/common_patches/h7_wwdg.yaml
+++ b/devices/common_patches/h7_wwdg.yaml
@@ -1,6 +1,5 @@
 WWDG:
-  CFG:
+  CFR:
     _modify:
       WDGTB:
         bitWidth: 3
-        
--- a/devices/common_patches/h7_wwdg.yaml
+++ b/devices/common_patches/h7_wwdg.yaml
@@ -1,4 +1,4 @@
-WWDG:
+"WWDG,WWDG?":
   CFR:
     _modify:
       WDGTB:
--- a/devices/common_patches/h7_wwdg.yaml
+++ b/devices/common_patches/h7_wwdg.yaml
@@ -1,0 +1,6 @@
+WWDG:
+  CFG:
+    _modify:
+      WDGTB:
+        bitWidth: 3
+        
--- a/devices/common_patches/h7_wwdg1.yaml
+++ b/devices/common_patches/h7_wwdg1.yaml
@@ -1,0 +1,5 @@
+WWDG1:
+  CFR:
+    _modify:
+      WDGTB:
+        bitWidth: 3
--- a/devices/common_patches/h7_wwdg1.yaml
+++ b/devices/common_patches/h7_wwdg1.yaml
@@ -1,5 +1,0 @@
-WWDG1:
-  CFR:
-    _modify:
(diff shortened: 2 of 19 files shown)


diondokter:
`svdtools.patch.MissingPeripheralError: Could not find WWDG`
How do I solve this?

adamgreig:
Ah, looks like it's `CFR` not `CFG`, a typo in my suggestion earlier. https://stm32-rs.github.io/stm32-rs/STM32H7x5_CM7.html#WWDG1:CFR

diondokter:
Ah, I should have caught that. The G4 doesn't have a patch for this, so I guess the SVD was already correct in that regard.

diondokter:
Strange... The original svd file says this:
```yaml
...
    <peripheral>
      <name>WWDG</name>
      <description>WWDG</description>
      <groupName>WWDG</groupName>
      <baseAddress>0x50003000</baseAddress>
      <addressBlock>
        <offset>0x0</offset>
        <size>0x400</size>
        <usage>registers</usage>
      </addressBlock>
...
```

Is it renamed somewhere?

adamgreig:
I think you should be able to just have a single h7_wwdg.yaml and write it like:

```yaml
"WWDG,WWDG?":
    ...
```

(which is what the G4 WWDG file does: https://github.com/stm32-rs/stm32-rs/blob/012135527a3cfe215e6d00dda4595603084df565/peripherals/wwdg/wwdg_v2.yaml#L7)

diondokter:
> 
> 
> I think you should be able to just have a single h7_wwdg.yaml and write it like:
> 
> ```yaml
> "WWDG,WWDG?":
>     ...
> ```
> 
> (which is what the G4 WWDG file does:
> 
> https://github.com/stm32-rs/stm32-rs/blob/012135527a3cfe215e6d00dda4595603084df565/peripherals/wwdg/wwdg_v2.yaml#L7
> )

But will that change the bitWidth property as well? That seems important to change...

adamgreig:
I mean, in the `devices/common_patches/h7_wwdg.yaml`, which changes the `bitWidth`, you can match against all spellings of WWDG, rather than having `h7_wwdg.yaml` and a separate `h7_wwdg1.yaml`, and that way also match on WWDG2 (which some H7s have).

adamgreig (approved):
Looks good, thanks!

bors merge

diondokter:
Awesome! Thanks for you patience and help :)

PR #506, opened 2021-03-01
Feature/stm32 g4

kevswims:
With these changes I was able to port the USB CDC example from the F3 HAL to the G4.

### DIFF
--- a/devices/stm32g473.yaml
+++ b/devices/stm32g473.yaml
@@ -19,3 +19,5 @@ _include:
  - ../peripherals/opamp/opamp_g4_opamp4_5.yaml 
  - ../peripherals/opamp/opamp_g4_opamp6.yaml 
  - ../peripherals/i2c/i2c_v2.yaml
+ - ./common_patches/rename_USB_FS_peripheral_to_USB.yaml
+ - ./common_patches/g4_usb.yaml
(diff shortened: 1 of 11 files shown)


devices/stm32g473.yaml
```diff
@@ -18,3 +18,5 @@ _include:
  - ../peripherals/opamp/opamp_g4_common.yaml 
  - ../peripherals/opamp/opamp_g4_opamp4_5.yaml 
  - ../peripherals/opamp/opamp_g4_opamp6.yaml 
+ - ./common_patches/rename_USB_FS_peripheral_to_USB.yaml
+ - ./common_patches/g4_usb.yaml
```
kevswims, line 22: I believe these are probably needed for other G4 series as well but I do not have a good way to test those changes. Should they be added?

adamgreig:
Thanks for this PR!

The new BCDR register is good, but it looks like including the `rename_USB_FS_peripheral_to_USB.yaml` file isn't working, because in this SVD it's called `USB_FS_device` instead of `USB_FS`. I think just modifying that file to match against `"USB_FS*"` instead should do what you want, or you could create a new version to specifically match `USB_FS_device`.

kevswims:
I'll take a look at that when I'm back from vacation. I thought I had to have that change to get my hal port from the f3 working but I might have changed it in the hal instead or messed something else up along the way.

kevswims:
@adamgreig I updated this to use the wildcard for the USB renaming. I modified the HAL changes I have for this on the G473 and tested that the USB works.

adamgreig (approved):
Looks good, thank you!

bors merge

PR #507, opened 2021-03-02
STM32WL: svd updates

ofauchon:
Updated SVD(s) for STM32WL5x (dual core CM4/CM0) and STM32WLE5 (Single core CM4) (18/01/21 update on st.com)


adamgreig (approved):
Thanks for this PR! I've modified it slightly to keep the existing WLE5 part with the same name, since it only has one core anyway, and added the new parts to the parts table.

bors merge

Issue #504, opened 2021-02-25
STM32U5


jonathanpallant:
Cortex-M33, ultra lower power device. New out today. Let's get to it!

https://www.st.com/en/microcontrollers-microprocessors/stm32u5-series.html

jonathanpallant:
Ah, main availability won't be until September and it looks like CubeMX (and hence the packs containing the SVD files) won't be around until then.

adamgreig:
Update from September: the GitHub repo exists and has C headers, but no sign of an SVD file there or on st.com yet.

milesgranger:
Seems like they heard you talking about it; [available here](https://www.st.com/en/microcontrollers-microprocessors/stm32u5-series.html#cad-resources), uploaded on the same day. :sweat_smile:

jerabaul29:
Just being curious: what do you think is a typical / reasonable timeline for a dedicated crate for this MCU? :) .

jonathanpallant:
I have no need for the chip now, so I won't be writing anything (unless I'm paid to do so).

jerabaul29:
The board is indeed in stock at Mouser for example at the time of writing this :) : https://no.mouser.com/ProductDetail/STMicroelectronics/NUCLEO-U575ZI-Q?qs=Wj%2FVkw3K%252BMD4hdUJJxGixg%3D%3D .

Any update? Being curious / n00b: once the pull request from @Urhengulas is merged, does it mean that there will be full support from Rust for all functionalities, or will this be only basic support, with more work needed to get different peripherals / advanced functions to work? :)

Urhengulas:
> Any update? Being curious / n00b: once the pull request from @Urhengulas is merged, does it mean that there will be full support from Rust for all functionalities, or will this be only basic support, with more work needed to get different peripherals / advanced functions to work? :)

As far as I understand (which might be wrong) this repository only generates "device support crates" aka. "peripheral access crates" (PAC). After my PR lands we still needs to be build the "hardware abstraction layer" (HAL), for it to be actually useful.

jerabaul29:
Aaah, ok, many thanks for the explanation, and I guess the HAL is the more labor intensive part, right? Do you know if there are plans / ongoing work for a dedicated HAL for this chip once your PAC PR lands? :)

Urhengulas:
[...]

Issue #477, opened 2021-01-10
STM32H725 / STM32H735 / STM32H730 support


richardeoin:
Discussion here: https://github.com/stm32-rs/stm32h7xx-hal/issues/163

richardeoin:
Closed by #554

PR #508, opened 2021-03-05
Add comprehensive CEC support

wallacejohn:
This PR adds comprehensive CEC support for all devices with CEC IP and serves as a proposal for an RM-based/peripheral-based patch approach that can coexist with the current patching methodologies. This method is conducive to decreased dependence on SVD files (like stm32-rs/meta#4 proposes) and faster support for new devices in the future (#477 and #504). CEC support is added for the following RMs/devices (including devices that do not yet have a patch file and/or SVD file):

- RM0041: F100
- RM0091: F042, F048, F051, F058, F071, F072, F078, F091, F098
- RM0313: F373, F378
- RM0385: F745, F746, F750, F756
- RM0390: F446
- RM0399: H745, H747, H755, H757
- RM0410: F765, F767, F769, F777, F778, F779
- RM0433: H742, H743, H750, H753
- RM0436: MP157
- RM0441: MP151
- RM0442: MP153
- RM0444: G071, G081
- RM0455: H7A3, H7B0, H7B3
- RM0468: H723, H725, H730, H733, H735

The goal of this approach is to create a self-contained definition of the CEC peripheral that can replace the sometimes-erroneous SVD definition and unify the register/field descriptions for all SVDs. The basic idea is that there is a unique YAML file in common_patches/cec for each reference manual that creates a new CEC peripheral to match that RM's description. The register and field definitions are imported from other YAML files that define the IP common to different RMs. A separate cec_delete.yaml file removes the original SVD's CEC definition. Some SVDs have a CEC definition when the device itself doesn't actually have CEC IP, so the separate delete file allows easy removal. If stm32-rs is no longer dependent on SVD files at some point in the future, then the delete file can just go away. Some code duplication could be reduced with new features in svdtools, but I haven't looked at this in-depth yet.

One aspect that doesn't really fit into this peripheral-based approach is the peripheral reset and enable fields in the RCC registers. Most SVDs already had them, but F446 needed CECRST added. F730, F7x2, and F7x3 needed CECRST removed. These were just done in the top device YAML file in the absence of a better idea.

The CEC peripheral is pretty straightforward with only two IP versions (v1 for F100 and v2 for all others), so some of the design decisions aren't applicable here but make more sense for more complicated peripherals (like the DFSDM, which I'm also working on).

I don't have a use case for the CEC or an easy way to test it, so I can't speak for the accuracy of this patch with respect to the actual silicon. However, I've reviewed the RMs and mmap diffs very thoroughly and am pretty confident in the accuracy of this patch. There are some notable differences that show up in the mmap diffs besides basic name and description changes:

- stm32f0x1: The original SVD's CFGR register is messed up
- stm32f0x2: Same
- stm32f0x8: Same
- stm32f301: Completely removed the CEC peripheral
- stm32f373: Same as f0xx above
- stm32f446: The original SVD is missing the CEC interrupt and CECRST
- stm32f730: Removed CECRST
- stm32f7x2: Same
- stm32f7x3: Same

I'd appreciate any feedback or suggestions for improvement on this approach, as well as any errors that I missed (or added). Absent any problems, this should be ready to merge.

### DIFF
--- a/devices/common_patches/cec/cec_RM0041.yaml
+++ b/devices/common_patches/cec/cec_RM0041.yaml
@@ -1,7 +1,3 @@
-_delete:
-  - CEC
-  - HDMI_CEC
-
 _add:
   CEC:
     description: "HDMI-CEC controller"
--- a/devices/common_patches/cec/cec_RM0041.yaml
+++ b/devices/common_patches/cec/cec_RM0041.yaml
@@ -1,0 +1,21 @@
+_delete:
+  - CEC
+  - HDMI_CEC
+
+_add:
+  CEC:
+    description: "HDMI-CEC controller"
+    baseAddress: 0x40007800
+    addressBlock:
+      offset: 0x0
+      size: 0x400
+      usage: "registers"
+    interrupts:
+      CEC:
+        description: "HDMI-CEC global interrupt"
+        value: 42
+    registers: {}
+
+CEC:
+  _include:
+    - registers/cec_v1_registers_common.yaml
--- a/devices/common_patches/cec/cec_RM0091.yaml
+++ b/devices/common_patches/cec/cec_RM0091.yaml
@@ -1,5 +1,6 @@
 _delete:
   - CEC
+  - HDMI_CEC
 
 _add:
   CEC:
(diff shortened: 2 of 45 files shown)


wallacejohn:
Per the discussion on Matrix, this approach would benefit from dedicated directives in svdtools. I can put together a proposal for some additional syntax.

PR #510, opened 2021-03-07
i2C register definitions for stm32g4xx

Geens:
I added register definitions for the I2C peripheral for the g4 series. First time I'm doing this, so feedback is much appreciated.

I noticed the module is named i2c1, is there a way to change this to i2c, without the 1 at the end? Or is there a reason it's there?


adamgreig:
Thanks for this PR!

Your new descriptions look generally good with a few style points:

* We use capital letters for the variant names, especially for example `Disabled`, `Enabled`, and `Clear`
* We use camel-case for the variant names, for example instead of `set` and `not_set` we'd have `Empty` and `NotEmpty`

However, when I checked, it looks like the existing file `peripherals/i2c/i2c_v2.yaml` (mostly with descriptions in the included `i2c_v2_base.yaml` file) already covers the G4's I2C peripheral, it just hasn't been included in the G4 device files yet. I used the `scripts/matchperipherals.py` script to check the existing YAML applies cleanly, and also quickly diffed it with your new file to check there were no other differences. This sucks after you've put in all the effort to write out new descriptions, so I'm sorry! But if the existing file does match this peripheral, we should use it, so that it's consistent with other devices.

> I noticed the module is named i2c1, is there a way to change this to i2c, without the 1 at the end? Or is there a reason it's there?

The module gets named after whichever is the 'base' peripheral from which the others derive, so in this case I2C1. There's no way in svd2rust to tell it what name to use for the module without also renaming the peripheral (which we can't do, since it needs to be I2C1 since there's also I2C2, 3, etc).

Geens:
I have no idea how I missed this. I really thought I checked.

I'll remember the style guide and other advice for next time and consider this a learning exercise. Thanks you for the explanation!

I removed the g4 file and updated the g4 devices to use the v2 file.

adamgreig (approved):
Great, thank you again!

bors merge

PR #514, opened 2021-03-14
Update STM32G0xx svd files

tachiniererin:
This adds the 50, 51, 61, B0, B1 and C1 variants and updates the existing files. ST still hasn't published the official SVD zip on their website, but thankfully the files provided by STM32CubeIDE come with an Apache 2.0 license header.

The changes in the existing files are minimal and seem to mostly be fixes and improvements in the register descriptions.


adamgreig:
Thanks for this update! I'm glad more SVDs are becoming available under the Apache 2 licence, I plan to do some consolidation work soon to make the licensing situation clearer.

The CI failure for compare_mmaps is expected because of new devices being added, so don't worry about that one.

I noticed in this update that the STM32G070 and 081 SVDs still have the weird error in the DMA ISR and IFCR registers, which you've removed the existing fix for. Perhaps it just needs to apply to a different set of devices? The DMA clustering also seems to be broken for the 070 and 081. The 071 has ISR correct but IFCR still wrong, and also fails to cluster.

tachiniererin:
I missed the first time that the whole DMA controller seems to be missing from the SVD files for B{0,1} and C1, also they have one 7ch and one 5ch controller. I reverted the changes which broke the existing parts and wrote the new DMA controller definitions from scratch.
I hope i didn't miss anything again, staring at svd files for hours gives me a bit tunnel vision :/

adamgreig (approved):
Thanks again for the PR! Sorry it took so long to merge, I thought it would take longer to review based on the size but actually it wasn't so bad. All looks correct to me.

We need to add these new devices to the `stm32_part_table.yaml` file but I'll do that separately, this PR has waited long enough!

bors merge

PR #517, opened 2021-03-18
Fix bit shift in AHB1 register for CRC enable/reset/sleep-mode bits, STM32L4x3.svd patch 

cyrusmetcalf:
Discovered when attempting to enable CRC hardware module.
```rust
    let crc_guy = dp.CRC.constrain(&mut rcc.ahb1);
```
CRC output would be 0. Found with debugger that the enable bit was not
being set in rcc.ahb1 power enable register. Other peripherals in the
register (DMA1/2) could be enabled and the remaining bits in the
register are not writable.

Tested against the following code provided by STM32CubeMX, in which the
CRC module is functional.

## Sources 
 - CMSIS/Device/ST/STM32L4xx/Include/stm32l433xx.h
 - STM32L43xxx... Reference Manual RM0394 Rev 4. Sections 6.4.9, 6.4.15, 6.4.21

 - Bug Verified/Fixed/Tested on STM32L433 (Nucleo-L433RC-P)

### DIFF
--- a/devices/stm32l4x3.yaml
+++ b/devices/stm32l4x3.yaml
@@ -12,19 +12,19 @@ RCC:
         name: USBFSEN
         description: USB FS clock enable
         
-    # SVD incorrectly shifts CRCEN/CRCRST/CRCSMEN  11 bits instead of 12
-    AHB1ENR:
-        _modify:
-            CRCEN:
-                bitOffset: 12
-    AHB1RSTR:
-        _modify:
-            CRCRST:
-                bitOffset: 12
-    AHB1SMENR:
-        _modify:
-            CRCSMEN:
-                bitOffset: 12
+  # SVD incorrectly shifts CRCEN/CRCRST/CRCSMEN  11 bits instead of 12
+  AHB1ENR:
+    _modify:
+      CRCEN:
+        bitOffset: 12
+  AHB1RSTR:
+    _modify:
+      CRCRST:
+        bitOffset: 12
+  AHB1SMENR:
+    _modify:
+      CRCSMEN:
+        bitOffset: 12
 
 _modify:
   # The SVD calls ADC1 ADC.
--- a/devices/stm32l4x3.yaml
+++ b/devices/stm32l4x3.yaml
@@ -12,8 +12,7 @@ RCC:
         name: USBFSEN
         description: USB FS clock enable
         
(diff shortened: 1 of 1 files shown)


devices/stm32l4x3.yaml
```diff
@@ -11,6 +11,21 @@ RCC:
       USBF:
         name: USBFSEN
         description: USB FS clock enable
+        
+# SVD incorrectly shifts CRCEN/CRCRST/CRCSMEN  11 bits instead of 12
+RCC:
```
adamgreig, line 16: Currently this duplicates the RCC key in the YAML, so the previous patch doesn't get applied, which knocks out the USFB -> USBFSEN fix. Just need to remove `RCC:` to fix it up.

```suggestion
        
    # SVD incorrectly shifts CRCEN/CRCRST/CRCSMEN  11 bits instead of 12
```

adamgreig (changes requested):
Thanks for this PR! The changes look good, just need to fix a syntax problem in the YAML (below).

devices/stm32l4x3.yaml
```diff
@@ -11,6 +11,20 @@ RCC:
...
+            CRCEN:
+                bitOffset: 12
+    AHB1RSTR:
+        _modify:
+            CRCRST:
+                bitOffset: 12
+    AHB1SMENR:
+        _modify:
+            CRCSMEN:
+                bitOffset: 12
```
adamgreig, line 27:
```suggestion
  # SVD incorrectly shifts CRCEN/CRCRST/CRCSMEN  11 bits instead of 12
  AHB1ENR:
    _modify:
      CRCEN:
        bitOffset: 12
  AHB1RSTR:
    _modify:
      CRCRST:
        bitOffset: 12
  AHB1SMENR:
    _modify:
      CRCSMEN:
        bitOffset: 12
```

adamgreig (changes requested):
Sorry, looks like my indentation wasn't quite right in that suggested change - the new content had been indented with 4 spaces per tab, so when we combined the sections they were indented underneath the existing APB1ENR1 register and that's why the CI failed.

devices/stm32l4x3.yaml
cyrusmetcalf: Awesome!  Thanks for the help with the fix

adamgreig (approved):
Thank you, looks good!

bors merge

PR #518, opened 2021-03-21
Fix TIM15 CCR2 address offset for STM32L5xx

### LINKED ISSUE #511: L5 TIM15 SVD error
As @adamgreig pointed out in Matrix, TIM15 on the L5 SVD has an error that should be manually corrected: The TIM15 `CCR2` offset is listed at `0x28`, when it should be `0x38`. `0x28` is the PSC register. Neither registers work as a result.

Is the way to fix this PRing [this file](https://github.com/stm32-rs/stm32-rs-nightlies/blob/master/stm32l5/src/stm32l552/tim15.rs#L22) in the `stm32-nightlies` repo?

wallacejohn:
This PR closes #511. TIM15 on L552 and L562 are the only timers affected by this issue.

### DIFF
--- a/devices/stm32l552.yaml
+++ b/devices/stm32l552.yaml
@@ -58,6 +58,11 @@ TIM7:
         description: TIM7 global interrupt
         value: 50
 
+TIM15:
+  _modify:
+    CCR2:
+      addressOffset: 0x38
+
 _include:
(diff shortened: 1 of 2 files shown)


adamgreig (approved):
Thanks!

bors merge

Issue #515, opened 2021-03-16
Are GigaDevice GD32 Cortex-M MCUs in scope for this repository?


qwandor:
Hello,
I'm interested in using Rust on a GD32F130 series microcontroller. As far as I can see there are currently no device or HAL crates to support it. However, the GD32F families seem to have peripherals and register layouts mostly compatible with the corresponding STM32 families, apart from some differences in things like clock speed or number of USARTs, so it seems likely that a lot of the existing code from the crates in this repository and the HAL crates could be reused.

Given this, would it make sense to work on supporting them within this same repository? For example, to support the GD32F130 with a feature on the `stm32f1` and `stm32f1xx-hal` crates.

For reference, https://www.gigadevice.com/wp-content/uploads/2021/03/2021-GigaDevice-Selection-Guide_Final.pdf has an overview of the device families, and http://www.gd32mcu.com/en/download/ has datasheets (and SVD files within the addons e.g. GigaDevice.GD32F1x0_DFP.3.1.0.pack).

adamgreig:
Closing, see https://github.com/stm32-rs/stm32-rs/pull/520#issuecomment-805302011 for rationale.

qwandor:
For reference in case anyone else finds this issue, I ended up creating a new repository for this, based on the infrastructure of this one: https://github.com/qwandor/gd32-rs

PR #520, opened 2021-03-21
Add support for GD32F1x0

qwandor:
As mentioned in #515, most of the GD32 series have very similar peripherals to the various STM32 microcontrollers. Unfortunately, the register and field names are not always the same, so the peripheral patches don't quite apply.

This PR adds the GD32F1x0 family, with support for some of its peripherals. More peripherals will follow.

Thoughts? Does it makes sense to put it in this repository? Sharing the rest of the infrastructure certainly seems to make sense.


adamgreig:
Thanks for this PR, and sorry for not responding to #515 earlier. I've thought about this carefully over the weekend and decided to keep this repository as STM32 only. As much as the GD32 devices are extremely similar, they're still different devices from a different manufacturer, and I don't want to extend the scope of this repository that far. You're of course extremely welcome to copy all the tooling (and applicable YAML files, patches, etc etc) from here, which should hopefully make it straightforward to build support for them elsewhere, and I'd be happy to answer questions about how it goes together too. Sorry this isn't the answer you wanted!

PR #528, opened 2021-03-28
Added some new enumerations for G4's RCC

timblakely:
Heya! 

Adding a few new enum fields to the G4 RCC. Let me know if there's a better place to organize/put these. 

Note: I only have access to the STM32G474, but I suspect that these enumerations are common across the G4 family seeing as the only differences appear to be RAM/ROM; they all share the same max frequency AFAICT.

### DIFF
--- a/peripherals/rcc/rcc_g4.yaml
+++ b/peripherals/rcc/rcc_g4.yaml
@@ -1,0 +1,334 @@
+RCC:
+  CR:
+    CSSON:
+      "Off": [0, "Clock security system disabled (clock detector OFF)"]
+      "On": [1, "Clock security system enable (clock detector ON if the HSE is ready, OFF if not)"]
+    HSEBYP:
+      NotBypassed: [0, "HSE crystal oscillator not bypassed"]
+      Bypassed: [1, "HSE crystal oscillator bypassed with external clock"]
+    "*RDY":
+      _read:
+        NotReady: [0, "Clock not ready"]
+        Ready: [1, "Clock ready"]
+    HSION,HSEON,PLLON,PLLI2SON,PLLSAION:
+      "Off": [0, "Clock Off"]
+      "On": [1, "Clock On"]
+  
+  CFGR:
+    MCOPRE:
+      Div1: [0, "MCO divided by 1"]
+      Div2: [1, "MCO divided by 2"]
+      Div4: [2, "MCO divided by 4"]
+      Div8: [3, "MCO divided by 8"]
+      Div16: [4, "MCO divided by 16"]
+    MCOSEL:
+      None: [0, "MCO output disabled, no clock on MCO"]
+      SYSCLK: [1, "SYSCLK system clock selected"]
+      MSI: [2, "MSI clock selected"]
+      HSI: [3, "HSI clock selected"]
+      HSE: [4, "HSE clock selected"]
+      PLL: [5, "Main PLL clock selected"]
+      LSI: [6, "LSI clock selected"]
+      LSE: [7, "LSE clock selected"]
+      HSI48: [8, "Internal HSI48 clock selected"]
+    PPRE*:
+      Div1: [0, "HCLK not divided"] # Same for [0, 8]
+      Div2: [4, "HCLK divided by 2"]
+      Div4: [5, "HCLK divided by 4"]
+      Div8: [6, "HCLK divided by 8"]
+      Div16: [7, "HCLK divided by 16"]
+    HPRE:
+      Div1: [0, "SYSCLK not divided"] # Same for [0, 7]
+      Div2: [8, "SYSCLK divided by 2"]
+      Div4: [9, "SYSCLK divided by 4"]
+      Div8: [10, "SYSCLK divided by 8"]
+      Div16: [11, "SYSCLK divided by 16"]
+      Div64: [12, "SYSCLK divided by 64"]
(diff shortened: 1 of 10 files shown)


adamgreig:
Hi, thanks for the PR!

Your changes look good, but we try to keep describing field variants (this sort of change) to the `peripherals/` directory, rather than inside `common_patches`. In fact, it looks like all/most of the existing [rcc_v3](https://github.com/stm32-rs/stm32-rs/blob/master/peripherals/rcc/rcc_v3.yaml) (and [rcc_v3_pll](https://github.com/stm32-rs/stm32-rs/blob/master/peripherals/rcc/rcc_v3_pll.yaml)) might apply although it was originally written for the H7. It looks like you might have a few extra things not covered there, though. Perhaps you could add those (to a new file if necessary) and then include the rcc_v3 in the existing stm32g4*.yaml devices? Or if they don't quite merge, you could create a new `peripherals/rcc/rcc_g4.yaml` with just your changes from this PR so far, but hopefully they're similar enough that we can mostly share one definition.

Please shout if anything doesn't make sense, I know it's not the easiest thing to understand from scratch.

timblakely:
Gotcha; understood about the division of field variants vs field fixes (kinda sad that ST makes us fix so much, ya? :). Though I haven't gone through the H7 reference manual just yet, I did check the clock config on the STM32CubeMX and compared it to the G4's. It seems like there's enough difference - particularly in the PLL domain - that I'd be concerned with me breaking any existing H7 users by changing field names and/or APIs (divisor enums vs ranges, etc) if I were to modify the `rcc_v3(_pll)?` files.

I've moved the changes into a new `rcc_g4.yaml` file under `peripherals/`. Since the RCC is common to all g4 chips it's now `include:`'d it from the `common_patches/g4_rcc.yaml` file, though do let me know if that's not the right place to include it!

adamgreig:
Thanks! Ok, fair enough on the H7 one, no problem having a new rcc_g4.yaml.

I'll try to get this fully reviewed at the weekend, but in the meantime could you move the include to be done from each device file instead of from the common_patches file? It just helps make it easier to find out where files are being included if there aren't too many layers of indirection, especially across the "patch"/"peripheral" split.

timblakely:
Sounds good! No hurry on the review; just have it workin' in my own local repo so I figured I'd share it with the rest of the community :)

peripherals/rcc/rcc_g4.yaml
```diff
@@ -0,0 +1,381 @@
...
+      Div15: [14, "pll_p_ck = vco_ck / 15"]
+      Div16: [15, "pll_p_ck = vco_ck / 16"]
+    PLLSRC:
+      None: [0, "No clock sent to PLL"]
+      HSI16: [2, "No clock sent to PLL"]
+      HSE: [3, "No clock sent to PLL"]
+  
+  BDCR:
+    LSCOSEL:
+      LSI: [0, LSI clock selected"]
```
adamgreig, line 253:
```suggestion
      LSI: [0, "LSI clock selected"]
```

peripherals/rcc/rcc_g4.yaml
```diff
@@ -0,0 +1,381 @@
...
+    LSEON:
+      "Off": [0, "LSE oscillator Off"]
+      "On": [1, "LSE oscillator On"]
+
+  CSR:
+    _modify:
+      WDGRSTF:
+        name: IWDGRSTF
+      PADRSTF:
+        name: PINRSTF
```
adamgreig, line 304: These changes really belong in the `devices/common_patches/g4_rcc.yaml` file instead.

peripherals/rcc/rcc_g4.yaml
```diff
@@ -0,0 +1,381 @@
...
+      _write:
+        Clear: [1, "Clears the reset flag"]
+    LSIRDY:
+      _read:
+        NotReady: [0, "LSI oscillator not ready"]
+        Ready: [1, "LSI oscillator ready"]
+    LSION:
+      "Off": [0, "LSI oscillator Off"]
+      "On": [1, "LSI oscillator On"]
+
```
adamgreig, line 335:
```suggestion
```

This section seems to be duplicated?

adamgreig (changes requested):
Thanks again for the PR! I've fully reviewed it and it basically all looks good to me, I just had three minor comments then we're good to merge:

peripherals/rcc/rcc_g4.yaml
timblakely: Done!

peripherals/rcc/rcc_g4.yaml
timblakely: Moved!

peripherals/rcc/rcc_g4.yaml
timblakely: Oops, you're right! Removed.

timblakely:
Oof, sorry for the delay, been a nutty sprint. I think I got all the requested changes, but please do double-check; was trying out the VSCode Pull Request extension and things went sideways in a hurry 😅

timblakely:
I've got a few more sequential updates to other various G4 peripherals. Do you prefer submitting those as individual PRs, or is one larger update better/easier to review? No preference either way :)

adamgreig (approved):
Oops, sorry, I missed the earlier commit. I'll merge these now and then let's put the new changes in a separate PR.

bors merge

PR #531, opened 2021-04-05
Added peripherals/gpio/gpio_v2.yaml to stm32g4 series

Geens:
I think the register definitions in gpio_v2.yaml match those of the stm32g4 series so I added them to the device files.


adamgreig (approved):
Thanks!

bors merge

PR #533, opened 2021-04-08
STM32F446: FLASH description, RCC.PLLCFGR.PLLR

cyberillithid:
So I almost made the same mistake as #374, but actually managed to notice that before starting a pull request, so here I just included the [flash_v1.yaml](https://github.com/stm32-rs/stm32-rs/blob/master/peripherals/flash/flash_v1.yaml) to the f446 device file.
(Apparently, the decision was made not to fiddle with `_read/_write` for 'write 1 to reset' in the library, so I abstained from including that for the FLASH.SR bits)
Moreover, SVD apparently missed PLLR field for some reason, so I based that fix on the similar solution from the f412 patch.

### DIFF
--- a/devices/stm32f446.yaml
+++ b/devices/stm32f446.yaml
@@ -66,6 +66,13 @@ RCC:
     _modify:
       SPDIFSEL:
         name: SPDIFRXSEL
+  PLLCFGR:
+    _add:
+      PLLR:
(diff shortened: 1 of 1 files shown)


adamgreig (approved):
Great, thanks!

bors merge

PR #534, opened 2021-04-10
STM32F401: Add FLTR register to all I2C peripherals

yusefkarim:
This adds the I2C FLTR register defined in RM0368. I have tested this result in code that compiles and looks correct to me (see below). However, I have not tested that this indeed does the right thing in hardware. Not even sure how I would do that :).

```rust
i2c1.fltr.write(|w| w.anfoff().enabled().dnf().no_filter());
```

I only made changes in `devices/stm32f401.yaml` as I personally don't know whether other similar devices have this. For example, STM32L4 does not follow this structure, but maybe other STM32F4's do? Don't know.

### DIFF
--- a/devices/stm32f401.yaml
+++ b/devices/stm32f401.yaml
@@ -12,46 +12,6 @@ _rebase:
       TIFRFE:
         name: "FRE"
 
-"I2C*":
-  _add:
-    # FLTR register defined RM0368
-    FLTR:
-      description: FLTR register
-      addressOffset: 0x24
-      access: read-write
-      resetValue: 0x0000
-      fields:
-        ANFOFF:
-          description: Analog noise filter
-          bitOffset: 4
-          bitWidth: 1
-        DNF:
-          description: Digital noise filter
-          bitOffset: 4
-          bitWidth: 4
-
-  FLTR:
-    ANFOFF:
-        Enabled: [0, "Analog noise filter enabled"]
-        Disabled: [1, "Analog noise filter disabled"]
-    DNF:
-        NoFilter: [0, "Digital filter disabled"]
-        Filter1: [1, "Digital filter enabled and filtering capability up to 1 tI2CCLK"]
-        Filter2: [2, "Digital filter enabled and filtering capability up to 2 tI2CCLK"]
-        Filter3: [3, "Digital filter enabled and filtering capability up to 3 tI2CCLK"]
-        Filter4: [4, "Digital filter enabled and filtering capability up to 4 tI2CCLK"]
-        Filter5: [5, "Digital filter enabled and filtering capability up to 5 tI2CCLK"]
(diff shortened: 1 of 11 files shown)


devices/stm32f401.yaml
```diff
@@ -12,6 +12,46 @@ _rebase:
...
+      access: read-write
+      resetValue: 0x0000
+      fields:
+        ANFOFF:
+          description: Analog noise filter
+          bitOffset: 4
+          bitWidth: 1
+        DNF:
+          description: Digital noise filter
+          bitOffset: 4
```
adamgreig, line 30:
```suggestion
          bitOffset: 0
```

adamgreig (changes requested):
Thanks for this PR! It's interesting, it looks like this FLTR register is widely missing but is present in many F4 devices (RM0090 says "F42x and F43x only", so it's _not_ in the earliest F405/F407, but seems to be in everything else).

Since it looks like this change is also needed in many other F4 devices, would you mind putting these changes into new files? The part where you add FLTR could go in `devices/common_patches/i2c_v1_fltr.yaml` and the part describing the allowed values in `peripherals/i2c/i2c_v1_fltr.yaml`.

Those two files should then be included here in f401, and also in f410, f411, f412, and the `peripherals` file only in f413, f427, f429, f446, and f469 (which all already have the register but not the descriptions).

That would then complete coverage for I2C on all the F4 devices, which would be really nice!

Let me know if that doesn't make sense to you, or if you don't have time to do it yourself.


Also, I noticed the `bitOffset` was wrong, should be 0 instead of 4 since the field starts at bit 0.

devices/stm32f401.yaml
yusefkarim: Pushed an update. Thank you for looking into all of this, very nice this applies to many other devices as well!

Sorry about the `bitOffset`, thanks for finding that error. I also noticed I had `ANFOFF` when the correct name is `ANOFF`. Fixed that.

adamgreig: Thanks! And good spot on ANOFF. It turns out actually every other STM32 (those using the newer "v2" I2C peripheral) call it ANFOFF in their CR2 register, so maybe that's where you got it from? Some of the F4s that also include that newer I2C peripheral (generally I2C4 or FMP_I2C) also have ANFOFF. However, the reference manual is clear that all the F4s with the v1 peripheral and FLTR register call it ANOFF, so your patches seem good to me.

adamgreig (approved):
Thanks for the update!

bors merge

devices/stm32f401.yaml
yusefkarim: Yea, I'm reading a [book](http://web.eece.maine.edu/~zhu/book/) that's using the STM32L4. Supposedly the L4 has the `ANFOFF` (as opposed to `ANOFF`) in CR1!

PR #535, opened 2021-04-10
SPI for L4 series

kenbell:
RM0394 documents the L4 series.

The common SPI mostly applies, but L4 series SPI differs from SPI_v1, v2 and v3 variants - so defining an 'spi_l4' variant.

Two SVD fields need renaming as part of this to be consistent with documented field names (and the ST-provided C HAL):

- CR1.CRCL
- SR.FRE


adamgreig (approved):
Thanks, looks good!

bors merge

PR #538, opened 2021-04-13
F0 patches that rename registers should modify the existing register

lynaghk:
While consuming a patched F0 SVD from https://github.com/lynaghk/svd2zig/ I noticed that a few patches add registers which alias existing ones at the same `bitOffset`.
E.g., [AUTDLY](https://github.com/stm32-rs/stm32-rs-nightlies/blob/7f67ec12360dfdfa9ce08aa76bf088c17f21cdc9/stm32f0/src/stm32f0x0/adc/cfgr1.rs#L1180-L1184) and [WAIT](https://github.com/stm32-rs/stm32-rs-nightlies/blob/7f67ec12360dfdfa9ce08aa76bf088c17f21cdc9/stm32f0/src/stm32f0x0/adc/cfgr1.rs#L1230-L1234).

Based on the ST reference doc:

Screen Shot 2021-04-13 at 7 54 06 PM

I suspect the patch intent was actually to replace the former with the latter.
Ditto with DMA, which should also (I suspect) modify rather than alias:

Screen Shot 2021-04-13 at 7 55 49 PM

### DIFF
--- a/devices/common_patches/f0_adc_fixes.yaml
+++ b/devices/common_patches/f0_adc_fixes.yaml
@@ -1,4 +1,4 @@
-# Renames a few fields in the F0 ADC and adds the WAIT file which was missed
+# Renames a few fields in the F0 ADC and replaces AUTDLY with WAIT
 
 ADC:
   ISR:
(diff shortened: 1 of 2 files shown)


adamgreig (approved):
Good catch, thanks!

bors merge

PR #540, opened 2021-04-17
updates for working with next svd2rust release

burrbull:
https://github.com/rust-embedded/svd2rust/pull/508

### DIFF
--- a/peripherals/flash/flash_latency16.yaml
+++ b/peripherals/flash/flash_latency16.yaml
@@ -1,6 +1,6 @@
 "FLASH":
   ACR:
-    LATENCY: 
+    LATENCY:
       WS0: [0, "0 wait states"]
       WS1: [1, "1 wait states"]
       WS2: [2, "2 wait states"]
--- a/peripherals/flash/flash_latency16.yaml
+++ b/peripherals/flash/flash_latency16.yaml
@@ -1,0 +1,19 @@
+"FLASH":
+  ACR:
+    LATENCY: 
+      WS0: [0, "0 wait states"]
+      WS1: [1, "1 wait states"]
+      WS2: [2, "2 wait states"]
+      WS3: [3, "3 wait states"]
+      WS4: [4, "4 wait states"]
+      WS5: [5, "5 wait states"]
+      WS6: [6, "6 wait states"]
+      WS7: [7, "7 wait states"]
+      WS8: [8, "8 wait states"]
+      WS9: [9, "9 wait states"]
+      WS10: [10, "10 wait states"]
+      WS11: [11, "11 wait states"]
+      WS12: [12, "12 wait states"]
+      WS13: [13, "13 wait states"]
+      WS14: [14, "14 wait states"]
+      WS15: [15, "15 wait states"]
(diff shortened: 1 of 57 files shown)


burrbull:
cc @adamgreig 

So, I see a lot of `bitWidth` bugs.
```
[ERROR svd2rust] In device `STM32G0B0`
    
    Caused by:
        0: In peripheral `TIM14`
        1: In register `CCMR1_Output`
        2: Parsing field #4
        3: In field `OC1M2`
        4: Value 2 out of range 0..2
```

burrbull:
Rebased on #546 

Require https://github.com/stm32-rs/svdtools/pull/57 !!!

burrbull:
@adamgreig rebased

adamgreig:
Thanks for working through this! I need to check some of the G0 ones a bit more carefully, but otherwise the main question is about all the F7 and L4 `tim_ccm_v2.yaml` files that are commented out: what needs to be done with them?

peripherals/flash/flash_latency16.yaml
```diff
@@ -0,0 +1,19 @@
...
+      WS6: [6, "6 wait states"]
+      WS7: [7, "7 wait states"]
+      WS8: [8, "8 wait states"]
+      WS9: [9, "9 wait states"]
+      WS10: [10, "10 wait states"]
+      WS11: [11, "11 wait states"]
+      WS12: [12, "12 wait states"]
+      WS13: [13, "13 wait states"]
+      WS14: [14, "14 wait states"]
+      WS15: [15, "15 wait states"]
```
adamgreig, line 19: Do we even need this latency16 file? Looks like all users of flash_v1.yaml use the flash_latency8.yaml version.

adamgreig:
I've finished reviewing and I guess it makes sense to merge this before tackling #546 since it's based off it. My only outstanding question really is about the commented-out CCMR files, what do you think?

* Make new CCMR YAMLs for the different field widths and apply them as required to each device?
* Merge this now and do that in another PR?
* Something else?

burrbull:
> I've finished reviewing and I guess it makes sense to merge this before tackling #546 since it's based off it. My only outstanding question really is about the commented-out CCMR files, what do you think?
> 
>     * Make new CCMR YAMLs for the different field widths and apply them as required to each device?
> 
>     * Merge this now and do that in another PR?
> 
>     * Something else?

The comments of  CCMR files is fast fix to make compile, as currently those files are broken.
A real fix of this you can see in #546. So I propose merge this. And then review #546.

peripherals/flash/flash_latency16.yaml
burrbull: Looks like I planned to do something, but forgot.

burrbull: According to RM only 405/415 407/417 series in STM32F4 don't support higher states.

burrbull: Fixed

adamgreig (approved):
Thank you very much for working this out!

Understood about CCR being fixed in #546.

I've just noticed on a second review that the F7x2's rebase of TIM5 to use TIM2 means its option register is no longer correct, but let's fix that in another PR. Everything else LGTM, thanks for fixing the flash latency bits.

bors merge

PR #552, opened 2021-05-12
Fix LCD RAM_COM register size on L4x6

AndreasKarg:
- Make RAM_COM registers bigger
- Replace individual segment flags with one 40-bit field

# Note
If I understand https://www.st.com/content/st_com/en/search.html#q=lcd-t=products-page=1 and the reference manual (RM0351) right, the L4x6 is the only L4 with this LCD peripheral.

The SVDs for the 4x1, 2, 3 and 5 seem to contain the same peripheral definition though.

Due to this contradiction, I have only applied the patch to the 4x6 for now.
If someone could throw another pair of eyes on the documentation and confirm whether the other chips do or don't have this peripheral, I'm happy to update the PR and patch those as well.

If it turns out that the SVDs are wrong and the LCD peripherals don't exist, I suggest tackling that in a separate PR.


richardeoin:
Hi @AndreasKarg, thanks for the PR!

It seems there's two groups of parts that have this "Segment LCD" peripheral:
* L4x6, reference manual = RM0351
* L4x3, reference manual = RM0394

Looking through the [part table](https://www.st.com/en/microcontrollers-microprocessors/stm32l4-series.html#products) there are a couple of other parts that appear when filtering for "Segment LCD" (such as the L452) but then looking at the corresponding datasheet there's no mention of it. I'm guessing those are probably just a mistake on ST's website.

So could you apply this patch to the L4x3 also? 

The patch itself looks good, but typically we put patches that "fix" the SVD (using _add/_modify/_delete etc) in the `devices/` hierarchy. In this case I would put the patch somewhere like `devices/common_patches/l4_lcd_segment.yaml`. The files in `peripherals/` are used for the field variants (the enumerated values) or ranges.

adamgreig (approved):
bors merge

PR #556, opened 2021-05-27
Build peripheral comparisons pages as part of html target

qwandor:
Also cleaned up the HTML a bit.


adamgreig (approved):
Nice work, thanks!

bors merge

PR #566, opened 2021-06-03
Fix bit offset for RMVF in STM32l0x2 and STM32l0x3

lulf:
According to the reference manuals for STM32L0x{2, 3}, the RCC_CSR
register RMVF has the offset 23 and not 24. Offset 24 is occupied by a
Firewall reset which is also added in this change

Page 219 in the STM32L0x2 release manual can be used to verify.

### DIFF
--- a/devices/stm32l0x2.yaml
+++ b/devices/stm32l0x2.yaml
@@ -73,6 +73,15 @@ RCC:
     _modify:
       LPWRSTF:
         name: LPWRRSTF
+    _modify:
+      RMVF:
+        bitOffset: 23
+    _add:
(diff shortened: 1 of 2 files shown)


richardeoin:
The change itself looks good, thanks! Once small regression - you've accidentally over-written the `_modify:` section immediately above that's also for the `CSR` register. If you just delete the `_modify` line you added then the two sections will be combine and all will be dandy.

lulf:
@richardeoin Thanks, pushed a fix.

richardeoin (approved):
Thanks!

bors r+

PR #571, opened 2021-06-07
Fix wrong bit offset of CRCEN/CRCRST/CRCSMEN and USARTXSMEN for L4 family

lulf:
Various SVD fixes to L4 chips.

### DIFF
--- a/devices/common_patches/stm32l4x2_l412.yaml
+++ b/devices/common_patches/stm32l4x2_l412.yaml
@@ -4,6 +4,10 @@
 # SVD incorrectly labels APB1ENR1 bit 26 as USBF instead of USBFSEN.
 # SVD incorrectly labels APB1ENR1 bit 14 as SPI1EN instead of SPI2EN.
 # SVD incorrectly omits APB1ENR1 bit 1 (TIM3RST), which is present for
+# SVD incorrectly shifts CRCRST 11 bits instead of 12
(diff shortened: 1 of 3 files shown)


richardeoin:
Thanks for the PR! It all looks correct, except the field at bit 19 of APB1SMENR1 is called UART4SMEN (UART not USART).

This is very tedious, but there's also some very similar errors for L4x1 and L4x2. It would be good to have these in the PR also:
* Bit 18 of APB1ENR1 should be called USART3EN (3 not 1)
* Bit 18 of APB1RSTR1 should be called USART3RST (3 not 1)

lulf:
@richardeoin Sorry for the late reply. Fixed up the rename.

Regarding the USART1EN -> USART3EN rename, that is already part of the l4x1 and l4x2 patches from what I an see. (Line 12-14 in stm32l4x1.yaml for instance).

richardeoin:
Indeed the USART1EN -> USART3EN rename is being applied for the l4x2 already - I missed that. For the l4x1 it should be applied but the original patches have an error - there's two `_add` sections, so the first one is being ignored.

Actually this PR has the same error with two `_add` sections (so `USART3SMEN` is ignored), which should definitely be fixed.

richardeoin:
Thanks, looks good now!

bors r+

richardeoin (approved):
Thanks, looks good now!

bors r+

Issue #572, opened 2021-06-07
L4x5 apb1enr1 missing crsen


David-OConnor:
https://docs.rs/stm32l4/0.13.0/stm32l4/stm32l4x5/rcc/apb1enr1/index.html

It's present on other L4 variants.

adamgreig:
Closed by #580

Issue #570, opened 2021-06-06
WB55 missing some exti fields


David-OConnor:
EXTI_C1IMR1, EXTI_FTSR1 and EXTI_RTSR1 are missing all fields.

adamgreig:
Closed by #580

Issue #565, opened 2021-06-03
STM32L4x3 USB points to wrong addr


David-OConnor:
https://github.com/stm32-rs/stm32-rs-nightlies/blob/master/stm32l4/src/stm32l4x3/mod.rs#L1347

Should be 0x4000_6800. The current one is the *USB SRAM* address. 

L4x2 PAC is correct. L4x1 PAC doesn't have USB. L4x6 PAC uses OTG. L4x5 PAC also has this error, but if I understand correctly, shouldn't even have this periph since it uses the OTG periph.

To fix, could copy + paste L4x2 USB module. Should be the same.

adamgreig:
Closed by #580

Issue #563, opened 2021-06-02
STM32L4x3 PAC missing USB_BCDR register


David-OConnor:
(Battery charging detector reg) Note: This is not present on some USB FS peripherals like on F3, which are otherwise the same. Offset 0x58.

Note that the [stm32-usbd PAC](https://github.com/stm32-rs/stm32-usbd/blob/master/src/pac/usb.rs) does include it.

adamgreig:
Closed by #580

PR #573, opened 2021-06-07
Fixed various issues with L4 and WB

### LINKED ISSUE #555: L41x and L42x use diff RTC regs

### LINKED ISSUE #567: STM32L4x3 missing USB interrupt
Should be line 67, omitted here. L4x2 is fine. 

https://github.com/stm32-rs/stm32-rs-nightlies/blob/master/stm32l4/src/stm32l4x3/mod.rs#L276

David-OConnor:


David-OConnor:
Messed up the Git. See https://github.com/stm32-rs/stm32-rs/pull/576

PR #574, opened 2021-06-10
Make FDCAN_TEST register writeable for H7 family

ijager:
To make sure the Writable trait is also implemented for the TEST register


adamgreig (approved):
Thanks!

bors merge

Issue #575, opened 2021-06-12
WBx5 tim16 CR1 wrong fields


David-OConnor:
Tim16 CR1 fields are wrong, as are TIM16 and TIM17 DIER. All 3 of these are showing the same fields.

[tim16 cr1 fields](https://docs.rs/stm32wb/0.13.0/stm32wb/stm32wb55/tim16/cr1/index.html)

(And the correct TIM17, which this should be the same as)
[tim17 cr1 fields](https://docs.rs/stm32wb/0.13.0/stm32wb/stm32wb55/tim17/cr1/index.html)

Incorrect DIER fields: https://docs.rs/stm32wb/0.13.0/stm32wb/stm32wb55/tim16/dier/index.html
Correct DIER fields (TIM1): https://docs.rs/stm32wb/0.13.0/stm32wb/stm32wb55/tim1/dier/index.html

adamgreig:
Closed by #580

PR #576, opened 2021-06-18
Fixed various issues with L4 and WB take 2

### LINKED ISSUE #555: L41x and L42x use diff RTC regs

David-OConnor:

See https://github.com/stm32-rs/stm32-rs/pull/573; "unrecoverable" git errors.


David-OConnor:
The 412 RTC changes don't appear to be taking effect, after building and inspecting the module. Any ideas?

adamgreig:
I guess the problem is with the `_include: stm32l4x2.yaml` at the end; I suspect that will overwrite things you don't want to overwrite, such as `_svd`, in the parent YAML. I don't think it's intended that you'd import a top-level YAML into another top-level YAML, so that's my guess as to the problem. Try just including the relevant bits of that file directly?

By the way, instead of changing the SVD zip file (which should ideally be exactly as-downloaded from ST), you can modify `extract.sh` to add a new copy line to create a new identical SVD for L412, like https://github.com/stm32-rs/stm32-rs/blob/master/svd/extract.sh#L9

David-OConnor:
Done! That sorted it. Marking as draft for now; running into empty built-libs on L4x3, L412, and WB55, for reasons I haven't yet determined.

David-OConnor:
Fixed the not-building problems; various syntax errors, eg `read` instead of `read-only`. Currently, some of the L412 RTC reg changes are taking place, but not others, upon generated-module inspection. I wonder if this is an order-error, eg if a reg is added to a place that's not yet freed from a later-modified reg. (?)

David-OConnor:
Fixed most L412 RTC problems; syntax error re ordering of `_modify` and the reg name. Currently looks good, except 3 out of 4 of the added regs (at the end of the block) are not adding. Maybe need to increase size of reg block somehow?

David-OConnor:
Broke git again

#576

PR #578, opened 2021-06-21
Various L4 and WB fixes, take 4

### LINKED ISSUE #555: L41x and L42x use diff RTC regs

David-OConnor:

See https://github.com/stm32-rs/stm32-rs/pull/577; "unrecoverable" git errors.

Not currently working:
- Last 3 (by address) added L412 RTC regs are not adding
- WB HSEM and IPCC renamed regs are not changing


David-OConnor:
Also removed the L4x2 `USB_FS` rename to `USB`: the RM calls it `USB_FS`.

David-OConnor:
Broke git again

PR #580, opened 2021-06-24
L4 and WB fixes

### LINKED ISSUE #555: L41x and L42x use diff RTC regs

David-OConnor:

See #578; "unrecoverable" git errors.

Not currently working:

- Last 3 (by address) added L412 RTC regs are not adding
- WB HSEM and IPCC renamed regs are not changing
- BCDR field added to USB

Removed the L4x2 `USB_FS` rename to `USB`: the RM calls it `USB_FS`. Previous justification, from a code comment: 
> we call the resulting peripheral `USB` instead of `USB-FS` to be aligned with `mvirkkunen/stm32f103-usb`

The *peripheral* is called `USB`; the interrupt is called `USB_FS`.


David-OConnor:
Of note, re the `USB_FS` change on L4x2. Unable to get the module to compile with that rename removed. Eg `USB_FS: USB_FS {` in `mod.rs`: "not found in this scope for the last". I can revert the change, or maybe there's a way to fix it and keep the `_FS` name for the interupt.

David-OConnor:
All issues I posted about are fixed. Thanks @adamgreig!

adamgreig (approved):
@David-OConnor thanks for this PR! I've made a few tweaks in the recent commits, mainly:

* some bits were wrong in the TIM16/TIM17 fixes for WB55
* you can use `_split` to break a wide field into lots of individual bits, without having to write them all out
* (so I applied it to the other similar fields in WB55 EXTI too)
* I refactored the L4x2 and L412 common patches into a separate file so they can't drift apart

bors merge

PR #581, opened 2021-07-05
l0: fix CNT, ARR, CCRx register sizes

kenbell:
The existing patch for l0 timers was setting these registers to 16-bit as a side-effect of redefining TIM2/TIM3 - but for all TIM objects, the corresponding reference manuals indicate CNT, ARR and CCRx values should be 16-bit registers.

Relevant ref manuals: RM0367, RM0376, RM0377

### DIFF
--- a/devices/common_patches/l0_tim.yaml
+++ b/devices/common_patches/l0_tim.yaml
@@ -1,3 +1,14 @@
+# Per RM0367,RM0376, RM0377 all timers have 16 bit registers for CNT,ARR,CCRx
+TIM*:
+  _modify:
+    CNT:
+      size: 16
+    ARR:
+      size: 16
+    CCR*:
+      size: 16
+
+
 # Fix Timers for L0 Flash
 "TIM[23]":
   _delete:


adamgreig:
The memory map comparison doesn't seem to show any change, which timers are currently the wrong size?

kenbell:
I think I messed up.  I cancel this PR, and try again.  I made the identical changes in the individual device files, which seemed to work - then realized it should be a common fix.  For some reason it's not applying from this common patch file.

kenbell:
Closing - will re-open with a working patch.

kenbell:
Re-opening.  It looks like the patch is working (at least locally for me).  Before the patch:
stm32l0x2.svd.patched has TIM6:
```xml
        <register>
          <name>CNT</name>
          <displayName>CNT</displayName>
          <description>counter</description>
          <addressOffset>0x24</addressOffset>
          <size>0x20</size>
          <access>read-write</access>
          <resetValue>0x00000000</resetValue>
          <fields>
            <field>
              <name>CNT</name>
              <description>Low counter value</description>
              <bitOffset>0</bitOffset>
              <bitWidth>16</bitWidth>
            <writeConstraint><range><minimum>0</minimum><maximum>65535</maximum></range></writeConstraint>
            </field>
          </fields>
        </register>
```

after the patch has TIM6:
```xml
        <register>
          <name>CNT</name>
          <displayName>CNT</displayName>
          <description>counter</description>
          <addressOffset>0x24</addressOffset>
          <size>16</size>
          <access>read-write</access>
          <resetValue>0x00000000</resetValue>
          <fields>
            <field>
              <name>CNT</name>
              <description>Low counter value</description>
              <bitOffset>0</bitOffset>
              <bitWidth>16</bitWidth>
            <writeConstraint><range><minimum>0</minimum><maximum>65535</maximum></range></writeConstraint>
            </field>
          </fields>
        </register>
```

The size field has gone from 0x20 (32) to 16

adamgreig (approved):
Ah, got you, I was looking at the field size and not the overall register size. The reference manual does say the registers can be written by half-words (16bit) _or_ words (32bit) and read by both too, so I guess this shouldn't affect functionality but no harm in matching the width shown on the manual.

Thanks!

bors r+

PR #543, opened 2021-04-18
Fix name of bit fields LPVAE and LPHFPE in DSIHOST VMCR for STM32F469

Jsbjr:

PR #585, opened 2021-07-22
strip DSI

burrbull:
Instead of #543


adamgreig (approved):
Thanks @burrbull and @jsbjr! 

bors merge

PR #588, opened 2021-08-01
Fix #587

### LINKED ISSUE #587: Not all patches in stm32l4x1.yaml are applied correctly
The output (stm32l4x1.svd.patched) does not contain the USART1 -> USART3 fix:
https://github.com/stm32-rs/stm32-rs/blob/master/devices/stm32l4x1.yaml#L6-L10

Other patch rules like the SPI fix from the same file are applied correctly.

timokroeger:
Duplicate `_modify` prevented the fix to be applied.


adamgreig (approved):
Great, thanks!

bors merge

PR #589, opened 2021-08-03
Fix U(S)ART RCC enable/reset bits for F4 devices

timokroeger:
Make bit names consistent with peripheral names:
USART1-3, USART6
UART4-5, UART7-10


adamgreig (approved):
Nice, thank you!

bors merge

PR #591, opened 2021-08-15
[scripts/makehtml] Fix displaying of registers with the same addressOffset and parsing of addressOffset

maximeborges:
This pull request fix two bugs in the `makehtml.py` script:

- When parsing the addressOffset value, the value is parsed as base16 even if it's in base10.
- When two registers have the same addressOffset, one of the two is suppressed

The first bug is fixed by checking if the addressOffset starts with `0x` to parse from base16, or parse from base10 by default (when a register is generated from a patch).

The second bug is fixed by changing the index used for the list of registers from the addressOffset to the register name.

### DIFF
--- a/scripts/makehtml.py
+++ b/scripts/makehtml.py
@@ -69,7 +69,8 @@ def parse_device(svdfile):
             rdesc = rtag.findtext('description')
             rrstv = rtag.findtext('resetValue')
             raccs = rtag.findtext('access') or "Unspecified"
-            roffset = int(rtag.findtext('addressOffset'), 16)
+            addr_offset = rtag.findtext('addressOffset')
+            roffset = int(addr_offset, 16 if addr_offset.startswith('0x') else 10)
             for ftag in rtag.iter('field'):
                 register_fields_total += 1
                 fname = ftag.findtext('name')
@@ -146,7 +147,7 @@ def parse_device(svdfile):
             # Bodge to prevent /0 when there are no fields in a register
             if register_fields_total == 0:
                 register_fields_total = 1
-            registers[roffset] = {"name": rname,
+            registers[rname] = {"name": rname,
                                   "offset": "0x{:X}".format(roffset),
                                   "description": rdesc, "resetValue": rrstv,
                                   "access": raccs, "fields": fields,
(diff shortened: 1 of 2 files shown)


maximeborges:
Note about the memory map comparison:
It looks like a lot of registers with splitted fields now have a proper bitOffset, so it explains why there is tons of modifications here

scripts/makehtml.py
```diff
@@ -69,7 +69,8 @@ def parse_device(svdfile):
             rdesc = rtag.findtext('description')
             rrstv = rtag.findtext('resetValue')
             raccs = rtag.findtext('access') or "Unspecified"
-            roffset = int(rtag.findtext('addressOffset'), 16)
+            addr_offset = rtag.findtext('addressOffset')
+            roffset = int(addr_offset, 16 if addr_offset.startswith('0x') else 10)
```
adamgreig, line 73:
```suggestion
            roffset = int(rtag.findtext('addressOffset'), 0)
```

Unless we have any octal expressions I think this should work and is a bit simpler, what do you think?

scripts/makehtml.py
```diff
@@ -146,7 +146,7 @@ def parse_device(svdfile):
             # Bodge to prevent /0 when there are no fields in a register
             if register_fields_total == 0:
                 register_fields_total = 1
-            registers[roffset] = {"name": rname,
+            registers[rname] = {"name": rname,
```
adamgreig, line 149:
```suggestion
            registers[(roffset, rname)] = {"name": rname,
```

I had to check to remember what this is used for, but it's inside `makehtml.template.html`, where the registers are iterated over by sorted key order, so previously they were in register-offset order but now they'd be in register-alphabetical order. If we make this key (offset, name), we preserve the sort order but allow the overlapping registers to still appear.

I think it would be good to also change `makehtml.template.html` L121 to something like `{% for _, register in peripheral.registers|dictsort %}` (changing `roffset` to `_`) to make it clear that the key isn't used later, since it's now a tuple.

Finally the indentation of the fields below needs updating since this line changes length.

maximeborges: Good catch, I thought registers were inserted by their roffset already, but I missed the sorting function.

adamgreig (approved):
Great, thanks!

bors merge

PR #595, opened 2021-08-17
Allow using cortex-m-rt 0.7

adamgreig:


mattico:
Could this be backported to v0.13.x? I'd appreciate it!

adamgreig:
I'm hoping to release 0.14 pretty soon with this and all the other recent changes; would that help or do you specifically want a 0.13 (for a particular family or all of them)? It's a non-trivial amount of work and resources (on crates.io/docs.rs as well) for each release, so I'd rather avoid a 0.13 backport with just this change if I can help it, but if there's a particular family where it would be useful to have something ASAP I could probably give it a go.

mattico:
> I'm hoping to release 0.14 pretty soon with this and all the other recent changes

Works for me, thanks!

PR #546, opened 2021-04-24
rework OCxM fields in timers

burrbull:
изображение

MP1 is not covered by this PR

Closes #541

PR #596, opened 2021-08-17
Resolve small breakage in OCxM fields for stm32h735 in #546

richardeoin:
After 0cb0f60346d21320bbfde60b37d2d14aa49439d2 the peripheral patches for the OCxM fields were no longer applied for these parts. This PR fixes that.

* Add `peripherals/tim/v2/ccm.yaml` that appeared during the refactor but wasn't added to this device
* Remove duplicate include of `common_patches/tim/tim_ccr.yaml`


adamgreig (approved):
Ah, oops, I'd caught this TIM17 problem in the PR but it looks like the fix missed H743 and I missed that one checking the fix. Thanks for catching it and fixing the peripheral patch too.

bors r+

PR #597, opened 2021-08-17
stm32f7x7: Fix missing fields for timers

maximeborges:
A lot of fields are wrong or missing for the timers for the STM32F7x7.
This pull request fix all of them regarding their definitions in the reference manual.

### DIFF
--- a/devices/stm32f7x7.yaml
+++ b/devices/stm32f7x7.yaml
@@ -191,6 +191,164 @@ RCC:
         bitWidth: 3
         access: read-write
 
+TIM*:
+  CR1:
+    _add:
+      UIFREMAP:
+        description: UIF status bit remapping
+        bitOffset: 11
+        bitWidth: 1
+
+TIM[1-589],TIM1[0-4]:
+  SMCR:
+    _add:
+      SMS_3:
+        description: Slave mode selection - bit 3
+        bitOffset: 16
+        bitWidth: 1
+
+TIM[1-589],TIM1[04]:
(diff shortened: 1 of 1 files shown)


adamgreig:
Thank you!

Often when one device is wrong like this the related devices (f7x6 and f7x9 especially but maybe other f7) have the exact same problems, so it might be worth a quick check to see and then these patches could go into a common patch file included in both. Would you mind having a look?

maximeborges:
> Often when one device is wrong like this the related devices (f7x6 and f7x9 especially but maybe other f7) have the exact same problems, so it might be worth a quick check to see and then these patches could go into a common patch file included in both. Would you mind having a look?

I'll check tomorrow if I have some time.

The CI is failing with the error:

    svdtools.patch.SvdPatchError: register CCMR1_Output already has a field OC1M_3

Do you have any idea why the CI is failing but I can't reproduce the error locally ?

devices/stm32f7x7.yaml
```diff
@@ -191,6 +191,164 @@ RCC:
...
+  OR:
+    _delete:
+      - E*
+      - T*
+    _modify:
+      ITR1_RMP:
+        bitOffset: 10
+        bitWidth: 2
+
+# This is not working?!
```
maximeborges, line 337: Also this seems to not be applied, neither in the doc nor in the code.
Any idea ?

maximeborges:
Supersedded by https://github.com/stm32-rs/stm32-rs/pull/546

PR #599, opened 2021-08-19
Fix STM32WL EXTI14 using values for EXTI15

newAM:
cc @jorgeig-space


newAM:
Force pushed to rebase onto the latest master since that memory map comparison was hard to read.

adamgreig (approved):
Thanks!

bors merge

PR #608, opened 2021-08-30
Escape square braces in STM32WL docs

newAM:
Fixes all the dead links for the STM32WL crate.

cc @jorgeig-space

### DIFF
--- a/peripherals/pwr/pwr_wl.yaml
+++ b/peripherals/pwr/pwr_wl.yaml
@@ -45,8 +45,8 @@ PWR:
       Disabled: [0, "Internal wakeup line interrupt to CPU disabled"]
       Enabled: [1, "Internal wakeup line interrupt to CPU enabled"]
     EWRFIRQ:
-      Disabled: [0, "Radio IRQ[2:0] is disabled and does not trigger a wakeup from Standby event to CPU."]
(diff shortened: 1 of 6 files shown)


adamgreig:
PR looks fine, but any idea what's misinterpreting the unescaped square brackets? I don't think anything inherent to SVD requires it (see the `\n` fiasco in the RP2040 SVDs...) since the datatype for `description` is specified to be `xs:string` which may contain whatever.

newAM:
> PR looks fine, but any idea what's misinterpreting the unescaped square brackets? I don't think anything inherent to SVD requires it (see the `\n` fiasco in the RP2040 SVDs...) since the datatype for `description` is specified to be `xs:string` which may contain whatever.

That's a very good question, and now I am unable to reproduce the doc build with the broken links. :unamused:
I am going to close this for now until I can trace back to where exactly this is breaking when I encounter it again.

newAM:
I stumbled across this again, updating the PR for the git archeologists of the future.

Short version: problem between keyboard and chair.
Long version: [cargo-deadlinks](https://github.com/deadlinks/cargo-deadlinks) emits errors pointing at dependencies when using `rustdoc` instead of `cargo doc` after a `cargo clean`.  Due to some unfortunate formatting the error messages lead me to mistakenly believe that the bit ranges in the PAC (e.g. `[3:1]`) were generating invalid links.

PR #609, opened 2021-08-30
Fix dead link in generated README

newAM:
I am not exactly sure what this link was supposed to point to; previous it generated https://github.com/stm32-rs/stm32-rs/tree/master/stm32wl for the STM32WL which is invalid.

Now it generates a link to https://crates.io/crates/stm32wl which is valid.


adamgreig (approved):
Ah, thanks, that's for when the generated crate directories were kept in git (without generated source code), including the README file which had that table. Since we no longer keep the READMEs etc in this repository, the new link seems good.

bors merge

PR #611, opened 2021-08-31
[scripts] Fix int base when parsing

maximeborges:
When comparing some peripherals I saw that some `baseAddress` where parsed as base 10 but still prefixed with "0x".
I fixed every int parsing left in the `scripts` folder to automatically use the right base.


adamgreig (approved):
Thanks!

bors merge

PR #612, opened 2021-09-01
[stm32f7] Fix SYSCFG registers

maximeborges:
Following #606, fix fields  in SYSCFG for every F7 chips


adamgreig:
Could you rebase this now that the timers PR is merged? It would make it easier to review and to compare the memory maps.

maximeborges:
> Could you rebase this now that the timers PR is merged? It would make it easier to review and to compare the memory maps.

Done.

devices/common_patches/f730_f7x2_f7x3_syscfg.yaml
```diff
@@ -0,0 +1,22 @@
+SYSCFG:
+  PMC:
+    _delete:
+      - PB*
+      - I2C*
```
adamgreig, line 5: I don't understand these deletions, it looks like RM0431 has the I2C and PB fields in PMC, and doesn't have the MII_RMII_SEL? Whereas RM0410 does have MII_RMII_SEL as well as PB and I2C, and only RM0385 (stm32f745, f750, f7x6) has the MII_RMII_SEL but no PB/I2C fields.

devices/stm32f7x6.yaml
```diff
@@ -77,6 +77,15 @@ CRYP:
...
       - "IV*"
 
+SYSCFG:
+  MEMRMP:
+    _delete:
+      - FB_MODE
+  PMC:
+    _delete:
+      - PN*
+      - I2C*
```
adamgreig, line 87: Looking at the [memory map](https://github.com/stm32-rs/stm32-rs-mmaps/compare/pr-612-1212829#diff-f146746c79235e17e7a8674bbb31537af0c2d61e3d3552ab34f078fed0fd67d8) it doesn't seem like these deletions are doing anything?

adamgreig (changes requested):
Thanks! This looks mostly sensible but I didn't totally follow the deletions from the PMC register:

devices/common_patches/f730_f7x2_f7x3_syscfg.yaml
adamgreig: I don't know why 0385 doesn't have them, but they're not in any version of the RM and no mention of it in the errata... I don't have any of those F7 to test with. Maybe ST's HAL has clues? I think we can leave them out for now and add later if it turns out they should be here.

adamgreig (approved):
All looks good now, thanks. As mentioned in the comment, we can add the FMP fields to 745/750/7x6 (RM0385) later if they do actually exist.

bors merge

PR #614, opened 2021-09-02
[MP1] Add barebones support for STM32MP153

jhbruhn:
I have added the most recent SVD-file from ST (by requesting it here https://community.st.com/s/question/0D53W00000G1pQg/your-stm32mp157-svd-file-is-highly-incomplete-and-plain-wrong-at-times-could-you-please-update-it?t=1629885026205 ) because the "official" one did contain barely any interrupts.
Currently I am using it only for the interrupt descriptions, thus the peripherals are basically untested (except for compile checks).


adamgreig (approved):
Thanks!

bors merge

PR #621, opened 2021-09-08
[html-compare] f7: Change order of MCUs in the table

maximeborges:
When comparing peripherals and fields from the F7 family, there is 3 big groups of MCU that we can distinguish:

- RM0431: F7x2, F7x3, F30x
- RM0385: F745, F750, F7x6
- RM0410: F765, F7x7, F7x9

This PR reorder the chips in the table so it's easier to detect a missing field or register in each group.


adamgreig (approved):
Thanks, looks good.

bors merge

PR #623, opened 2021-09-08
stm32wb: ADC -> ADC1 and add ADC_COMMON

karlp:
As is commonly done, use "ADC1" even when there's only one ADC.  This
makes it not only more self consistent, (interrupt names were for ADC1)
but also more consistent with sibling targets like STM32L4.

Also, like the L4, the WB svd is missing the ADC_Common registers.  Add
them back in, but note that there's less bits in the common registers.

Signed-off-by: Karl Palsson <karlp@etactica.com>

### DIFF
--- a/devices/common_patches/wb_adc_common.yaml
+++ b/devices/common_patches/wb_adc_common.yaml
@@ -1,0 +1,91 @@
+# Not quite the same as L4 ADC_Common, as there's no slave ADC on WB.
+_add:
+  # This SVD is missing the ADC_Common peripheral that most other parts with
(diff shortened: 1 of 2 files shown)


karlp:
Hrm, that memory map comparison is neat, looks like CCR was already there, btu underneath ADC1 itself. I'll try and remove them out.

karlp:
Ok, that's good now.  I could easily be conviced to drop the renames of the CH17SEL/CH18SEL, but those names match the ref man, and the L4 implelmentation

adamgreig (approved):
Looks good, thank you!

bors merge

PR #624, opened 2021-09-08
stm32wb: syscfg: fix split sections

karlp:
Fixes: https://github.com/stm32-rs/stm32-rs/pull/473

Fixes the base address, and then for the registers after the split,
fixup their addresses manually.

Signed-off-by: Karl Palsson <karlp@etactica.com>

Is there really any need to specifically mark the VREFBUF as "reserved" within the syscfg block as discussed in #473? is that important? the memory map generated looks correct.

### DIFF
--- a/devices/stm32wb55.yaml
+++ b/devices/stm32wb55.yaml
@@ -5,6 +5,10 @@ _modify:
   ADC:
     # Consistent naming across families
     name: ADC1
+  SYSCFG:
+    baseAddress: "0x40010000"
+    addressBlock:
+      size: 0x200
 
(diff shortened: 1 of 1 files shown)


adamgreig (dismissed):
Thanks, this looks like it should work fine.

bors merge

adamgreig:
Ah, boo, looks like it conflicts with #623. Could you rebase please?

karlp:
No problem, rebased.  (They both were doing section renames)

adamgreig (approved):
Thanks!

bors merge

Issue #636, opened 2021-10-02
H743 and H743 DFSDM offsets wrong


David-OConnor:
The offsets appear to be correct for all other H743 variants. They all appear to be wrong on H743 and H743V. Working on a fix, but can't get changes to take ATM.

Attempting to apply a common patch of the form:

```yaml
DFSDM:
  _modify:
    DFSDM_CHCFG0R1: 
      addressOffset: 0x00
    DFSDM_CHCFG0R2: 
      addressOffset: 0x04
    DFSDM_AWSCDR0R:
      addressOffset: 0x08
    DFSDM0_DAT0R:
      addressOffset: 0x0C
    DFSDM0_DATIN0R:
      addressOffset: 0x10

    DFSDM_CHCFG1R1: 
      addressOffset: 0x20
    DFSDM_CHCFG1R2: 
      addressOffset: 0x24
    DFSDM_AWSCDR1R:
      addressOffset: 0x28
    DFSDM0_DAT1R:
      addressOffset: 0x2C
    DFSDM0_DATIN1R:
      addressOffset: 0x30.
..
```

Additionally, the SVD names don't match the form of the RM. Eg `FLT0CR2` vs ` DFSDM0_CR2` and `CH2CFGR1` vs `DFSDM_CHCFG0R1`

Also, indexed channels and filters would be nice, eg H7's DMA streams. I'm going to take a stab at the wrong addresses. Any ideas why the fixes I put in aren't taking?

adamgreig:
Could you push a branch that shows them not working? It's hard to tell from just that snippet, which looks OK in isolation.

David-OConnor:
Done - https://github.com/stm32-rs/stm32-rs/pull/637

David-OConnor:
I think the "not working" is due to typos in the reg names I added.

adamgreig:
If nothing else, we usually strip the peripheral name from field names using the `_strip` command for things like this ([example](https://github.com/stm32-rs/stm32-rs/blob/master/devices/common_patches/sdio_f446.yaml#L12)).

David-OConnor:
Added strip, and the address offsets appear to have taken. Leaving as draft until further QC and ops testing. Appreciate it!

adamgreig:
[...]

PR #637, opened 2021-10-02
Fix address offsets on H743/V DFSDM

David-OConnor:
Addresses #636


adamgreig:
It looks like only the 743/743v use the names you've got in the file, the other H7 series use names either without the DFSDM prefix for the common registers like `CH0CFGR1`, or with it for the filter block ones like `DFSDM_FLT0CR1`. What a pain! I think the patch file you've got now is good for the 743/743v, but the other H7s probably just need the `_strip: DFSDM` alone and I don't think anything else?

David-OConnor:
Done: Strip for non-H743. Redo names/offsets for 743/V

adamgreig:
Thanks for this PR! I've updated it a bit to get cluster support too.

Getting clusters to work was a pain because of a few more weird inconsistencies (how on _earth_ did they end up with `FLT1CHGR` when all the other FLTs are `FLTxJCHGR`???). In the end it seemed best to split it into separate files for 7B3, 743, and 735/47/53.

The downside is that the clusters do not form arrays because there is a gap between each cluster element, and svd2rust 0.19 cannot turn that into an array - it can't add padding because the padding might overlap over registers (though not in this case), and a recent PR to use an array proxy in this case hasn't been released yet (see https://github.com/rust-embedded/svd2rust/pull/531 and https://github.com/rust-embedded/svd2rust/pull/534).

Anyway, they are at least the same _type_, and in future will probably get arrayified.

adamgreig (approved):
bors merge

PR #640, opened 2021-10-03
H7 RM0433: Add NOMCK at SAI [AB]CR1 bit 19 and remove MCKEN

### LINKED ISSUE #618: H7 SAI xCR Master clock error
It looks like the variant used for WB is correct, but it's also the same on H7. Specifically: The H7 PAC (probably SVD) has ACR and BCR bit 19 as NODIV, and 27 as MCKEN. This is correct per WB RM. H7 RM states 27 is not used, and 19 is NOMCK.

Of note, the SAI periph seems similar or identical other than this, and NODIV and MCKEN both relate to generating a master clock.

In the meanwhile, without forking, you can pretend the H7 NODIV field is NOMCK.

richardeoin:
Other H7 parts are unchanged. Improve description of NOMCK/NODIV bit

See previous PR #347. Fixes #618

### DIFF
--- a/devices/common_patches/h7_sai.yaml
+++ b/devices/common_patches/h7_sai.yaml
@@ -3,6 +3,11 @@
     _modify:
       MCKDIV:
         bitWidth: 6
+    _add:
+      NODIV:
+        description: No fixed divider between MCLK and FS
+        bitOffset: 19
+        bitWidth: 1
   AFRCR:
     _modify:
       FSDEF:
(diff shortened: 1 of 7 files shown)


adamgreig:
What do you think of leaving the nodiv field in for these parts too, at the same offset to the nomck field? That way drivers/HAL can keep doing the same thing but users can call it by the RM name if they prefer. Maybe it's just more confusing in the end, I don't know how much the HAL really cares about this bit.

richardeoin:
The default value of nodiv is the typical use case, whilst setting it is a less common use case (I think). However the HAL does write to it a couple of times, so having nodiv at that offset for all parts would be a nice solution

richardeoin:
Thanks for the suggestion, the RM0433 parts now have both `NODIV` and `NOMCK` at the same offset

adamgreig (approved):
Thanks, looks good!

bors merge

PR #649, opened 2021-10-20
Fix Overdrive and BDMADR fields on h743/h753

sephamorr:
Adding fixes to two fields on h743/h753 micros. These bugs were found by comparing against h747 svd files and verified to not be present on any other h7 micros (beyond those fixed in this PR).

ODEN should be 1-bit and have an enumeration. H747 had the correct bit width already; the oden patch is now applied to more devices.
BDMADR should be at 0x70, not 0x60.

### DIFF
--- a/devices/common_patches/h743_hrtim_common.yaml
+++ b/devices/common_patches/h743_hrtim_common.yaml
@@ -1,0 +1,4 @@
+HRTIM_Common:
+   _modify:
+     BDMADR:
+       addressOffset: 0x70
--- a/devices/common_patches/h74x_syscfg_pwrcr.yaml
+++ b/devices/common_patches/h74x_syscfg_pwrcr.yaml
@@ -1,0 +1,6 @@
+SYSCFG:
+  PWRCR:
+    _modify:
+      ODEN:
+        bitWidth: 1
+        description: "Overdrive enable"
--- a/devices/stm32h743.yaml
+++ b/devices/stm32h743.yaml
@@ -100,3 +100,5 @@ _include:
  - common_patches/h7_crc_addr_fix.yaml
  - common_patches/h7_wwdg.yaml
  - common_patches/h7_ethernet_mac.yaml
+ - common_patches/h74x_syscfg_pwrcr.yaml
+ - common_patches/h743_hrtim_common.yaml
--- a/devices/stm32h743v.yaml
+++ b/devices/stm32h743v.yaml
@@ -102,3 +102,5 @@ _include:
  - common_patches/h7_crc_addr_fix.yaml
  - common_patches/h7_wwdg.yaml
  - common_patches/h7_ethernet_mac.yaml
+ - common_patches/h74x_syscfg_pwrcr.yaml
+ - common_patches/h743_hrtim_common.yaml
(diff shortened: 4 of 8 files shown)


sephamorr:
@adamgreig Are these acceptable locations? It seemed that there were some in peripheral subfolders but the bulk just in the root devices/common-patches folder.

There may be more consequences to the name change made I made in this commit for H7{4,5}3v - this fixes the generation of the web pages (makehtml.py) which otherwise could randomly pick h743 or h743v for the h743 entry since the name in the SVD file was identical where is should instead have a line item for all four H7{4,5}3{none, v}

richardeoin:
```
SYSCFG:
  PWRCR:
    ODEN:
      Disabled: [0, "Overdrive mode disabled"]
      Enabled: [1, "Overdrive mode enabled (the LDO generates VOS0 for VCORE)"]
```
This is a human-readable enumeration, so it lives in `peripherals/`

---

```
SYSCFG:
  PWRCR:
    _modify:
      ODEN:
        bitWidth: 1
        description: "Overdrive enable"
```
This changes the register layout, so lives in `devices/` 

---

You should be able to leave the existing `syscfg_h747.yaml` alone, and just add `common_patches/h74x_syscfg_pwrcr.yaml` as a new file. 

For the `BDMADR` patch, `devices/common_patches/h743_hrtim_common.yaml` is good already!

andrewgazelka:
> ```
> SYSCFG:
>   PWRCR:
>     ODEN:
>       Disabled: [0, "Overdrive mode disabled"]
>       Enabled: [1, "Overdrive mode enabled (the LDO generates VOS0 for VCORE)"]
> ```
> 
> 
>      
>   
> 
> This is a human-readable enumeration, so it lives in `peripherals/`
> 
> 
> 
> ```
> SYSCFG:
>   PWRCR:
>     _modify:
>       ODEN:
>         bitWidth: 1
>         description: "Overdrive enable"
> ```
> 
> 
>      
>   
> 
> This changes the register layout, so lives in `devices/`
> 
> 
> 
> You should be able to leave the existing `syscfg_h747.yaml` alone, and just add `common_patches/h74x_syscfg_pwrcr.yaml` as a new file.
> 
> 
> For the `BDMADR` patch, `devices/common_patches/h743_hrtim_common.yaml` is good already!

Should be fixed. :) Let me know if there is anything else you would like changed.

richardeoin:
Looks good, thanks! Do you know the reason for adding the modify `name: STM32H753v` on the stm32h753v SVD? Since that 'v' isn't officially in the part number (it was a bit of a hack to track the chip revision..), I'd say we should just leave the current name of `STM32H753`

sephamorr:
> Looks good, thanks! Do you know the reason for adding the modify `name: STM32H753v` on the stm32h753v SVD? Since that 'v' isn't officially in the part number (it was a bit of a hack to track the chip revision..), I'd say we should just leave the current name of `STM32H753`

The reason for adding the -V suffix fixes a separate problem where the website (https://stm32-rs.github.io/stm32-rs/) doesn't disambiguate the H743 and H743v so only presents the H743 svd and documentation status.

andrewgazelka:
@richardeoin

richardeoin:
Understood about changing the name to fix the website, but I would prefer to do that in a separate PR. If you can keep the name the same for this PR I can get it merged

andrewgazelka:
> Understood about changing the name to fix the website, but I would prefer to do that in a separate PR. If you can keep the name the same for this PR I can get it merged

Fixed

richardeoin (approved):
bors r+

PR #652, opened 2021-10-30
H7: Add UART9/USART10 for RM0468 parts, fix base address for RM0455 parts

richardeoin:
RM0468:
Base address RM0468 Rev 2 page 137
Interrupt offset RM0468 Rev 2 page 741

RM0455:
Base address RM0455 Rev 6 page 133
Interrupt offset RM0455 Rev 6 page 722

### DIFF
--- a/devices/common_patches/h7_common_highmemory.yaml
+++ b/devices/common_patches/h7_common_highmemory.yaml
@@ -382,14 +382,14 @@ _add:
         value: 83
   UART9:
     derivedFrom: USART1
-    baseAddress: 0x40018000
+    baseAddress: 0x40011800
     interrupts:
       UART9:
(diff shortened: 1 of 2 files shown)


richardeoin:
The base address error for the rm0455 parts is a bit of a bummer since it's present in release 0.14.0

Confirmed that USART10 does transmit bytes with this PR and https://github.com/stm32-rs/stm32h7xx-hal/pull/268 on a NUCLEO-H7A3ZI-Q, and of course does not with stm32h7 0.14.0 (as it happens the HAL gets stuck waiting for ISR TXE to go high, but that's not really important).

adamgreig (approved):
bors r+

PR #653, opened 2021-10-31
Fix rename of H7 RM0455 OCTOSPI peripheral

richardeoin:
Previous modification did not work as this peripheral is `derivedFrom` OCTOSPI in the original SVD


adamgreig (approved):
bors r+

Issue #655, opened 2021-11-09
`f765`, `f7x7` and `f7x9` doesn't have `sdmmc2en` and `sdmmc2rst` fields in `RCC`.


systec-ms:
`f765`, `f7x7` and `f7x9` doesn't have `sdmmc2en` and `sdmmc2rst` fields in `RCC`.
This should be fixed in `stm32-rs` first

_Originally posted by @burrbull in https://github.com/stm32-rs/stm32f7xx-hal/pull/152#discussion_r745388947_

I haven't looked at the `f745`, and `f765`, so it could be missing somewhere else.

PR #662, opened 2021-11-20
Add SDMMC2EN and SDMMC2RST to F765, F7x7, F7x9

systec-ms:
Fix https://github.com/stm32-rs/stm32-rs/issues/655.


adamgreig (approved):
Looks good, thank you!

bors merge

Issue #641, opened 2021-10-04
G0 DMA IFCR 0.14 reversion


David-OConnor:
[0.14](https://docs.rs/stm32g0/0.14.0/stm32g0/stm32g081/dma/ifcr/index.html)
[0.13](https://docs.rs/stm32g0/0.13.0/stm32g0/stm32g081/dma/ifcr/index.html)
This was correctly AFAIK in 0.13, but now the fields aren't right.

burrbull:
Are problems in patching or svd2rust?

burrbull:
Looks like original ST SVDs are incorrect.

adamgreig:
Thanks for the bug report!

This definitely looks like a regression, I suspect from #514 but then not caught by the memory map comparison because it can't work on PRs with updated SVD files. It seems ST's SVDs for many G0s originally had this error, and it was patched in #398, but reverted in #514 because the new SVD files fixed the error. However, it looks like not all the SVD files had the error fixed (G070 and G081) and hence the wrong field values have returned.

I'll try and get this fixed and do a patch release of stm32g0 soon, though if anyone else wants to have a go at a fix please go ahead!

DerFetzer:
The current SVD release on the ST website (https://www.st.com/resource/en/svd/stm32g0_svd.zip) seems to include the missing peripherals.
@adamgreig Are you working on this already or should I give it a go? This would be my first contribution to this repo. Are there any common pitfalls I should be aware of?

nickmertin:
Did #666 fix this?

dimpolo:
> #514 seems to have some missing registers for the new variants. [docs](https://docs.rs/stm32g0/0.14.0/stm32g0/stm32g0b0/index.html) Registers like `RCC` , `PWR`, `GPIOA` or `FLASH` are not available.

This is still true ([current docs](https://docs.rs/stm32g0/0.15.1/stm32g0/stm32g0b0/index.html)) and ST's SVD haven't been updated since :/
I opened an issue with them a while ago, but never got a response.
Does someone have a good connection to them?

burrbull:
Should be fixed. Reopen if it is not.

Issue #548, opened 2021-04-27
stm32g0b1 missing RCC registers


timothyhollabaugh:
It looks like the new stm32g0b1, b0, and c1 are missing definitions for the RCC peripheral registers. The svd files are missing them as well. These new devices share a reference manual with the older stm32g0 devices, so I would assume it should be there in the same place.

Would it work to copy the RCC section from the stm32g071 svd file? Would a change this big make sense to put in the yaml files?

timothyhollabaugh:
Actually, it looks like quite a few peripherals are missing from the SVD files, at least for the stm32g0b1.

I went ahead and extracted the list of peripherals with

`cat stm32g0b1.svd | rg -A 1 '<peripheral>' | tr -d ' \t' |rg '<name>' | sort`

which gives

```
<name>CRC</name>
<name>DAC</name>
<name>DBG</name>
<name>DMAMUX</name>
<name>GPIOA</name>
<name>GPIOB</name>
<name>HDMI_CEC</name>
<name>I2C1</name>
<name>IWDG</name>
<name>LPTIM1</name>
<name>TIM14</name>
<name>TIM15</name>
<name>TIM16</name>
<name>TIM1</name>
<name>TIM2</name>
<name>TIM6</name>
<name>USART1</name>
<name>VREFBUF</name>
<name>WWDG</name>
```

Doing the same for the stm32g071 gives

```
<name>ADC</name>
<name>COMP</name>
<name>CRC</name>
<name>DAC</name>
<name>DBG</name>
<name>DMAMUX</name>
<name>DMA</name>
<name>EXTI</name>
<name>FLASH</name>
<name>GPIOA</name>
<name>GPIOB</name>
<name>HDMI_CEC</name>
<name>I2C1</name>
<name>IWDG</name>
<name>LPTIM1</name>
<name>LPUART</name>
<name>MPU</name>
<name>NVIC</name>
<name>PWR</name>
<name>RCC</name>
<name>RNG</name>
<name>RTC</name>
<name>SCB</name>
<name>SPI1</name>
<name>STK</name>
<name>SYSCFG_VREFBUF</name>
<name>TAMP</name>
<name>TIM14</name>
<name>TIM15</name>
<name>TIM16</name>
<name>TIM1</name>
<name>TIM2</name>
<name>TIM6</name>
<name>UCPD1</name>
<name>USART1</name>
<name>WWDG</name>
```

[...]

PR #666, opened 2021-11-22
Update SVDs for STM32G0B1/C1

DerFetzer:
As written in https://github.com/stm32-rs/stm32-rs/issues/641#issuecomment-974727789 I gave updating SVDs for STM32G0B1/C1 a try.

This should add the missing peripherals mentioned there and in #548.

I am not quite sure if removing the duplicate USB register enums counts as fixing the structure of the SVD or if it rather should live in the peripherals folders.


devices/common_patches/g0_usb_enum.yaml
```diff
@@ -0,0 +1,21 @@
+# Fixes duplicate enum values
+USB:
+  CNTR:
+    USBRST:
+      _replace_enum:
+        B_0x0: [0, "No effect"]
```
burrbull, line 6: In such form enums are useful. Give them meaningful name.

DerFetzer: Thanks, should be better now.

adamgreig (approved):
Thanks, LGTM!

bors merge

PR #667, opened 2021-11-23
svdtools 0.2.6

burrbull:
For https://github.com/stm32-rs/svdtools/pull/89


adamgreig:
svdtools 0.2.6 is released now, should be able to update this PR to use it

burrbull:
> 

1 minute

burrbull:
Done

adamgreig (approved):
Thanks, looks good!

I've checked manually since the mmap action won't work until after this is merged. Unfortunately the mmap generation currently doesn't take into account derived registers (https://github.com/stm32-rs/svdtools/issues/122) so it thinks the TSTR etc regs have lost all their fields, but the actual svd files (and the svd2rust generation) do the right thing.

Besides that there's no change caused by this PR, so LGTM.

bors r+

PR #668, opened 2021-11-27
Updates and fixes for stm32g4xx: FDCAN, FLASH, TIM

timblakely:
1) Add missing fields and enumerations to FDCAN via fdcan_g4.yaml
2) Add registers for FLASH handling. Note that there are some commented out since different devices have different support for secure flash based on their device ID and flash capacity.
3) Fixes for some timer values. In particular, lower non-32-bit timers back down to 16 bit, since 20 bit is only enabled via dithering.

Let me know if you prefer these to be in a different location; kind of hard for me to grok what goes where in terms of peripheral and register updates/modifications 😅


timblakely:
Gah. I'm seeing some failures in the presubmits, but I can't seem to replicate them on my end...? The pip version of svdtools doesn't seem to fail D:

devices/common_patches/fdcan/fdcan_g4.yaml
```diff
@@ -301,3 +337,170 @@ _include:
...
+      ack:        [0b11, ack]
+      bit1:       [0b100, bit1]
+      bit0:       [0b101, bit0]
+      crc:        [0b110, crc]
+      no_change:  [0b111, no_change]
+
+  HPMS:
+    _modify:
+      FIDX:
+        msb: 12
```
burrbull, line 416: STM32 usually use `bitOffset/bitWidth` instead of `msb/lsb`

adamgreig:
Thanks for the PR and your other recent G4 work!

The failures are coming from the linter which checks the generated SVD complies with the CMSIS-SVD schema: if you run `make lint` locally it should hopefully reproduce the errors locally (you need `xmllint` installed).

I think the main problem is that where you've used `lsb` and `msb`, the old `bitOffset` and `bitWidth` tags still exist for those elements, creating an invalid SVD that specifies both sets instead of only one or the other. So far in stm32-rs we've exclusively used `bitOffset` and `bitWidth` (although the svdtools patcher does know about `lsb` and `msb`), so maybe the best thing to do is keep using `bitOffset` and `bitWidth` here for consistency. To fix the failure otherwise you'd have to explicitly delete the existing `bitOffset` and `bitWidth` tags too.

Besides that, on lines like:

```yaml
    accept_rx_fifo0:  [0b00, accept_rx_fifo0]
````

the last item should be a string description of the field, not just its name repeated: something like `"Accept in Rx FIFO 0"` or even `"Accept non-matching extended frames into Rx FIFO 0"`. It's what's used to generate the documentation for this enum variant.

> Let me know if you prefer these to be in a different location; kind of hard for me to grok what goes where in terms of peripheral and register updates/modifications sweat_smile

Yep, this is pretty tough to track as the project evolved! The gist is that anything "structural" i.e. changing registers or fields should go into `common_patches`, and anything that adds enumerated values (or permitted write ranges) to a field goes into `peripherals/`. So, in this case, the various enums you've added should be split out from the patch file and put into a new file inside `peripherals/`.

**Comment by timblakely** [comment]
`Thanks for the PR and your other recent G4 work!`

My pleasure! Seeing as the G4 is likely the only chip I have ~~forever~~ for the near future thanks to the global semiconductor apocalypse I figured I might as well contribute to it 🥴 

I've updated the PR and run the linter. PTAL! :)

**Comment by timblakely** [comment]
Let me know if there's something I need to do to trigger the review notification again (happened last time I submitted a PR too, so I'm probably just missing a step in the flow somewhere, apologies!)

File: peripherals/flash/flash_g4.yaml [hunk 2]
```diff
@@ -0,0 +1,131 @@
...
+        name: SRAM_PE
+    _add:
+      DBANK:
+        description: Single or dual bank mode
+        bitOffset: 22
+        bitWidth: 1
+      BFB2:
+        description: Dual-bank boot
+        bitOffset: 20
+        bitWidth: 1
```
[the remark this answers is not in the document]
richardeoin: These bits are also quite annoying because their presence depends on the category of the device (Section 3, 4 or 5 of RM0440). OPTR bit 20 even gets re-used for two different functions 🤦  

I would at least put them in a separate file under `common_patches` and include "category_3" in the name so we can decide what to do with them

peripherals/can/fdcan.yaml
```diff
@@ -0,0 +1,95 @@
...
+      sample_point: [0b01, "Sample point can be monitored at pin FDCANx_TX"]
+      dominant:     [0b10, "Dominant (0) level at pin FDCANx_TX"]
+      recessive:    [0b11, "Recessive (1) at pin FDCANx_TX"]
+
+  CCCR:
+    TEST:
+      normal:       [0b0, "Normal operation, register TEST holds reset values"]
+      test:         [0b1, "Test Mode, write access to register TEST enabled"]
+    DAR:
+      retransmit:   [0b0, "Automatric retransmission of messages not transmitted successfully enabled"]
```
richardeoin, line 17: Automatic

richardeoin (commented):
Don't worry, this PR hasn't been forgotten! I had a look through, and I didn't spot any errors compared to RM0440 🎉   Nice work!

I've noted where declarations be moved to match the rules Adam mentioned earlier, and we need to decide what to do with those flash CR / OPTR bits that only appear based on yet another configuration item

burrbull:
@timblakely do you plan to finish the PR?

PR #669, opened 2021-11-29
devices/stm32l4r9.yaml: patch TIM3 and TIM4

jspngh:
TIM3 and TIM4 work the same way in stm32l4r9 and stm32l4x6.
Since `stm32l4x6.yaml` contains these `_copy` and `_derive` lines, they should also be added to this yaml file.

Note that I've simply verified this empirically, with these lines added I can now use TIM3 and TIM4 to generate a PWM signal,
and that I don't fully comprehend the SVD files yet.

### DIFF
--- a/devices/stm32l4r9.yaml
+++ b/devices/stm32l4r9.yaml
@@ -28,6 +28,13 @@ RCC:
         name: SPI3EN
         description: SPI peripheral 3 clock enable
 
+_copy:
+  TIM3:
+    from: TIM2
+
+_derive:
+  TIM4: TIM3
+
 MPU:
   _strip:
     - "MPU_"


jspngh:
> [Memory map comparison](https://github.com/stm32-rs/stm32-rs-mmaps/compare/pr-669-f6afb30)

That's pretty neat.

adamgreig (approved):
Thanks for this PR!

Ideally I guess TIM3 and TIM4 should not have the high fields copied from TIM2, but this is strictly better than the current setup where they just have a full 32-bit register, so let's get this in now and it could always be improved further later.

bors merge

adamgreig:
Spurious network error

bors retry

PR #671, opened 2021-12-05
Add bit ranges to HDMI CEC registers

newAM:
This applies to the H7, F0x{1,2,8}, F7{6,7} which share identical register layouts.

### DIFF
--- a/devices/common_patches/hdmi_cec.yaml
+++ b/devices/common_patches/hdmi_cec.yaml
@@ -1,0 +1,6 @@
+CEC:
+  CFGR:
+    OAR: [0, 0x7FFF]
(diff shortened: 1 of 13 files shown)


devices/common_patches/hdmi_cec.yaml
```diff
@@ -0,0 +1,6 @@
+CEC:
+  CFGR:
+    OAR: [0, 0xEFFF]
```
adamgreig, line 3: should this be 0x7FFF?

newAM: Yes it should, good catch!

adamgreig (approved):
thanks!

bors merge

Issue #670, opened 2021-12-03
OTG_HS interrupt possibly missing from stm32h7xx_hal::interrupt enum for STM32H723ZG


TomDeRybel:
I'm porting over some code to a ST Nucleo-H723ZG board and ran into a little snag, shown below. The other interrupts used in the code compile fine (eg: EXTI15_10, TIM1_UP).

The OTG_HS interrupt and friends are listed on the documentation website and worked fine on the H743 I used before.
https://docs.rs/stm32h7xx-hal/latest/stm32h7xx_hal/enum.interrupt.html#variant.OTG_HS

However, none of the USB interrupts show up in the documentation for the STM32H735 crate, generated with cargo doc.

`error[E0599]: no variant or associated item named `OTG_HS` found for enum `stm32h7xx_hal::interrupt` in the current scope
   --> src/main.rs:251:17
    |
251 |     #[task(binds = OTG_HS, shared = [usb_dev, usb_cdc])]
    |                    ^^^^^^ variant or associated item not found in `stm32h7xx_hal::interrupt`
`

I'm using the following versions/cargo lines:
`stm32h7 = { git = "https://github.com/stm32-rs/stm32-rs-nightlies", features = ["stm32h735","rt"] }`
`stm32h7xx-hal = { git = "https://github.com/stm32-rs/stm32h7xx-hal", features = ["stm32h735","rt","usb_hs"] }`

As an aside, some register field names seem to have changed between the H743 and H725. Not sure if this is intentional:
H743
`ctx.device.DBGMCU.cr.modify(|_, w| {
			w.dbgsleep_d1().set_bit();
			w.dbgstby_d1().set_bit();
			w.dbgstop_d1().set_bit()
		});`

H723
`ctx.device.DBGMCU.cr.modify(|_, w| {
			w.dbgslpd1().set_bit();
			w.dbgstbd1().set_bit();
			w.dbgstpd1().set_bit()
		});`

Thank you for all the work done on these chips! These crates make them so much more friendly to use.

adamgreig:
Thanks for reporting, this looks like an error in the H735 SVD file from ST. If you'd like to give fixing it a go, the fix is to add new interrupts to the `devices/stm32h735.yaml` file in this repository, something like:

```yaml
OTG_HS:
    _add:
        _interrupts:
            OTG_HS:
                value: 77
                description: OTG_HS global interrupt
```

and so forth.

[...]

PR #677, opened 2021-12-08
STM32H735 yaml fixes

TomDeRybel:
This pull request is related to issue: https://github.com/stm32-rs/stm32-rs/issues/670

In this pull request are a bunch of fixes for the STM32H735-family of parts. This mainly concerns the stm32h735.yaml file.

Note: This is my first time (attempting) to contribute to stm32-rs, and there are several TODO's in the code, asking questions. This pull request is likely not good enough as-is.

What it does:
- There were about 19 missing interrupts. These were all added, though several are left commented-out because I'm uncertain on how to integrate them properly. (To find these missing interrupts, I manually compared the entire interrupt vector table (Table 140) in RM0468 to what I could find in the mod.rs for the device.)
- There were several naming discrepancies. These were aligned to RM0468.
- Entire peripherals were missing. I have added CORDIC and FMAC (as I'll be using those in the future, and will thus be able to test properly). There are more to be done, as well as variants on others (such as I2C5, UART9, USART10).
- The USB OTG-HS interrupts are also fixed, but not satisfactory. They should be assigned under the OTG-HS block, for consistency, but I don't know how to do that properly.

I hope this can be of some use.


adamgreig:
Thanks for this PR! I'll need a bit of time to review it properly but I can give you some pointers first:

* you might find the `svd interrupts` tool useful (part of svdtools), it outputs a sorted list of interrupts for a specific svd file (original or patched), and shows which are are missing in sequence order too. It also shows which peripheral an interrupt is associated with.

> # TODO: The blocks that are commented-out below should either go into an existing
> #       block, or have a new block made for them. It's probably better that I 
> #       don't improvise...

By "block" do you mean in the YAML file or an SVD peripheral? The peripherals probably already exist in the SVD so long as you use the right name.

TomDeRybel:
Thank you for the tips and positive response. Working on these few definitions in the PR really drove home just how important this crate really is to get anything to work right with these processors, and the amount of work that has already been done to enable this.

Unfortunately, I seem to have used the word "block" to mean multiple things, depending on the context.

What I meant is that some of the peripheral modifications I included in the PR YAML file, I could not assign to a particular, pre-existing peripheral, and they were thus left commented-out. This is either because:
- I couldn't find the correct, pre-existing name. Or was too unsure about it.
- There was no pre-existing name, but it seemed closely related to existing peripheral family (eg: USART10, I2C5) and probably needs to be included under their umbrella somehow.

adamgreig (approved):
Thanks for this PR! I've gone through the changes which all look good. I rebased to fix merge conflicts with master and tidied up the yaml file a little (enabled a couple of interrupts, removed some that have since been added separately, removed trailing whitespace, etc).

bors r+

TomDeRybel:
Thank you! Happy to see that it was useful.

PR #678, opened 2021-12-15
Add enumerated values for STM32F405/415/407/417/427/437/429/439 TIM2 ITR1_RMP

dgoodlad:
These values are from STMicro RM0090, 18.4.19:

image

This reference manual covers "STM32F405/415, STM32F407/417, STM32F427/437 and STM32F429/439", so it's likely that this change could apply to those other devices as well. I don't have one to test, though, so defer to maintainers' preferences here. ~~Happy to extract this to a common patch if folks are happy with that approach.~~ EDIT: I've pushed 3a7e8e4 which does just this.


adamgreig (approved):
Thanks!

bors r+

PR #679, opened 2021-12-16
Fix make html on macOS with python 3.8+

sephamorr:
Python3.8 and above [changed](https://chrissardegna.com/blog/multiprocessing-changes-python-3-8/) the behavior of the multiprocessing module on macOS, resulting in variable scoping being tighter than on other platforms or versions.

As a simplified case, this code crashes on such a machine:

```python
import multiprocessing

def worker(local):
	print(v)

if __name__ == "__main__":
	v = "can i access this?"
	with multiprocessing.Pool() as p:
		devices = p.map(worker, [None])
```

This patch passes in the relevant variable into the worker in a different style to fix. I've tested this on linux and Mac platforms for regressions.

### DIFF
--- a/scripts/makehtml.py
+++ b/scripts/makehtml.py
@@ -304,16 +304,17 @@ def process_svd(svdfile):
     return device
 
 
-def generate_if_newer(device):
+def generate_if_newer(arg):
+    device, htmldir = arg
     pagename = "{}.html".format(device["name"])
-    filename = os.path.join(args.htmldir, pagename)
+    filename = os.path.join(htmldir, pagename)
     isfile = os.path.isfile(filename)
     if not isfile or os.stat(filename).st_mtime < device['last-modified']:
         page = generate_device_page(device)
         print("Generating", pagename)
         with open(filename, "w", encoding='utf-8') as f:
             f.write(page)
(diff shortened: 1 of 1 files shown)


scripts/makehtml.py
```diff
@@ -322,10 +323,9 @@ def generate_if_newer(device):
     parser.add_argument("svdfiles", help="Path to patched SVD files", nargs="*")
     args = parser.parse_args()
     devices = {}
-
     with multiprocessing.pool.ThreadPool() if sys.platform == 'win32' else multiprocessing.Pool() as p:
         devices = p.map(process_svd, args.svdfiles)
-        p.map(generate_if_newer, devices)
+        p.map(generate_if_newer, [(device,args.htmldir) for device in devices])
```
adamgreig, line 328:
```suggestion
        p.map(generate_if_newer, [(device, args.htmldir) for device in devices])
```

adamgreig (commented):
Thanks, this looks good! Any reason it's still marked as a draft?

Only tiny comment is I like to stick to PEP8 which would complain about this missing space.

sephamorr:
Fixed and marked for review at your convenience.

adamgreig (approved):
thanks!

bors merge

Issue #342, opened 2020-02-26
STM32L4x5/6 Missing Registers 


unizippro:
According to the [reference manual](https://www.st.com/resource/en/reference_manual/dm00083560.pdf) page 309-310, the GPIO port bit reset register `GPIOx_BRR` and the GPIO port analog switch control register `GPIOx_ASCR` register should exist, under the respective GPIOs. They are in the SVD but are missing in the [library](https://stm32.agg.io/rs/STM32L4x5.html#GPIOA).

unizippro:
Likewise the Prescaler field under ADC common control register`ADC_CCR` is also missing. Reference manual page 613.

PR #680, opened 2021-12-21
STM32L4xx: Add GPIOx ASCR & BRR registers

MathiasKoch:
Also corrects naming on L4x5 ADC `Mult` to `Dual` field, as per the reference manual, and adds prescaler field under ADC_CRR.

Fixes https://github.com/stm32-rs/stm32-rs/issues/342

### DIFF
--- a/devices/common_patches/l4_gpio_ascr_brr.yaml
+++ b/devices/common_patches/l4_gpio_ascr_brr.yaml
@@ -1,0 +1,20 @@
+"GPIO*":
+  _add:
+    ASCR:
+      description: GPIO port analog switch control register
+      addressOffset: 0x2C
+      fields:
+        ASC:
+          description: These bits are written by software to configure the analog connection of the IOs.
+          access: read-write
+          bitOffset: 0
+          bitWidth: 16
+    BRR:
+      description: GPIO port bit reset register
+      addressOffset: 0x28
+      fields:
+        BR:
+          description: These bits are write-only. A read to these bits returns the value 0x0000
+          access: read-only
+          bitOffset: 0
+          bitWidth: 16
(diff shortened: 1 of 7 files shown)


devices/common_patches/l4_gpio_ascr_brr.yaml
```diff
@@ -0,0 +1,20 @@
...
+          access: read-write
+          bitOffset: 0
+          bitWidth: 16
+    BRR:
+      description: GPIO port bit reset register
+      addressOffset: 0x28
+      fields:
+        BR:
+          description: These bits are write-only. A read to these bits returns the value 0x0000
+          access: read-only
```
richardeoin, line 18: Guess this should be `write-only`

richardeoin (changes requested):
Thanks for the PR! Couple of comments:
* The RM describes the BRR / ASCR registers with a series of one-bit fields `BRR0`, `BRR1`, ... and this is the case for all the other GPIO register definitions also. I think this would be preferred instead of a 16-bit field for consistency if nothing else
* BRR is also missing on stm32l4x1, stm32l4x2, stm32l412, stm32l4x3. These are parts that don't have ASCR, so it would make sense to split the patch file and apply the BRR section to stm32l4x1.. also

MathiasKoch:
@richardeoin Thanks for the feedback.

I believe this should address all your points?

richardeoin (approved):
Perfect, thanks!

bors r+

richardeoin (approved):
Tidied merge conflict

bors r+

richardeoin:
Not sure why CI failed. Let's retry

bors retry

burrbull:
Enums in common_patches

Issue #382, opened 2020-05-14
DBGMCU naming inconsistencies


unizippro:
The `APB` freeze registers are named with underscore in [`stm32l4x5-6`](https://stm32-rs.github.io/stm32-rs/STM32L4x5.html#DBGMCU:APB1_FZR1) where as it are without in [`stm32l4x1-3`](https://stm32-rs.github.io/stm32-rs/STM32L4x1.html#DBGMCU:APB1FZR1).
They are named with in all other processors i checked.

PR #681, opened 2021-12-21
STM32L4x1/L4x2/L4x3: Rename DBGMCU APB_FZR registers to match general convention

MathiasKoch:
Fixes https://github.com/stm32-rs/stm32-rs/issues/382

### DIFF
--- a/devices/common_patches/l4_dbg_apb_fzr_rename.yaml
+++ b/devices/common_patches/l4_dbg_apb_fzr_rename.yaml
@@ -1,0 +1,8 @@
+DBGMCU:
+  _modify:
+    APB1FZR1:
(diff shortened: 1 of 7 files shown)


richardeoin:
Thanks for responding to the issue with this fix. The reference manuals call these APBxFZR registers (without the underscore), so I think the best fix would actually be to remove the underscore on stm32l4x5, stm32l4x6 and stm32l4r9 instead! Would you consider doing that?

It would then be consistent with the L5, WLE5.. parts which also have xxxxFZR registers without underscore

MathiasKoch:
That should do it :+1:

richardeoin (approved):
Looks good!

bors r+

PR #687, opened 2022-01-17
G0

burrbull:


richardeoin (approved):
Thanks for the PR! I've had a look through this and I don't see any issues. So let's go ahead and merge.

The naming for ADC offset 0x28 is consistent now - `CHSELR0` and `CHSELR1`. These are aliases, but svd2rust handles this acceptably with method accessors rather than fields.

bors r+

PR #690, opened 2022-02-02
Make FLASH_WRP??R and FLASH_SECR writable

davidlattimore:
See reference manual for [STM32G0x0](https://www.st.com/resource/en/reference_manual/rm0454-stm32g0x0-advanced-armbased-32bit-mcus-stmicroelectronics.pdf) and [STM32G0x1](https://www.st.com/resource/en/reference_manual/rm0444-stm32g0x1-advanced-armbased-32bit-mcus-stmicroelectronics.pdf). Both define FLASH_WRP1AR etc as read-write (but read-only in the SVD). STM32G0x1 defines FLASH_SECR as read-write (but read-only in the SVD). The STM32G0x0 doesn't have a FLASH_SECR register.

### DIFF
--- a/devices/common_patches/g0_flash.yaml
+++ b/devices/common_patches/g0_flash.yaml
@@ -1,0 +1,15 @@
+FLASH:
+  _modify:
+    # Fix access and reset values.
+    WRP??R:
+      access: read-write
+      resetValue: 0x000000ff
+    PCROP??SR:
+      access: read-write
+      resetValue: 0xffffffff
+    PCROP??ER:
+      access: read-write
+      resetValue: 0x00000000
+    SECR:
+      access: read-write
+      resetValue: 0x00000000
--- a/devices/stm32g030.yaml
+++ b/devices/stm32g030.yaml
@@ -36,3 +36,4 @@ _include:
  - common_patches/tim/common.yaml
  - ../peripherals/tim/tim_basic.yaml
  - ../peripherals/tim/tim_advanced.yaml
+ - ./common_patches/g0_flash.yaml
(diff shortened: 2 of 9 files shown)


richardeoin:
Thanks for the PR! This looks good, couple of comments:

* Looks like the `PCROP1??R` registers are also erroneously `read-only`
* The `resetValue` for these registers are very wrong in the SVD, perhaps safer to at least make them zero. Alternatively use the ST production value listed in the RM.
* This patch doesn't apply to a few parts (`g050`, `g051`, `g061`, `g0b0`) as they don't have flash blocks in the SVD. 👀  That can be out of scope for this PR, but should be fixed eventually.
* Could you rebase after #687 was merged? The rebase should be simple (choose both)

davidlattimore:
I rebased, updated to include PCROP* and fixed the reset values. I went with the reset values from the datasheet, since for write protection and ROP, the end address is inclusive. So setting WRP to 0 would mean to write protect the first page of flash - something that might be surprising.

The value I've observed for WRP1AR on an stm32g071 is 0x3f, not 0xff as the datasheet says. The g071 has 128KB of flash, so 0x3f is the last page of flash. i.e. start=0x3f, end=0 (don't write protect anything).  I tried setting it to 0xff and the value (after applying option bytes and reloading) ended up as 0x3f anyway. I think it probably makes sense that the reset value is 0xff, since that would presumably accommodate all devices in the g0 series.

burrbull:
> The value I've observed for WRP1AR on an stm32g071 is 0x3f, not 0xff as the datasheet says.

If something is really different from RM, it is better to add detailed comment to prevent somebody "fixing" your fix.

davidlattimore:
> > The value I've observed for WRP1AR on an stm32g071 is 0x3f, not 0xff as the datasheet says.
> 
> If something is really different from RM, it is better to add detailed comment to prevent somebody "fixing" your fix.

To clarify, the value I put in the patch is the value from the RM, not the value that I read from the device - but I think either value would likely be fine - in fact any value where start > end should be fine.

PR #696, opened 2022-02-06
F301: Add missing P to JADST

Sh3Rm4n:
The f301 is the only device in the F3 family, which has the field name
wrong. In the datasheet [RM0366] 12.5.3 `JADSTP` is also used.

[RM0366]: https://www.st.com/resource/en/reference_manual/dm00094350.pdf

### DIFF
--- a/devices/stm32f301.yaml
+++ b/devices/stm32f301.yaml
@@ -45,6 +45,8 @@ ADC1:
     _modify:
       '*':
         access: read-write
+      JADST:
+        name: JADSTP
 
 ADC1_2:
   _delete:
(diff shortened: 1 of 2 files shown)


richardeoin (approved):
Thanks, looks good.

In case anyone missed it, the reason for the peripheral rename is in the commit message

> Only the peripheral is named ADC_Common. ADC1_2 is consistently used for
> every F3 family device except F3x4 (and F373, which has no common ADC).
> 
> The address is the same compared to the other devices and in other
> places ADC12 is already used for some fields and descriptions

bors r+

PR #698, opened 2022-02-08
Fix missing author name in APACHE license

Sh3Rm4n:
Hey @adamgreig I've noticed that your name is not mentioned in the Apache license, while being in the MIT license.

I've copied name and year from the MIT license.

### DIFF
--- a/LICENSE-APACHE
+++ b/LICENSE-APACHE
@@ -186,7 +186,7 @@ APPENDIX: How to apply the Apache License to your work.
    same "printed page" as the copyright notice for easier
    identification within third-party archives.
 
-Copyright [yyyy] [name of copyright owner]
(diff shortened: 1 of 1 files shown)


adamgreig:
Thank you for the PR! I think that part of the Apache licence is deliberately meant to be left as a placeholder; it's a template for other people to see how to apply the licence to their own work. Ideally that text should be inserted into every relevant file in the project with the copyright details filled in. In practice we don't do that here and assume it suffices to clearly state that the project is licensed under both licenses; we only really need one copyright statement so I guess having it in the MIT licence file will suffice.

Sh3Rm4n:
Alright, good to know. Learned something new :) Makes sense.

PR #699, opened 2022-02-09
Fix ADC disable

Sh3Rm4n:
`ADDIS`, as `ADEN`, is a signal to the ADC peripheral to change it's
state. Both signals have to be "1" to change the state.

In case of `ADDIS` it will be only reset to "0", after the state has
successfully changed.

The procedure of disabling is the same between `stm32f303xc` (RM0316
15.3.10), where I've noticed the bug, and `stm32h7x3` (RM0443 25.4.9),
for which this code was initially committed, so I assume this affects
every device using the `adc_v3.yaml`.

### DIFF
--- a/peripherals/adc/adc_v3.yaml
+++ b/peripherals/adc/adc_v3.yaml
@@ -97,7 +97,7 @@
       Start: [1, "Starts conversion of channel"]
     ADDIS:
       _write:
-        Disable: [0, "Disable ADC conversion and go to power down mode"]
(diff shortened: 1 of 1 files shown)


richardeoin:
This is an error in the enumerations, thanks for highlighting it. We actually have better enumerations for these bits in `CR` in `adc_l0.yaml` (among others) already, so I'm going to add those on top also

richardeoin (approved):
bors r+

PR #700, opened 2022-02-10
F7/tim: access to bits() without unsafe for single-field registers

maximeborges:
When writing to  `CNT`/`ARR`/`CCR%s` single-field registers via `bits()`, it requires the use of unsafe for timers 9 to 14.
This PR allows to use safe `bits()` writing just like the other timers


adamgreig (approved):
Makes sense, thanks!

bors r+

PR #701, opened 2022-02-11
Rust svdtools for CI

burrbull:


burrbull:
bors try

burrbull:
@adamgreig Can't test mmaps as they action uses old version of ci.

.github/workflows/ci.yaml
```diff
@@ -28,51 +32,44 @@ jobs:
...
+
+      - name: Cache dependencies
+        uses: Swatinem/rust-cache@v1
         with:
-          crate: svd2rust
-          version: 0.21.0
-      - name: Copy svd2rust to cache directory
-        if: steps.cache-cargo.outputs.cache-hit != 'true'
+          working-directory: ${{ matrix.crate }}
+          key: svdtools-${{ env.SVDTOOLS_VERSION }}-svd2tust-${{ env.SVD2RUST_VERSION }}
```
adamgreig, line 60:
```suggestion
          key: svdtools-${{ env.SVDTOOLS_VERSION }}-svd2rust-${{ env.SVD2RUST_VERSION }}
```

.github/workflows/ci.yaml
```diff
@@ -28,51 +32,44 @@ jobs:
...
-      - name: Put new cargo binary directory into path
-        run: echo "~/cargo-bin" >> $GITHUB_PATH
+          cargo install svdtools --version $SVDTOOLS_VERSION --target-dir deps-target
+          cargo install svd2rust --version $SVD2RUST_VERSION --target-dir deps-target
+          rm -rf deps-target
+
       - name: Patch SVDs
         run: make -j2 patch
-      - name: Check SVDs
-        run: make lint
```
adamgreig, line 74: I'd rather keep the lint step if we can, why remove it?

adamgreig (changes requested):
Thanks for this PR, it will be good to get CI swapped over to the Rust tools.

I don't think the caching is doing what I want - the old caching actions were to cache the svd2rust binary (and could also cache the svdtools binary); I don't really care about caching the crate's own dependencies as they're extremely small and quick to build, while building svd2rust from source every time takes ages. I think the rust-cache action used here would not cache svd2rust and only cache things like bare-metal and cortex-m, which isn't as useful.

.github/workflows/ci.yaml
burrbull: Why? It's only wastes time and do nothing already made by `svd_encode::Encode` (formats during encoding) and `svd2rust --strict` (makes a lot of checks)

adamgreig: It checks the SVDs definitely comply with the published SVD specification though, which has caught errors in tooling before and might again in the future. If it passes we know the patched SVDs are valid, even if they're not going to be used by svd2rust.

adamgreig: bors just checks CI (we don't have a separate CI job just for bors now) and there's no separate CI on publish either, so having it here just makes sure it does pass with bors or publish. Is it very slow?

burrbull:
> I think the rust-cache action used here would not cache svd2rust and only cache things like bare-metal and cortex-m, which isn't as useful.

It caches both project and contents of `~/.cargo/bin` etc.

adamgreig:
> It caches both project and contents of `~/.cargo/bin` etc.

Oh, great, that's fine then. It looks like that was added to the action recently and isn't in the readme/documentation yet, so I missed it. Much nicer than the path hackery I had before!

.github/workflows/ci.yaml
```diff
@@ -28,51 +32,50 @@ jobs:
...
+
+      - name: Install tools
         run: |
-          mkdir ~/cargo-bin
-          cp ~/.cargo/bin/svd2rust ~/cargo-bin
-      - name: Put new cargo binary directory into path
-        run: echo "~/cargo-bin" >> $GITHUB_PATH
+          cargo install svdtools --version $SVDTOOLS_VERSION --target-dir deps-target
+          cargo install svd2rust --version $SVD2RUST_VERSION --target-dir deps-target
+          rm -rf deps-target
```
adamgreig, line 69: Looking good. What's the point of using deps-target and deleting it?

adamgreig:
It looks like the XML linting is catching a bunch of invalid SVDs... :grimacing:

burrbull:
> It looks like the XML linting is catching a bunch of invalid SVDs... grimacing

Or correct SVDs and invalid linting.

.github/workflows/ci.yaml
burrbull: 1 target dir for rule them all.
Umm. For speed up builds when rebuild both.
Cleaning to make cache smaller.

adamgreig: Makes sense, thanks for explaining.

adamgreig:
The schema file is provided by Arm and is the definition of a correct SVD, so I don't think the linting can be wrong here. Apparently a `dimIndex` field must be a `dimIndexType` which is defined as matching

`[0-9]+\-[0-9]+|[A-Z]-[A-Z]|[_0-9a-zA-Z]+(,\s*[_0-9a-zA-Z]+)+`

so `1` isn't valid.

adamgreig:
Though I guess we have found a bug in the XML schema before... https://github.com/stm32-rs/stm32-rs/blob/master/svd/cmsis-svd.xsd#L26-L28

burrbull:
> The schema file is provided by Arm and is the definition of a correct SVD, so I don't think the linting can be wrong here. Apparently a `dimIndex` field must be a `dimIndexType` which is defined as matching
> 
> `[0-9]+\-[0-9]+|[A-Z]-[A-Z]|[_0-9a-zA-Z]+(,\s*[_0-9a-zA-Z]+)+`
> 
> so `1` isn't valid.

I can make workaround for this in `svd-encode` like I already done in python's `collect_in_array`. But I still think 1-element arrays should be supported natively.

.github/workflows/ci.yaml
burrbull: Oh. Cleaning doesn't need as it is custom dir now. I've added it early when used `target` for this.

adamgreig (commented):
Thanks! I think this is pretty close now but I have a couple of questions about the cache:

* Right now cargo reinstalls svdtools, svd2rust, and form every time the job runs, because that step isn't disabled if the cache was present. We could add an `if` to that step that skips it if the cache was loaded, like it used to have, otherwise we have to recompile svd2rust every time. The rust-cache action has a `cache-hit` output that should work for this just like the old step.

* Right now the nightlies job uses form, but uses the same cache key as the CI job, so if we did this it might load a cache that doesn't have form installed. Maybe it's better to have all the jobs install svdtools, svd2rust, and form, and then share a cache key, and only re-install when changed, so we have a single cache with all three tools in it. Or use a separate cache key for the nightlies job.

burrbull:
> Thanks! I think this is pretty close now but I have a couple of questions about the cache:
> 
>     * Right now cargo reinstalls svdtools, svd2rust, and form every time the job runs, because that step isn't disabled if the cache was present. We could add an `if` to that step that skips it if the cache was loaded, like it used to have, otherwise we have to recompile svd2rust every time. The rust-cache action has a `cache-hit` output that should work for this just like the old step.
> 
>     * Right now the nightlies job uses form, but uses the same cache key as the CI job, so if we did this it might load a cache that doesn't have form installed. Maybe it's better to have all the jobs install svdtools, svd2rust, and form, and then share a cache key, and only re-install when changed, so we have a single cache with all three tools in it. Or use a separate cache key for the nightlies job.

On second question I doesn't have answer as I haven't test it.
Of first question. It rebuilds tools only when create new cache (when change cache name). Then:
```
 Downloading crates ...
  Downloaded svdtools v0.2.1
     Ignored package `svdtools v0.2.1` is already installed, use --force to override
 Downloading crates ...
  Downloaded svd2rust v0.21.0
     Ignored package `svd2rust v0.21.0` is already installed, use --force to override
```

adamgreig:
Ah, cool, that makes sense! Very neat.

I think this means the nightlies job will reinstall form every time instead of using the cached one, so maybe we should install form in the main CI job too so that it will be present in the cache. It adds a little time when the cache is rebuilt, but shouldn't be too bad normally. I don't think we need to put the form version in the key or anything.

adamgreig:
bors try

adamgreig:
Ah, that's annoying. Maybe just put "form-0.8.0" into the cache key after all, since that will trigger it to build it properly now?

adamgreig:
bors try

adamgreig (approved):
Thanks again!

bors merge

PR #702, opened 2022-02-15
Change H7 DMA CR patch to only include SxCR registers

omion:
I noticed that the patch to add a TRBUFF field to the H7 DMA SxCR registers also erroneously adds it to SxFCR, LIFCR, and HIFCR registers. This PR restricts it to only the SxCR registers.

It's not much of an issue since the incorrect fields are easily ignorable, but it is also easily fixable.


adamgreig (approved):
Good catch, thanks!

bors merge

Issue #694, opened 2022-02-04
lpuart: L4R9 missing PRESC reg


Dirbaio:

burrbull:
https://stm32-rs.github.io/stm32-rs/STM32L4R9.html#LPUART1

PR #703, opened 2022-02-26
Add stm32l4r5

gauteh:
- add stm32l4r5 device
- rename CH7WDATR (the second) to CH7DATINR
- l4r5: copy ADC1 from l4x5

### DIFF
--- a/devices/stm32l4r5.yaml
+++ b/devices/stm32l4r5.yaml
@@ -24,6 +24,13 @@ _modify:
   ADC:
     name: ADC1
 
+_copy:
+  TIM3:
+    from: TIM2
+
+_derive:
+  TIM4: TIM3
+
 DFSDM1:
   _delete:
     - CH7WDATR
--- a/devices/stm32l4r5.yaml
+++ b/devices/stm32l4r5.yaml
@@ -1,0 +1,84 @@
+_svd: ../svd/stm32l4r5.svd
+
+# Merge the thousands of individal bit fields into a single field for each
+# CAN filter register. This is not only much easier to use but also saves
+# a huge amount of filespace and compilation time etc -- as much as 30% of all
+# fields in many devices are just these CAN filter bank fields.
+"CAN*":
+  "F?R?":
+    _merge:
+      - "FB*"
+  "F??R?":
+    _merge:
+      - "FB*"
+
+# SVD incorrectly labels APB1ENR1 bit 15 as SP3EN instead of SPI3EN.
+RCC:
+  APB1ENR1:
+    _modify:
+      SP3EN:
+        name: SPI3EN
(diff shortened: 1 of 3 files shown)


adamgreig:
So far this looks pretty reasonable to me, let me know if you have any questions or any problems as you get it ready!

gauteh:
Thanks, I think it is good, I'll do a bit more testing and report back. But should otherwise be ready (for review) from my side.

gauteh:
Ok, I think the easiest is to go with ADC1 for the peripheral. The interrupt is already named ADC1. But it would make it more consistent to rename the ADC on the r9 to ADC1 as well (@jspngh might have objections to that).

jspngh:
I have no problem with renaming it to `ADC1` for all L4+ devices

gauteh:
Hi, I know you are all busy. I think this one is ready if you have the chance to look at it. Thanks, Gaute

richardeoin (commented):
I suspect in general we should prefer to merge PRs like this, and fix errors later. But I did find a few items to fix - and might apply to other L4 parts also.

Reference manual: RM0432 rev 9

* ADC3 interrupt when there's no adc3 peripheral, should be deleted. It's under ADC1 in the SVD, so you can fix it with something like:
```yaml
ADC1:
  _delete:
    _interrupts:
      - ADC3
```
* TIM1/8 `OR1` registers both have a field `ETR_ADC3_RMP` that doesn't exist in the RM.
* TIM1/8 `OR1` register description is wrong
* `OSPI1` bits are missing from RCC `AHB3ENR` `AHB3RSTR` `AHB3SMENR` (at least some R5 parts have two octospi, although it is hard to tell sometimes)
* RCC `CFGR` -> `MCOSEL` is 3 bits wide, should be 4 (and there is a code `1000` that's used for the HSI48)
* RCC missing `DLYCFGR` register
* PWR missing `PDCRI` `PUCRI` `CR5` registers
* Flash `ACR` -> `LATENCY` is 3 bits wide, should be 4
* Flash `SR` missing `PEMPTY` bit
* Flash `ECCR` bit broken, `ADDR_ECC` is too small and then `BK_ECC` `SYSF_ECC` at the wrong offsets
* Flash missing `CFGR` register

richardeoin:
Also a missing register on LPUART1: https://github.com/stm32-rs/stm32-rs/issues/694

richardeoin:
@gauteh Do you plan to respond to the review feedback in this PR? This part does look like a useful one to support

gauteh:
Hi, to be honest I don't know when I will have time to look into it. It all
seems sensible.

søn. 15. mai 2022, 13:25 skrev Richard Meadows ***@***.***>:

richardeoin:
Closed with #740 🎉

PR #704, opened 2022-02-27
H7 flash: Add registers that are mirrored between banks for rm0399

richardeoin:
See RM0399 Rev 3 Table 19. These mirrored registers are present on other H7 parts

Including these registers changes the svd2rust access style for bank2, from a public member to a method that returns a reference. This makes it consistent with other H7 parts

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Family-specific:
 
 * H7:
     * h7b3: clear all enumeratedValues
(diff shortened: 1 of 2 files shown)


richardeoin:
Table 19:

image

richardeoin:
This helps https://github.com/stm32-rs/stm32h7xx-hal/pull/322 to support all H7 parts

adamgreig (approved):
Thanks!

bors r+

PR #706, opened 2022-03-08
Apply OTG_FS v1 fixes to F215/F217

Gekkio:
These devices have OTG_FS with CID 0x0000 1200 (same as F401), and the SVD files seem to have the same bugs as F401/F411 SVD files.


adamgreig (approved):
Thanks!

bors r+

PR #709, opened 2022-03-27
Run sudo apt-get update before installing a package

richardeoin:
As documented [here](https://docs.github.com/en/actions/using-github-hosted-runners/customizing-github-hosted-runners#installing-software-on-ubuntu-runners) and also discussed [here](https://github.com/actions/virtual-environments/issues/1757)


adamgreig (approved):
Thanks!

bors r+

PR #710, opened 2022-03-31
CRC peripheral corrections

Pagten:
This PR proposes the following corrections related to the CRC peripheral.

### 1. Remove the CRC:Polysize field from the STM32F0x0
According to [RM0360](https://www.st.com/resource/en/reference_manual/rm0360-stm32f030x4x6x8xc-and-stm32f070x6xb-advanced-armbased-32bit-mcus-stmicroelectronics.pdf), these devices do not have a programmable polygonal size.

### 2. Ensure the polynomial field within CRC:POL has the same name in all devices
In the following devices, this field was named `Polynomialcoefficients` instead of `POL` like in other devices.
 * STM32L4x2
 * STM32L4R9
 * STM32L4x1
 * STM32L4x3
 * STM32L4x5
 * STM32L4x6

### 3. Add missing CRC:POL register to STM32F0x1, STM32F0x2, STM32F0x8
According to [RM0091](https://www.st.com/resource/en/reference_manual/rm0091-stm32f0x1stm32f0x2stm32f0x8-advanced-armbased-32bit-mcus-stmicroelectronics.pdf), these devices have a CRC:POL register but it is missing from the SVD.

Note: for the STM32F03x, STM32F04x and STM32F05x the CRC:POL register is read-only. However, since no separate SVD file is in place for these device sub-ranges, the register will be incorrectly marked as read-write for them. The same applies to the CRC:CR:POLYSIZE field, which was already present in these devices before this PR.

### 4. Merge `crc_with_polysize.yaml` into `crc_pol.yaml` and update devices accordingly
As far as I can tell, the CRC:CR:POLYSIZE field (for setting the size of the CRC polynomial) and the CRC:POL register (for reading/writing the CRC polynomial itself) always coincide.

Quite a number of devices only had `crc_with_polysize.yaml` without `crc_pol.yaml`, which does not make sense to me, because why would a device support setting the size of the polynomial if it doesn't support setting the polynomial itself? For all devices that now have access to the CRC:POL register because of this change, I verified that this is in line with the applicable reference manual.

### DIFF
--- a/devices/common_patches/crc/crc_add_pol.yaml
+++ b/devices/common_patches/crc/crc_add_pol.yaml
@@ -1,0 +1,16 @@
+CRC:
+  # The SVD is missing the programmable polynomial register "POL"
+  _add:
+    POL:
+      displayName: POL
+      description: CRC polynomial
+      addressOffset: 0x14
+      size: 0x20
+      access: read-write
+      resetValue: 0x04C11DB7
+      fields:
+        POL:
+          description: Programmable polynomial
+          bitOffset: 0
+          bitWidth: 32
+          access: read-write
--- a/devices/common_patches/crc/crc_rename_pol.yaml
+++ b/devices/common_patches/crc/crc_rename_pol.yaml
@@ -1,0 +1,5 @@
+CRC:
+  POL:
+    _modify:
+      Polynomialcoefficients:
+        name: POL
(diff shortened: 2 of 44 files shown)


Pagten:
I noticed the STM32L0x0, STM32L0x1, STM32L0x2 and STM32L0x3 device files were renaming the `Polynomialcoefficients` field to `POL` manually, so I added a commit to replace that manual renaming with including the `crc_rename_pol.yaml` file, to avoid the code duplication.

adamgreig (approved):
Nice, thanks!

bors r+

adamgreig (dismissed):
Rebased to fix merge conflicts.

bors r+

adamgreig (dismissed):
bors r+

adamgreig (approved):
bors r+

PR #713, opened 2022-04-04
f4xx: UART fix missing Guard Time and Prescaler Register, f413 UART should include USART-only fields.

tim-seoss:
Currently, the F413 UART peripheral is a copy of its USART definition, whereas all of the other F4xx family use a separate independent definition for the UART peripheral (the independent definition doesn't include USART-specific fields which are not present in the UART peripheral).  Fix by using the F405 UART definition on the F413.

All F4xx UARTs support IrDA mode, which utilises the PSC fields of the GTPR register.  This field is missing in the UART definition (except F413 - see above) across the F4 family (presence and correct operation of the GTPR PSC field was verified by the author on F429 and F446 microcontrollers).

n.b. The other GTPR field (GT) appears to be specific to SmartCard mode - which is a USART-only feature, not supported on UART according to the reference manual, so the PR does not currently include that field for UARTs.

### DIFF
--- a/devices/common_patches/f4_add_UART_GTPR.yaml
+++ b/devices/common_patches/f4_add_UART_GTPR.yaml
@@ -1,0 +1,11 @@
+"UART*":
+  _add:
+    GTPR:
+      description: Guard Time and Prescaler Register
+      addressOffset: 0x18
+      access: read-write
+      fields:
+        PSC:
+          description: IrDA Low-Power pulse width peripheral clock prescaler
+          bitOffset: 0
+          bitWidth: 8
--- a/devices/stm32f405.yaml
+++ b/devices/stm32f405.yaml
@@ -111,6 +111,7 @@ _include:
  - ../peripherals/usart/uart_sample.yaml
  - ../peripherals/usart/uart_uart.yaml
  - ../peripherals/usart/uart_usart.yaml
+ - common_patches/f4_add_UART_GTPR.yaml
  - ../peripherals/i2c/i2c_v1.yaml
  - ../peripherals/wwdg/wwdg.yaml
  - ../peripherals/iwdg/iwdg.yaml
--- a/devices/stm32f407.yaml
+++ b/devices/stm32f407.yaml
@@ -120,6 +120,7 @@ _include:
  - ../peripherals/usart/uart_sample.yaml
  - ../peripherals/usart/uart_uart.yaml
  - ../peripherals/usart/uart_common.yaml
+ - common_patches/f4_add_UART_GTPR.yaml
  - ../peripherals/i2c/i2c_v1.yaml
  - ../peripherals/wwdg/wwdg.yaml
  - ../peripherals/iwdg/iwdg.yaml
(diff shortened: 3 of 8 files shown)


adamgreig (commented):
Thanks, this looks good to me! Could you add the GT field to GTPR as well? As far as I can see it's present on all the F4 devices.

tim-seoss:
> Could you add the GT field to GTPR as well?

Hi @adamgreig , thanks for the review.  As far as I could see the Guard Time bits are only used in SmartCard mode² ³, and it looks like the SmartCard mode is only present on USART peripherals¹.

The prescaler (PSC) bits are used for both IrDA (on UART and USART peripherals) and also for SmartCards (USARTS), so I only put the PSC bits in the patch.

I don't really know anything about Smartcard communication, but I suppose it's conceivable that SmartCards can actually be used with UARTs (contrary to what the datasheet says)¹.

So....  I don't know whether it's best to leave out the GT field, or put it in.  I think it's probably either not present, or maybe present but not actually usable for anything, BICBW.

Happy to go with whatever you think is best.

References:

RM0430 (STM32F412/423 reference manual, rev 8, 23-May-2018):
1. Table 157 "UART features" p.885
2. "28.4.11 Smartcard" p.915
3. "28.6.7  Guard time and prescaler register (USART_GTPR)" p.933

adamgreig:
Ah, thanks, I missed that it wasn't present on the UARTs. Some of the RMs confirm this explicitly with a note that GT is not available on the UART fields. However, we don't distinguish the UARTs and USARTs for stm32f4; they all derive from the same USART register block, so I think we should add GT here so that it's usable for USARTs and just won't do anything for UARTs, same as the other smartcard mode features.

adamgreig (approved):
Sorry for the wait. You're right, let's get this in without the GT field.

bors r+

burrbull:
now even worse.
UART4 was changed (some enums like stopbits are missed), but UART5,7,8 still depends on USART1

tim-seoss:
@burrbull does this just need repeating for uart5,7,8 or are you seeing problems with uart4 now?  Looking at the memory map changes triggered by CI: https://github.com/stm32-rs/stm32-rs-mmaps/compare/pr-713-b732293 - I can't see any signs of stopbits being omitted for uart4.  I'm building locally now to check.

burrbull:
For more detail peripheral compare (with enums) you could use https://github.com/burrbull/svd-yaml-compare

tim-seoss:
Well, this has turned into quite the rabbit hole
```
~/prog/rust/stm32-rs$ git diff --stat
 devices/stm32f413.yaml | 77 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++--
 devices/stm32f427.yaml |  4 ++++
 devices/stm32f429.yaml |  4 ++++
 devices/stm32f469.yaml |  3 +++
 4 files changed, 86 insertions(+), 2 deletions(-)

~/prog/rust/stm32f4xx-hal$ git diff --stat
 src/serial.rs | 25 ++++++++-----------------
 1 file changed, 8 insertions(+), 17 deletions(-)
```
I have to go now, but will try to get this finished up later on today.

PR #715, opened 2022-04-05
Fix various issues with STM32L562

sphw:
This PR fixes two issues with the STM32L562's SVD files. In particular the DMA registers CCR registers only included the correct fields on the first register. This PR fixes that.

It also merges the GTZC VCTR into an array, rather than individual fields.


adamgreig (approved):
Thanks! I noticed those registers already had a bad "MA" (memory address) field taking all 32 bits as well, so I removed it from CNDTR/CCR.

bors r+

PR #716, opened 2022-04-06
Add SYSCFG to stm32g070

Wassasin:
The STM32G070 does not have the VREFBUF registers, but it does have SYSCFG. The SVD seems to correspond to the documentation with regards to SYSCFG.

Hence we should simply add SYSCFG instead of deleting it.


adamgreig (approved):
Thanks, seems reasonable!

bors r+

PR #719, opened 2022-04-17
Change the groupName of ADC_Common from ADC to ADC_Common

cpunion:
In the https://github.com/tinygo-org/stm32-svd project, it depends this project and generates types by the peripheral's groupName. Because the ADC_Common's groupName is ADC, it can't generate the types ADC and ADC_Common correctly. I want to change the ADC_Common's groupName to ADC to make it work fine.  See https://github.com/tinygo-org/tinygo/pull/2717 , https://github.com/tinygo-org/stm32-svd/pull/7 , but those are not good way.

I checked the generated rust code, it's not changed after apply this change and run make, seems it doesn't depend groupName. So can you accept this PR?

### DIFF
--- a/devices/common_patches/adc_common_group_name.yaml
+++ b/devices/common_patches/adc_common_group_name.yaml
@@ -1,0 +1,5 @@
+# Rename group name of ADC_Common from ADC to ADC_Common
+
+_modify:
+  ADC_Common:
+    groupName: ADC_Common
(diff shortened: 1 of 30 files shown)


adamgreig (approved):
Sure, I think this is fine. We don't use groupName at the moment, but we might if https://github.com/rust-embedded/svd2rust/pull/617 lands.

bors r+

adamgreig:
Rebased to resolve merge conflict.

bors r+

adamgreig (approved):
bors r+

adamgreig (approved):
bors r+

PR #720, opened 2022-04-17
Add more enums for L4 clock selections.

reitermarkus:
Added enums for clock enable and selection registers.


reitermarkus:
Some `UARTxEN` don't exist for some devices, since this applies only to STM32L47xxx, STM32L48xxx, STM32L49xxx and STM32L4Axxx devices. Not sure if there is a way to organize the YAMLs to account for this. Or is it somehow possible to optionally apply a patch, only if the corresponding field exists?

peripherals/rcc/rcc_l4.yaml
```diff
@@ -13,3 +57,45 @@ RCC:
...
+    "LPTIM[123]SEL":
+      PCLK: [0, "PCLK clock selected"]
+      LSI: [1, "LSI clock selected"]
+      HSI16: [2, "HSI16 clock selected"]
+      LSE: [3, "LSE clock selected"]
+    "I2C[123]SEL":
+      PCLK: [0, "PCLK clock selected"]
+      SYSCLK: [1, "SYSCLK clock selected"]
+      HSI16: [2, "HSI16 clock selected"]
+      # Reserved: [3, "Reserved"]
```
reitermarkus, line 81: Should this be included as a variant?

richardeoin: Nope, you should define only the valid values

richardeoin:
The best way of doing this in the current system is to have a separate yaml file for those uarts that exist only on some families, and then include that from the device yaml in the `devices/` folder. For example you would have `rcc_l4.yaml` included by all parts (as currently) and `rcc_l4_uartx_y.yaml` included by only some parts

Are the parts you're interested in covered by the part definitions in the `devices/` folder? It's possible we don't have all of them yet.

reitermarkus:
> Are the parts you're interested in covered by the part definitions in the devices/ folder?

In `devices/` there is e.g. `stm32l4x6.yaml`, this would need e.g. `stm32l49x.yaml`. Does that work?

reitermarkus:
Or do I need to specifically create e.g. `stm32l496.yaml?

reitermarkus:
I had a UART/USART typo before, I think it should be working now after also splitting it up into individual files.

reitermarkus:
@richardeoin, CI is passing now.

reitermarkus:
@richardeoin, CI is passing now.

peripherals/rcc/rcc_l4.yaml
```diff
@@ -13,3 +45,45 @@ RCC:
...
```
reitermarkus, line 69:
```suggestion
```

reitermarkus, line 69:
```suggestion
```

peripherals/rcc/rcc_l4.yaml
```diff
@@ -13,3 +45,44 @@ RCC:
...
+
+  BDCR:
+    RTCEN:
+      Disabled: [0, "RTC clock disabled"]
+      Enabled: [1, "RTC clock enabled"]
+    RTCSEL:
+      NoClock: [0, "No clock"]
+      LSE: [1, "LSE oscillator clock selected"]
+      LSI: [2, "LSI oscillator clock selected"]
+      HSE32: [3, "HSE oscillator clock divided by 32 selected"]
```
richardeoin, line 88:
```suggestion
      HSE: [3, "HSE oscillator clock divided by 32 selected"]
```
I would avoid using [name][number]. That is often used to specify a frequency in MHz, which is not the case here. "HSE" is fine

peripherals/rcc/rcc_l4.yaml
```diff
@@ -13,3 +45,44 @@ RCC:
...
       Range48M: [0b1011, "range 11 around 48 MHz"]
+
+  CCIPR:
+    ADCSEL:
+      NoClock: [0, "No clock selected"]
+      PLLSAI1: [1, "PLLSAI1 clock selected"]
+      PLLSAI2: [2, "PLLSAI2 clock selected"]
+      SYSCLK: [3, "SYSCLK clock selected"]
+    CLK48SEL:
+      HSI48: [0, "HSI48 clock selected (only for STM32L49x/L4Ax devices, otherwise no clock selected)"]
```
richardeoin, line 56: The HSI48 option is also valid for RM0394 parts (STM32L41x/L42x/L43x/L44x/L45x/L46x)

peripherals/rcc/rcc_l4.yaml
```diff
@@ -13,3 +45,44 @@ RCC:
       Range24M: [0b1001, "range 9 around 24 MHz"]
       Range32M: [0b1010, "range 10 around 32 MHz"]
       Range48M: [0b1011, "range 11 around 48 MHz"]
+
+  CCIPR:
+    ADCSEL:
+      NoClock: [0, "No clock selected"]
+      PLLSAI1: [1, "PLLSAI1 clock selected"]
+      PLLSAI2: [2, "PLLSAI2 clock selected"]
```
richardeoin, line 53: PLLSAI2 option is not valid for RM0394 parts (STM32L41x/L42x/L43x/L44x/L45x/L46x)

An "only for" comment in the description is OK to explain this

richardeoin (commented):
Thanks, I made a few comments inline

reitermarkus:
Thanks @richardeoin, I have made the requested changes.

richardeoin (approved):
Great!

bors r+

PR #723, opened 2022-04-23
Fix L4 `ADC?.DR.RDATA` field name and description.

reitermarkus:


richardeoin (approved):
LGTM, thanks!

bors r+

PR #729, opened 2022-04-29
g4 RCC Remove RNGSMEN -> RNGEN

sorki:
In peripheral registers overview table this is correctly
listed as RNGSMEN, only in detailed register description
it says RNGEN which seems incorrect (as all fields in the register
use `SMEN` suffix).

Comment also suggets that this is likely an error.


adamgreig (approved):
Thanks, I agree it seems RNGSMEN is better.

bors r+

PR #695, opened 2022-02-06
Fix STM32G0 SVDs

Disasm:

PR #730, opened 2022-04-29
Update G0 vendor svds to V1.2

sorki:
In progress. Some stuff was invalidated.

Also related to 
* #716 which seems to be present here
* #695


burrbull:
Replaced by #947

PR #736, opened 2022-05-14
ci: fix github pages build

newAM:
Currently failing: https://github.com/stm32-rs/stm32-rs/actions/runs/2324401052


burrbull:
Not needed
Just add env to make
```SVDTOOLS=svd make```

newAM:
Good call, fixed.

Running a test build here: https://github.com/newAM/not-a-stm32-rs-fork/runs/6435525025
Will mark as ready when that passes.

PR #738, opened 2022-05-15
Use precompilled svd2rust/svdtools/form

burrbull:


burrbull:
cc @adamgreig

.github/workflows/ci.yaml
```diff
@@ -68,16 +67,12 @@ jobs:
         run: make crates
 
       - name: Cache dependencies
-        uses: Swatinem/rust-cache@v1
+        uses: Swatinem/rust-cache@v2
```
newAM, line 70: The `working-directory` key is producing a warning after this update: https://github.com/stm32-rs/stm32-rs/pull/738/files

Looks like it was replaced with the `workspaces` option: https://github.com/Swatinem/rust-cache/blob/master/CHANGELOG.md#200

burrbull: Do you know how to fix (or do you have examples)? I don't understand the syntax.

newAM: Yeah that is confusing, and non-standard.

I found a working example in another repo: https://github.com/alexschrod/cargo-test-annotations/blob/273b8db69fe01651453551359898c99c1ab2d87a/action.yml

Based on that I think it should probably look like this for v2?

```yaml
      - name: Cache dependencies
        uses: Swatinem/rust-cache@v1
        with:
          workspaces: "${{ matrix.crate }} -> target"
```

newAM (dismissed):
Warnings are gone, looks good to me.

adamgreig:
Nice, thank you!

If someone runs `make install` by itself (as someone who's just cloned the repo might do), I think this will just do nothing, right? Maybe we should set a default version variable for tools in the Makefile, and then both general users and the CI scripts will all use that?

Is the cache doing much for us at all, now? I think svd2rust and svdtools were the big dependencies, I guess the rest is just whatever the crates need to check - perhaps it's simpler to remove the cache entirely? Not sure how much time it's saving any more.

burrbull:
@adamgreig cache completely removed. Also ./tool_install script behavior changed.
Now
`./scripts/tool_install.sh` or `make install` installs all tools with default versions or versions from `env`.
`./scripts/tool_install.sh svd2rust` installs only svd2rust with default version ...
`./scripts/tool_install.sh svd2rust v0.25.0` installs only svd2rust with version 0.25.0

I don't know how to pass arguments to `make install` so only first case works properly.

adamgreig (approved):
thanks!

bors r+

PR #757, opened 2022-08-13
move common patches to dirs

burrbull:


adamgreig:
Could you force push so the mmaps job runs, please?

adamgreig (approved):
Thanks for the big reorganisation, looks good!

bors r+

PR #758, opened 2022-08-13
ci: fix path to tool_install.sh for mmaps actions

### LINKED ISSUE #756: mmaps on master
https://github.com/stm32-rs/stm32-rs/runs/7736728303?check_suite_focus=true

cc @adamgreig 

As mmaps on master branch work independently I can't test this CI

maybe just `make install` instead of `./scripts/tool_install.sh` fix it

newAM:
Closes #756 hopefully.

cc @burrbull

### DIFF
--- a/.github/workflows/mmaps_master.yaml
+++ b/.github/workflows/mmaps_master.yaml
@@ -32,7 +32,7 @@ jobs:
 
       - name: Install tools
         run: |
-          ./scripts/tool_install.sh svdtools
+          ./stm32-rs/scripts/tool_install.sh svdtools
 
       - name: Build and publish
         run: |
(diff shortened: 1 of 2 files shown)


newAM:
I think compare will still fail because it uses `pull_request_target:`, and is getting the `mmaps_pr.yml` file from the master branch.

adamgreig:
We can't run the script from the PR, because it would add a security vulnerability - the PR may contain malicious code that compromises the GitHub token. That's why the action uses `pull_request_target` - we need to make sure we don't execute any code from the PR, only from the existing master branch. I think it should be OK to just use `master/` instead of `pr/` here?

newAM:
Good catch!  Fixed.

adamgreig (approved):
thanks!

bors r+

PR #761, opened 2022-08-14
ci: fix tool install for nightlies

newAM:
Follow up to #756 and #758


burrbull:
cc @adamgreig

adamgreig:
Can we just run `tool_install.sh` without any arguments, then it installs all three, right?

burrbull:
> 

Yes

newAM:
Good catch, fixed & rebased.

adamgreig (approved):
bors r+

Issue #760, opened 2022-08-14
L4x6 missing registers for dfsdm filter1


jonathanherbst:
There are registers for filters 0, 2, and 3, but not 1.  Also it would be nice to match the clustering from L4x5.

jonathanherbst:
https://github.com/stm32-rs/stm32-rs/pull/764

David-OConnor:
Of note, H7 is another example of clustering.

PR #764, opened 2022-08-14
 L4x6 missing registers for dfsdm filter1 #760 

jonathanherbst:
I copied a patch from l4x5 and removed some code that the l4x5 yaml doesn't have.  I've compiled it with stm32-hal and it seems to recognize all the filters and channels like with l4x5.  However I don't really know what I'm doing so let me know what you think.


adamgreig (approved):
thank you, and apologies for the long wait for a review!

PR #766, opened 2022-08-20
Adds DCMI/PSSI peripheral and adds DCMI/PSSI support to H735

FerdinandvHagen:
This PR adds initial support for the DCMI and PSSI peripherals.
So far I know of at least the H735 and L4+ series that support these peripherals.
As a starting point the peripherals have been added to the H735 series but are also available in other SVDs. Happy to add them to those if required - but I don't have experience with those devices so didn't want to add them initially.

Remarks:
- PSSI peripheral is on the same address as DCMI for those devices - maybe we could even merge both together but I'm not sure if there are any exceptions from this rule which would mandate two files
- I haven't found a solution to add the enumerated values (especially PSSI) directly when defining registers so had to split into two. It's kinda ugly - so any hints on how to merge them is highly appreciated

Additional Devices where DCMI is available in the SVDs:
- F215
- F217
- F405
- F407
- F427
- F429
- F446
- L4P5
- L4Q5

### DIFF
--- a/peripherals/dcmi/dcmi.yaml
+++ b/peripherals/dcmi/dcmi.yaml
@@ -1,0 +1,158 @@
+"DCMI":
+  _modify:
+    _interrupts:
+      DCMI:
+        name: DCMI_PSSI
+        description: DCMI/PSSI global interrupt
+
+  CR:
+    OELS:
+      Odd: [0, "Interface captures first line after the frame start, second one being dropped"]
+      Even: [1, "Interface captures second line from the frame start, first one being dropped"]
+    LSM:
+      All: [0, "Interface captures all received lines"]
+      Half: [1, "Interface captures one line out of two"]
+    OEBS:
+      Odd: [0, "Interface captures first data (byte or double byte) from the frame/line start, second one being dropped"]
+      Even: [1, " Interface captures second data (byte or double byte) from the frame/line start, first one being dropped"]
+    BSM:
+      All: [0, "Interface captures all received data"]
+      EveryOther: [1, "Interface captures every other byte from the received data"]
+      Fourth: [2, "Interface captures one byte out of four"]
+      TwoOfFour: [3, "Interface captures two bytes out of four"]
+    ENABLE:
+      Disabled: [0, "DCMI disabled"]
+      Enabled: [1, "DCMI enabled"]
+    EDM:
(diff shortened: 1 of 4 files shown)


peripherals/pssi/pssi.yaml
```diff
@@ -0,0 +1,206 @@
...
+            description: Extended data mode
+            bitOffset: 10
+            bitWidth: 2
+          RDYPOL:
+            description: Ready (PSSI_RDY) polarity
+            bitOffset: 8
+            bitWidth: 1
+          DEPOL:
+            description: Data enable (PSSI_DE) polarity
+            bitOffset: 8
```
richardeoin, line 48:
```suggestion
            bitOffset: 6
```

peripherals/pssi/pssi.yaml
```diff
@@ -0,0 +1,206 @@
...
+            bitWidth: 1
+          CKPOL:
+            description: Parallel data clock polarity
+            bitOffset: 8
```
richardeoin, line 52:
```suggestion
            bitOffset: 5
```

peripherals/dcmi/dcmi.yaml
```diff
@@ -0,0 +1,158 @@
+"DCMI":
+  _modify:
```
richardeoin, line 2: This `_modify` block should be in a file under `devices/common_patches/`

peripherals/pssi/pssi.yaml
```diff
@@ -0,0 +1,206 @@
+_copy:
+  PSSI:
+    from: DCMI
+
+_modify:
```
richardeoin, line 5: This `_modify` block should be split into a new file, and put in `devices/common_patches/pssi/pssi.yaml` for example.

richardeoin (commented):
Thanks for the PR! I haven't reviewed fully but made a few comments inline.

There's a rough rule for managing the repository that changes to the register structure (for example bit offsets, adding new fields/registers) should go in a file under `devices/common_patches/`. So in fact the register definitions should go there, separately from the enumerated values.

FerdinandvHagen:
Thank you very much @richardeoin for the review.

I have fixed the bit offsets and moved the modify blocks to the common_patches directory.

burrbull:
DCMI/PSSI is already added for L4+.
Could you rebase (or open new PR) taking this in account?

PR #767, opened 2022-08-20
Add reduced ADC3 found on RM0468 parts

richardeoin:
The ADC3 on RM0468 is a 12-bit ADC with somewhat reduced functionality, compared to the 16-bit ADCs found elsewhere in H7. Here we copy ADC3 from ADC1 (previously it was derived) and modify it to match RM0468 Rev 3 Section 29


burrbull:
Rebase, please.

burrbull:
@richardeoin looks like after ST archives update this was partially solved. Could you update this?

richardeoin:
Thanks @burrbull . It looks like this is fully solved on master now. I propose to close it

PR #769, opened 2022-08-22
fix clusters

burrbull:
cc @richardeoin


devices/stm32f7x9.yaml
```diff
@@ -3,6 +3,10 @@ _svd: ../svd/stm32f7x9.svd
 _rebase:
   USART1: USART6
 
+_copy:
+  DFSDM:
+    from: ../svd/stm32f7x7.svd:DFSDM
```
burrbull, line 8: In f7x9::DFSDM all `addressOffset`s are incorrect. So instead of fixing them I've just copied all block from f7x7

burrbull: cc @adamgreig

adamgreig (approved):
thanks! looks good, I think this should make the arrays of clusters nicer to use.

bors r+

PR #770, opened 2022-08-24
Add `critical-section` feature.

reitermarkus:
Depends on https://github.com/rust-embedded/svd2rust/pull/651.


adamgreig:
Any reason not to add critical-section to the default features here? In fact I wonder if it shouldn't even be optional in svd2rust for cortex-m targets.

I guess the main question is: if you don't use `take()` but still have the feature enabled, and don't provide a critical-section impl, do you get a linker error (because it can't find the impl) or not (because the call is optimised out)? If there's no error in that case then I don't think there's any need to even have it as an optional feature; if there is then at least we might consider making it default since the majority of users are likely to need it and it's unlikely a dependency accidentally enables it (same deal as `rt`).

reitermarkus:
Yeah, I guess it can be a default feature here since most likely you want to use `take` rather than `steal`.

I think it should still stay optional in `svd2rust` though, mainly because it *can* be made optional. The PAC crate can then decide if it should be enabled by default.

burrbull:
0.26 is released

reitermarkus:
@burrbull, thanks, rebased.

@adamgreig, this should be good to go.

scripts/tool_install.sh
```diff
@@ -1,30 +1,20 @@
 #!/usr/bin/env bash
 
+set -euo pipefail
+
 FORM_VERSION="${FORM_VERSION:-v0.10.0}"
 SVDTOOLS_VERSION="${SVDTOOLS_VERSION:-v0.2.6}"
```
burrbull, line 6:
```suggestion
SVDTOOLS_VERSION="${SVDTOOLS_VERSION:-v0.2.7}"
```

scripts/makecrates.py
```diff
@@ -72,21 +71,18 @@
 license = "MIT/Apache-2.0"
 
 [dependencies]
-bare-metal = "1.0.0"
+critical-section = {{ version = "1.0", optional = true }}
+cortex-m = "0.7.6"
+cortex-m-rt = {{ version = ">=0.6.15,<0.8", optional = true }}
```
burrbull, line 76: @adamgreig Do we still need this workaround?

adamgreig (approved):
Rebased onto master to resolve conflicts. Thanks for the script updates as well!

bors r+

scripts/makecrates.py
adamgreig: It doesn't hurt, we are still compatible with cortex-m-rt 0.6.15 and lets people still use it if they want.

PR #772, opened 2022-08-25
Update local build instructions in README

richardeoin:
Update local build instructions to match the configuation used in CI. The current instructions are broken for master.

svd2rust 0.25.1 is used since #769, and CI uses the Rust builds of svdtools

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,10 +90,14 @@ contain the latest patches and updates.
 
 ## Generating Device Crates / Building Locally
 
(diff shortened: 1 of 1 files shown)


adamgreig:
I wonder if we should just go for `make install`, which will pull in the currently used versions of svd2rust, svdtools, and form (via scripts/tool_install.sh). Make it's a bit more opaque though?

adamgreig (approved):
bors r+

PR #773, opened 2022-08-26
Initial support for STM32L4P5 (and therefore STM32L4Q5)

FerdinandvHagen:
This pull request adds initial support for STM32L4P5 (and therefore STM32L4Q5).

Done:
- Mmaps compared to reference manual revision 9 (should fully match now)
- Enumerated values added where available / possible

Known issues:
- few peripherals missing (most notably TAMPER peripheral)
- some registers / fields currently not named correctly

Will try to fix the register / field names in the next couple of days.
As this is quite a lot of patches and also new enumerated values I would kindly ask for feedback in case things should be organized differently.

PR #774, opened 2022-08-26
Fixes whitespaces and last lines for all yaml files

FerdinandvHagen:
This pull request fixes whitespaces and new lines for all yaml files.
Reference: https://github.com/stm32-rs/stm32-rs/pull/773#issuecomment-1228413123

This was done with [pre-commit](https://pre-commit.com/) - maybe something we could also just implement for the repository and make a small CI task?


adamgreig (approved):
Thank you! I wouldn't mind a CI task that checks for trailing whitespace and missing newlines, but I'm not completely sure about adopting pre-commit and its config files and etc just for that purpose. Maybe we could start with a CI job that does the check and see how that goes?

bors r+

PR #783, opened 2022-10-20
Fixes missing CRSEN bit for L4x5 devices

FerdinandvHagen:
For some reason https://github.com/stm32-rs/stm32-rs/issues/572 was not fixed.
This request adds the missing CRSEN bit to the L4x5 devices.


adamgreig (approved):
thanks, good catch! apologies for the long wait for a review.

PR #784, opened 2022-11-14
Updated SVD for g0b1 and g0c1 to 1.5

adamt55:
As mentioned [here](https://github.com/David-OConnor/stm32-hal/issues/44), I am trying to get HAL support for the G0B1 nucleo board working. I saw that the current SVD is tagged as version 1.3 in the XML. This PR just swaps out this SVD to version 1.5 provided by STM. The crate compiles fine on my machine, if there are any obvious issues with the different SVD let me know. New to embedded, any pointers are appreciated.


newAM:
Where did this come from?  I tried to double check this by downloading the SVD from st.com <https://www.st.com/en/microcontrollers-microprocessors/stm32g0-series.html#cad-resources>, but they only have a v1.2.

adamt55:
Yes, the release on the ST page is tagged as 1.2, but the 2 SVDs I swapped out from that 1.2 release into the existing one (g0b1 and g0c1) had version 1.5 as their versionnumber in the SVD itself. I suspect 1.2 is the release of the zip-file, while 1.5 is the release of the SVD itself. I did not swap out the entire zip, since that broke a lot of stuff that I am not able to fix currently

newAM:
> Yes, the release on the ST page is tagged as 1.2, but the 2 SVDs I swapped out from that 1.2 release into the existing one (g0b1 and g0c1) had version 1.5 as their versionnumber in the SVD itself. I suspect 1.2 is the release of the zip-file, while 1.5 is the release of the SVD itself. I did not swap out the entire zip, since that broke a lot of stuff that I am not able to fix currently

Thanks for the clarification!

sha256sum's match with the ST provided v1.2 zip for the modified files; looks good to me!
```
db52cc243ff64007a3c2d408e8312daac44eff53fbf54df990e3fb1771afb891  STM32G0B1.svd
b24e2faf17350e64cd7778de205110d1de20f5d3641d3c0d15a9f68b9149824f  STM32G0C1.svd
```

newAM (approved):
bors r+

PR #785, opened 2022-11-19
G0B1 patches for new SVD

adamt55:
I tried adding new patches for the g0b1 board with the updated SVD now. 

- Added USART CR1
- Removed EXTI for now since the SVD still doesnt mention it
- Applied some timer patches
- Renamed PLLCFGR register so it fits with the rest of the G0 family

Pointers appreciated, I am new to embedded

### DIFF
--- a/devices/common_patches/dma/g0_add_dmamux_c7cr_c11cr.yaml
+++ b/devices/common_patches/dma/g0_add_dmamux_c7cr_c11cr.yaml
@@ -1,0 +1,437 @@
+DMAMUX:
+  _add:
+    C7CR:
+      displayName: C7CR
+      description: DMAMUX request line multiplexer channel 7 configuration register
+      addressOffset: 0x01C
+      size: 0x20
+      access: read-write
+      resetValue: 0x00000000
+      fields:
+        DMAREQ_ID:
+          description: DMA request identification Selects the input DMA request. See the DMAMUX table about assignments of multiplexer inputs to resources.
+          bitOffset: "0"
+          bitWidth: "6"
+          access: read-write
+        SOIE:
+          description: Synchronization overrun interrupt enable
+          bitOffset: "8"
+          bitWidth: "1"
+          access: read-write
+          enumeratedValues:
+            enumeratedValue:
+              B_0x0:
+                description: interrupt disabled
+                value: "0x0"
+              B_0x1:
+                description: interrupt enabled
+                value: "0x1"
+        EGE:
+          description: Event generation enable
+          bitOffset: "9"
+          bitWidth: "1"
+          access: read-write
+          enumeratedValues:
+            enumeratedValue:
+              B_0x0:
+                description: event generation disabled
+                value: "0x0"
+              B_0x1:
+                description: event generation enabled
+                value: "0x1"
+        SE:
+          description: Synchronization enable
+          bitOffset: "16"
+          bitWidth: "1"
+          access: read-write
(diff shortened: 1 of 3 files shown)


adamt55:
Hey, maybe somebody could help me out with an issue I haven't figured out yet: the generated code of the RegisterBlock struct of the tim2 timer looks like this now:

```
#[doc = "0x20 - capture/compare enable register"]
pub ccer: CCER,
_reserved_9_cnt: [u8; 0x04],
#[doc = "0x28 - prescaler"]
pub psc: PSC,
#[doc = "0x2c - auto-reload register"]
pub arr: ARR,

```

Instead, I need the CNT register 0x24 to look like the other struct fields as my HAL expects this. The CNT field gets created like this in devices/common_patches/tim/tim2_common_32bit.yaml:


```
CNT:
  # Actualy a 32bit register
  _modify:
    CNT_H:
      name: CNTH
      description: Counter value
    CNT_L:
      name: CNTL
  _merge:
    - "CNT*"
```

Why does this result in a different struct?

adamt55:
Hey, maybe somebody could help me out with an issue I've been having. I included the following code from common_patches/tim/tim2_common_32bit.yaml:

```
CNT:
  # Actualy a 32bit register
  _modify:
    CNT_H:
      name: CNTH
      description: Counter value
    CNT_L:
      name: CNTL
  _merge:
    - "CNT*"
```

The `CNT` field of the `tim2 RegisterBlock` struct that gets generated by this code looks like this:

```

#[doc = "0x20 - capture/compare enable register"]
pub ccer: CCER,
_reserved_9_cnt: [u8; 0x04],
#[doc = "0x28 - prescaler"]
pub psc: PSC,

```

The CNT should look like the other fields as far as I understand, and if I change the resultig code it works with my HAL. How do I fix this? Thanks!

burrbull:
if register is not in `RegisterBlock`, this means it has alternate register with same address or its address is incorrect

adamt55:
Ah I see, thanks for the hint, I disabled the alternative mode for the TIM2 counter for now. Works now, thanks!

Do you maybe have some other comments/hints regarding this PR?

devices/common_patches/dma/g0_add_dmamux_c7cr_c11cr.yaml
```diff
@@ -0,0 +1,437 @@
...
+          bitWidth: "6"
+          access: read-write
+        SOIE:
+          description: Synchronization overrun interrupt enable
+          bitOffset: "8"
+          bitWidth: "1"
+          access: read-write
+          enumeratedValues:
+            enumeratedValue:
+              B_0x0:
```
burrbull, line 23: `B_0x0` is useless name for variant.
Also we don't add enums in `common_patched`. enum related patches are in `peripherals`

Elemecca:
Many of the changes in this PR have either already been applied or are handled by the latest update to the G0 vendor SVDs in #947.

On a quick review, the only changes I see here which are still relevant are:
- Add the missing `DMAMUX:CCR*` registers (7-11). This is still a good idea, and is relevant to the entire G0 family, but the approach taken in this PR isn't great. Ideally we would use something like the `_copy` verb, but I'm not sure that works on registers (it's only documented for peripherals).
- Renaming `RCC:PLLCFGR` on G0B1/G0C1 to match the rest of the family.

burrbull:
Closing as partially outdated. Open new PR if still interested in improvement.

PR #791, opened 2022-12-27
ci: fix node 12 deprecation warnings

newAM:
`actions-rs` appears to be unmaintained: https://github.com/actions-rs/toolchain/issues/216


adamgreig (approved):
thanks! it's a shame actions-rs is no longer maintained. I've seen some people just use the cargo that comes with the GHA images these days, but it seems nice to explicitly use the latest stable.

bors r+

PR #792, opened 2022-12-27
ci: use consistent ubuntu version

newAM:
Nightlies are failing due to a libc version mismatch: https://github.com/stm32-rs/stm32-rs/actions/runs/3788639981/jobs/6441622524

The `ci` workflow is being run with `ubuntu-latest` (22.04), and the rest with the older `ubuntu-20.04`, which is a likely source for nightlies failing while the main CI workflow runs.  Not sure if this will fix it, but it is a good first start.


newAM:
On a related topic, is there a reason that only the `gh-pages` CI file uses `svd` and everything else in CI uses `svdtools`? (@burrbull  ?)  The HTML build is failing from the STM32L4+ PR with an error that only shows when using `svd`.

burrbull:
> On a related topic, is there a reason that only the `gh-pages` CI file uses `svd` and everything else in CI uses `svdtools`? (@burrbull ?) The HTML build is failing from the STM32L4+ PR with an error that only shows when using `svd`.

I don't have question. But this says that something wrong with L4+ PR. I suggest to disable `PSSI` as fast fix and then investigate the cause.
```yaml
  - common_patches/dcmi/dcmi.yaml
  - ../peripherals/dcmi/dcmi.yaml
  #- common_patches/pssi/pssi.yaml
  #- ../peripherals/pssi/pssi.yaml
```

adamgreig:
> On a related topic, is there a reason that only the `gh-pages` CI file uses `svd` and everything else in CI uses `svdtools`? (@burrbull ?) The HTML build is failing from the STM32L4+ PR with an error that only shows when using `svd`.

I think it was just missed when the rest were swapped to svdtools in #701, I don't see why we couldn't swap it. Wanna do that in this PR?

newAM:
> Wanna do that in this PR?

Sure!  Done.

adamgreig (approved):
thanks!

bors r+

PR #797, opened 2023-01-27
Feature/g4 improvements

kevswims:
Fixes a copy-paste type in the PLLSRC definitions and adds flash wait states to the ACR Latency field.


newAM (approved):
Thanks!

bors r+

PR #798, opened 2023-01-28
svdtools 0.2.8

burrbull:
cc @newAM


newAM (changes requested):
The README should be updated as well (looks like we forgot last time).

burrbull:
done

newAM (approved):
Thanks!

bors r+

PR #799, opened 2023-01-30
`zeroToClear` for `TIMx SR UIF` field

AndreySmirnov81:


peripherals/tim/tim_no_opm.yaml
```diff
@@ -17,8 +17,12 @@ DIER:
...
 SR:
   UIF:
-    Clear: [0, "No update occurred"]
-    UpdatePending: [1, "Update interrupt pending."]
+    _read:
+      NoUpdateOccured: [0, "No update occurred"]
+      UpdatePending: [1, "Update interrupt pending"]
+    _W0C:
+      Clear: [0, "Clear flag"]
+      Keep: [1, "Not modify flag"]
```
adamgreig, line 25:
```suggestion
      NoUpdateOccurred: [0, "No update occurred"]
      UpdatePending: [1, "Update interrupt pending"]
    _W0C:
      Clear: [0, "Clear flag"]
```

adamgreig (changes requested):
Thanks for the PR!

So far, everywhere else in stm32-rs when we have W0C fields we just specify a `clear` variant, not a `keep` one. For now I think it's best to be consistent with that.

adamgreig (approved):

PR #804, opened 2023-02-05
Remove all ethernet related fields and peripherals from f405/f415

datdenkikniet:
The SVD for `stm32f405` and `stm32f415` contains ethernet peripherals and related fields, while they don't actually have that.

This PR adds a patch that removes those peripherals and fields from the f405.


adamgreig (approved):
thank you for fixing this long-standing error!

PR #805, opened 2023-02-06
F1: rename GPIOx_CRL/H CNF field enum items

AndreySmirnov81:
изображение


burrbull:
Do not see advantage in this change. Should be handled on HAL level (other `enum` for `Input` with `From` implementation).

PR #808, opened 2023-02-13
Replace makehtml.py with svd2html

burrbull:
yet another oxidized tool

possibly related to #807 

cc @newAM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+* Replace `makehtml.py` with `svd2html`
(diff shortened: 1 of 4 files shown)


newAM (approved):
LGTM, HTML output is near identical, and fixes #807.

Maybe remove `scripts/makehtml.py` and `scripts/makehtml.index.template.html`?  There's `makehtml.template.html` as well, but that appears to be in use by `scripts/makejson.py`.

@adamgreig any thoughts with the tool change here?

adamgreig (approved):
looks good, thanks! maybe later we can think about moving it into svdtools.

bors r+

PR #816, opened 2023-03-26
H5 patches

David-OConnor:
This is a WIP. Currently, the changes in it, eg removing underscores on registers, is not working; open to ideas. Likely many more patches to apply after that part is fixed.

Edit: Per a discussion with Burbull, it seems the changes are taking on the GH "Memory map comparison" build, but not on my machine.

### DIFF
--- a/devices/common_patches/h5.yaml
+++ b/devices/common_patches/h5.yaml
@@ -1,78 +1,81 @@
 _clear_fields: ["*"]
 
-# RCC:
-#   _strip:
-#     - "RCC_"
+asdfasdf:
+  asdf
 
-# ADC,ADC?:
-#   _strip:
-#     - "ADC_"
+RCC:
+  _strip:
+    - "RCC_"
 
-# DBGMCU:
-#   _strip:
-#     - "DBGMCU_"
+ADC,ADC?:
+  _strip:
+    - "ADC_"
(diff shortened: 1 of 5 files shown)


devices/common_patches/h5.yaml
```diff
@@ -1 +1,78 @@
...
+  _delete:
+    - CNT
+  _modify:
+    CNT_alternate:
+      name: "CNT"
+
+TIM*:
+  _strip:
+    - "TIM*__"
+    - "TIM?*_"
```
burrbull, line 78: I don't know what you wanted to do, but don't use `*` in `_strip` pattern.
`*` means any symbol<b>s</b>

```suggestion
TIM3:
  _delete:
    - CNT
  _modify:
    CNT_alternate:
      name: "CNT"

TIM*:
  _strip:
    - "TIM?_"
    - "TIM??_"
```

David-OConnor:
Update: The build process has gotten the better of me; putting this on hold.

David-OConnor:
This is in pretty good state now. Happy to merge, then patch up later A/R. Thanks Burbull for the stm32-rs and SVD2Rust patches!

adamgreig (approved):
thanks!

bors r+

PR #823, opened 2023-03-30
cargo make support

burrbull:


burrbull:
@David-OConnor Could you test this on **native** Windows, not WSL?
You needs `cargo-make` to be installed for this.

`cargo make patch`,  `mmaps` and `clean-all` should work.

`cargo make svd2rust`, `form` and `check-patched` also should work, but require `python3` to be installed (not tested yet)

burrbull:
cc @hydra

David-OConnor:
Nice! Would love x-platform support. 

`cargo make patch` succeeds
`cargo make mmaps` successds
`cargo make clean-all` results in this error:

```rust
PS C:\Users\david\Desktop\test\stm32-rs> cargo make clean-all
[cargo-make] INFO - cargo make 0.36.6
[cargo-make] INFO - Build File: Makefile.toml
[cargo-make] INFO - Task: clean-all
[cargo-make] INFO - Profile: development
[cargo-make] INFO - Running Task: legacy-migration
[cargo-make] INFO - Running Task: clean-rs

C:\Users\david\Desktop\test\stm32-rs>cd "C:\Users\david\Desktop\test\stm32-rs"

C:\Users\david\Desktop\test\stm32-rs>rm -rf $RUST_DIRS
'rm' is not recognized as an internal or external command,
operable program or batch file.

C:\Users\david\Desktop\test\stm32-rs>rm -f */src/generic.rs
'rm' is not recognized as an internal or external command,
operable program or batch file.
[cargo-make] ERROR - Error while executing command, exit code: 1
[cargo-make] WARN - Build Failed.
PS C:\Users\david\Desktop\test\stm32-rs> cargo make clean-all
```

I don't see the generated device crates on the stm32-rs root.

Makefile.toml
```diff
@@ -0,0 +1,327 @@
...
+YAMLS = array
+PATCHED_SVDS = array
+FORMATTED_SVDS = array
+RUST_DIRS = array
+for crate in ${crates}
+    crate_yamls = glob_array devices/${crate}*.yaml
+    for yaml in ${crate_yamls}
+        serie = substring ${yaml} 8
+        serie = substring ${serie} -5
+        array_push ${NAMES} ${serie}
```
berkus, line 17: `NAMES = concat ${NAMES} " " ${serie}`

and drop the array_join below?

Makefile.toml
```diff
@@ -0,0 +1,327 @@
...
+        need_update = less_than ${check_modified} ${mod_modified}
+    else
+        need_update = set true
+    end
+    if ${need_update}
+        echo "check: ${serie}"
+        cd ${crate}
+        exec cargo check --target-dir ../target/check/ --features rt,${serie}
+        cd ..
+	    touch ${checkpath}
```
berkus, line 81: indent is invalid

burrbull:
Thanks @berkus

Makefile.toml
burrbull: Can I avoid using substring here somehow? Take `serie` directly from glob request?

burrbull:
@David-OConnor `clean-all` should be fixed.

Try also `cargo make svd2rust`, please.

burrbull:
I've added PYTHON env value. So now you could try to:
```cmd
set PYTHON=/path/to/python.exe
```
and then run:
```cmd
cargo make svd2rust
```

David-OConnor:
> I've added PYTHON env value. So now you could try to:
> 
> ```batchfile
> set PYTHON=/path/to/python.exe
> ```
> 
> and then run:
> 
> ```batchfile
> cargo make svd2rust
> ```

This worked; generated crates.

hydra:
> cc @hydra

not sure why I was `@mention`ed, do you need me to try this on Cygwin or something?  If yes, let me know what you'll like me to try.

burrbull:
Of course NOT. It is first of all for **native** Windows support without any emulators.

Makefile.toml
berkus: Don't know about that.

Makefile.toml
```diff
@@ -0,0 +1,321 @@
...
+        YAMLS = concat ${YAMLS} " " ${yaml}
+        PATCHED_SVDS = concat ${PATCHED_SVDS} " " svd/${serie}.svd.patched
+        FORMATTED_SVDS = concat ${FORMATTED_SVDS} " " svd/${serie}.svd.formatted
+        RUST_DIRS = concat ${RUST_DIRS} " " ${crate}/src/${serie}/
+    end
+    release ${crate_yamls}
+end
+NAMES = trim ${NAMES}
+set_env NAMES ${NAMES}
+CRATE_NAMES = trim ${CRATE_NAMES}
```
berkus, line 28: Why do you need trim here? Is it possible some vars will be empty in the loop above?

burrbull:
@adamgreig stm32h5 svd's were rezipped due to incorrect permission attributes

burrbull:
cc @adamgreig rebased

README.md
```diff
@@ -103,6 +103,9 @@ contain the latest patches and updates.
 * Install rustfmt: `rustup component add rustfmt`
 * Unzip bundled SVD zip files: `cd svd; ./extract.sh; cd ..`
 * Generate patched SVD files: `make patch` (you probably want `-j` for all `make` invocations)
+    * Alternatively you could install `cargo-make` runner and then use it instead of `make`. Works on MS Windows nativelly:
```
adamgreig, line 106:
```suggestion
    * Alternatively you could install `cargo-make` runner and then use it instead of `make`. Works on MS Windows natively:
```

adamgreig (approved):
Thanks for this addition. I think let's merge it and then see how it works out, I don't know if supporting both cargo-make and regular make will end up annoying if we need to make future changes to the makefile. Perhaps we should end up with just cargo-make.

For now though it is helpful.

adamgreig (approved):

PR #829, opened 2023-04-06
set variant for `MT` field

### LINKED ISSUE #828: `MT` field of `DR` and `TSDR` registers of `RTC` may need a numeric type with variant definition.
It is a corner case, single bit `MT` field of `DR` and `TSDR` registers of `RTC` is marked as `bool` type, which is a little wired.

Since `MT` stands for **Month Tens**, it may have numeric type, with variant `MT_A::Zero` and `MT_A::One`.

I think this affects all chips with RTC embedded.

eZioPan:
set variant for `MT` field of `DR` register and `TSDR` register

will close #828

### DIFF
--- a/peripherals/rtc/rtc_common.yaml
+++ b/peripherals/rtc/rtc_common.yaml
@@ -15,7 +15,9 @@ RTC:
     YT: [0, 15]
     YU: [0, 15]
     WDU: [1, 7]
-    MT: [0, 1]
+    MT:
+      Zero: [0, "Month tens is 0"]
+      One: [1, "Month tens is 1"]
     MU: [0, 15]
     DT: [0, 3]
     DU: [0, 15]
@@ -195,7 +197,9 @@ RTC:
   TSDR:
     _read:
       WDU: [0, 3]
-      MT: [0, 1]
+      MT:
+        Zero: [0, "Month tens is 0"]
+        One: [1, "Month tens is 1"]
       MU: [0, 7]
       DT: [0, 3]
       DU: [0, 15]


adamgreig:
Thanks for the PR, I agree it's a bit weird for this field to be a boolean. Were you able to test this change? From a quick check it seems like svd2rust still generates a `BitWriter` instead of a `FieldWriter`, so the API wouldn't change.

eZioPan:
sure, I will test it when I get some time.

eZioPan:
Accroding to my test, after modification, `MT` field still have `BitWriter` / `BitReader` type. I might think `svd2rust` just map all single bit field into `bool` type, and `MT` is a corner case.

My patch should provide 2 named variants, so that user won't bother to fill this field with bool value directly.

I have raise an [issue on svd2rust repo](https://github.com/rust-embedded/svd2rust/issues/719) for the "numeric thing".

burrbull:
> Accroding to my test, after modification, `MT` field still have `BitWriter` / `BitReader` type. I might think `svd2rust` just map all single bit field into `bool` type, and `MT` is a corner case.

`BitWriter` is just 1-bit width `FieldWriter`. The only reason this is not type alias, but another type is simpler documentation.   `rustdoc` works ugly methods implemented on type aliases.

The implemented methods depend on generic type of `BitWriter`:
`bool` generic provides only `set/clear_bit` methods.
`Enum` generic also provides `variant` related methods.

eZioPan:
I have heard from `svd2rust` side that currently svd spec doesn't support specify a data type mapping.

It will be good for me as long as I don't need to type `bool` value directly.

burrbull:
> It will be good for me as long as I don't need to type `bool` value directly.

Just use `w.mt().variant(MT_A::Zero)` or `w.mt().zero()` instead of `w.mt().bit(false)`

eZioPan:
> > It will be good for me as long as I don't need to type `bool` value directly.
> 
> Just use `w.mt().variant(MT_A::Zero)` or `w.mt().zero()` instead of `w.mt().bit(false)`

yeap~ so I'm just waiting my PR to be merged ;)

adamgreig (approved):
OK, let's merge this then, thanks!

PR #831, opened 2023-04-08
prefix TIMPRE variant names with Max

eZioPan:
to indicate it's a maximum multiplier, rather a fixed one.


adamgreig:
Maybe `Mul1Or2`, `Mul1Or4` would be better? It's not a _maximum_ of 2, it's "either 1 or 2, depending on another setting"?

eZioPan:
I have read reference manual again, and what you suggest are more precise, thanks!

adamgreig (approved):
thanks!

PR #834, opened 2023-04-24
Apply TIM2 32-bit patch to STM32WB55

iostat:
Seems like TIM2 could've benefited from the `tim2_32bit` patch, looks like someone started it but never quite finished (or maybe the `tim2_32bit` patch didn't exist back then?)

Either way, this PR does exactly what the name suggests


adamgreig (approved):
Thanks for the PR!

PR #320, opened 2019-12-15
field arrays

burrbull:
See https://github.com/rust-embedded/svd2rust/pull/400

PR #844, opened 2023-05-30
DMA ISR fixes for G0, G4

burrbull:
Part of #320

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Replace `makehtml.py` with `svd2html`
 * Updated to svd2rust 0.28.0, svdtools 0.3.0, use tools binaries for CI
 * files in devices/common_patches moved to subdirectories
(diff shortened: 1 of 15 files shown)


burrbull:
cc @adamgreig

adamgreig:
Thanks! G4 changes look good. Couple questions on G0:

* Just to check, you deleted `access: read-only` from fields because the whole register is already read-only, right?
* Why is IFCR renamed to "high status register" instead of keeping "DMA interrupt flag clear register"?

burrbull:
> Thanks! G4 changes look good. Couple questions on G0:
> 
>     * Just to check, you deleted `access: read-only` from fields because the whole register is already read-only, right?

Yes. Easier to compare with other SVDs.

> 
>     * Why is IFCR renamed to "high status register" instead of keeping "DMA interrupt flag clear register"?

Reasonable. Fixed.

PR #846, opened 2023-06-04
atomics

burrbull:


adamgreig (approved):
thanks!

burrbull:
Look like it was mistake.
I've found 2 issues:
1. It is broken when feature is generated, but not enabled due to absent `cfg(feature)`. svd2rust needs fix https://github.com/rust-embedded/svd2rust/commit/7eb1098a8cc84a9adc4cef2ea55ab43c720afa94
2. Does not work on arm6 targets.

PR #848, opened 2023-06-08
rename `gint` field to `gintmsk`

eZioPan:
to match up with Reference Manual


eZioPan:
need more adjust, close for now

PR #849, opened 2023-06-09
doc on `OTG_FS_GLOBAL` for STMF4

eZioPan:
for every `F4` device with `USB_OTG`:  
rename `GINT` field of `GAHBCFG` register of `OTG_FS_GLOBAL` .

doc on `FS` only device:  
include 401, 411, 405, 407, 417, 427, 437, 429, 439  
not include 412, 413, 423, 446, 469, 479


eZioPan:
I may pause until #851 gets some progress

PR #850, opened 2023-06-10
check `CARGO_HOME` env var

eZioPan:
check `CARGO_HOME` env var before put binary into default place


eZioPan:
Are my changes conflict with CI workflow? The script runs fine on my linux machine.

adamgreig:
The CI failure is in the alternative cargo-make based `Makefile.toml` rather than the regular `Makefile`. I guess it's somehow ending up installing the tools to a directory that's not on PATH or something like that... if you run `cargo make` locally does it work out?

eZioPan:
`cargo make` run smoothly on my local linux machine. The only en vars I add is `RUSTUP_HOME` and `CARGO_HOME`.

eZioPan:
it's weird, in CI, PATH env inlude `${CARGO_HOME}/bin/` https://github.com/stm32-rs/stm32-rs/actions/runs/5239385480/jobs/9459159924

eZioPan:
@adamgreig sorry for take that long time to fix the bug. I would ask a review on this again.

PR #856, opened 2023-06-14
doc on `STK` core peripheral for STM32F4

eZioPan:
`STK` core peripherals is for Cortex-M4 devices, but only included into STM32F4 devices


burrbull:
core peripherals are removed in new SVD sources.
Use `cortex-m` instead.

PR #872, opened 2023-08-23
nvic register arrays

burrbull:


burrbull:
close in favor of #878

PR #875, opened 2023-09-02
Add `QUADSPI` peripheral

eZioPan:


burrbull:
There are also other chips that have quadspi:
изображение

The differences are minimal.

eZioPan:
@burrbull thanks for point it out!
@newAM would you consider remove `stm32f4` tag? Or add `g4` `wb` `h7` `f7` `l4`?

devices/common_patches/quadspi/dr_multi_access.yaml
```diff
@@ -0,0 +1,17 @@
...
+        bitOffset: 0
+        bitWidth: 8
+      DataHalfWord:
+        description: Half word (16 bytes) access to DATA field
+        bitOffset: 0
+        bitWidth: 16
+      DataFullWord:
+        description: Full word (32 bytes) access to DATA field
+        bitOffset: 0
+        bitWidth: 32
```
burrbull, line 17: I suppose you are trying to do volatile access to lower bytes of register accordind to:
> Word, halfword, and byte accesses to this register are supported.

In this case you need to add additional registers, not the fields. Also don't rename registers/fields if it is not really needed.
```suggestion
  _modify:
    DR:
      description: "Data register: full word (32 bytes) access" 
  _add:
    DR16:
      addressOffset: 0x20
      alternateRegister: DR
      size: 0x10
      description: "Data register: half word (16 bit) access" 
      fields:
        DATA:
          description: Data
          bitOffset: 0
          bitWidth: 16
    DR8:
      addressOffset: 0x20
      alternateRegister: DR
      size: 0x08
      description: "Data register: one byte (8 bit) access" 
      fields:
        DATA:
          description: Data
          bitOffset: 0
          bitWidth: 8
```

and in `base.yaml`:
```yaml
  DR:
    DATA: [0, 0xFFFF_FFFF]

  DR16:
    DATA: [0, 0xFFFF]

  DR8:
    DATA: [0, 0xFF]
```

eZioPan: thanks for helping!

burrbull (approved):

PR #880, opened 2023-09-14
Add dual bank related fields to stm32g4 cat 3 devices

usbalbin:
This adds the missing dual bank flash related fields to stm32g4 devices:
* stm32g471
* stm32g473
* stm32g474
* stm32g483
* stm32g484

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@
 * Update RNG for stm32h735
 * H5: Add CRS, WWDG, IWDG, I2C, SBS, PWR, GPIO, EXTI, GPDMA, SPI, UART, RCC, TIM1-8 definitios
 * Add GPIOx:HSLVR
+* Add dual bank flash related fields to g4 cat 3 devices
 * Merge USART BRR fields on G4
 * Add G4 DAC peripheral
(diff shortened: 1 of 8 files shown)


usbalbin:
Correct me if I am wrong, but from what I can tell [FLASH_OPTR](https://www.st.com/resource/en/reference_manual/rm0440-stm32g4-series-advanced-armbased-32bit-mcus-stmicroelectronics.pdf#page=137) gets its value from [here](https://www.st.com/resource/en/reference_manual/rm0440-stm32g4-series-advanced-armbased-32bit-mcus-stmicroelectronics.pdf#page=110) automatically at boot. Because of that, FLASH_OPTR does not have a reset value per say. However the place that it gets its value from does, 0xFFEFF8AA.

Would it then make sense to define FLASH_OPTR's resetValue as 0xFFEFF8AA?(see [c7fe7d7](https://github.com/stm32-rs/stm32-rs/pull/880/commits/c7fe7d73f7ceffc40418871f5003653741ff7aa2))

usbalbin:
@adamgreig or anyone else, might I request a review and a new CI run (I do not think/hope the error is because of me) :)

Also, what/are there any plans for a new release?

burrbull (approved):

PR #881, opened 2023-09-20
update to svdtools 0.3.4

burrbull:


burrbull:
After https://github.com/rust-embedded/svdtools/pull/145 `xmllint` fails.

adamgreig:
Ugh, yes, what a pain. Have you seen any SVDs from manufacturers that use `.` in the name (and so fail the schema)? I wonder what they all expect to do, given as the spec is clear to use `.` (and it seems like svdconv uses them too) but the schema is clear they're not allowed (perhaps incorrectly).

My guess is we patch the schema to allow `.` for this identifier since it matches the spec, if we can't find any other way out.

burrbull:
As I understand they just avoid using `enumeratedValues::derivedFrom` and derive full `field` instead as possible.
Possibly we should do the same, but it is big breaking change for our infrastructure.

burrbull:
> My guess is we patch the schema to allow `.` for this identifier since it matches the spec, if we can't find any other way out.

I think, yes. We should allow `.` in schema.

After that we should replace enum derives with field arrays and derives where it is possible (most of cases I think). After https://github.com/rust-embedded/svd2rust/pull/746 `svd2rust` processes `<field derivedFrom>` as expected (at least for cases I've tested).
Note that you need to use full path for crossregister field derives.

adamgreig (approved):

PR #897, opened 2023-10-11
Add definitions for TIM1-8 for H5 family

astapleton:
This does some light refactoring of the timer field definitions to account for similarities and differences between the H5 family and other families, and provides full definitions for Timers 1-8, which are common to the H5 family. Note: It does not provide any definitions for TIM1x found on the larger H562, H563 and H573 chips.

### DIFF
--- a/devices/common_patches/tim/tim_dither.yaml
+++ b/devices/common_patches/tim/tim_dither.yaml
@@ -1,0 +1,62 @@
+# Applicable to the H5 family
+# The timers support dithering, and the ARR and CCRx registers store the
+# dithering part of the value in the least significant 4 bytes. This adds
+# overlapping field definitions to address the integer and dithering parts
+# separately
+
+"TIM[13489],TIM??":
+  "CCR?s":
+    _add:
+      INTEGER:
+        description: Integer part in dithering mode
+        bitOffset: 4
+        bitWidth: 16
+        access: read-write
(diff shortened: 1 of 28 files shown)


devices/common_patches/tim/tim_dither.yaml
```diff
@@ -0,0 +1,62 @@
...
+      INTEGER:
+        description: Integer part in dithering mode
+        bitOffset: 4
+        bitWidth: 16
+        access: read-write
+      DITHER:
+        description: Dithered part in dithering mode
+        bitOffset: 0
+        bitWidth: 4
+        access: read-write
```
burrbull, line 19: @astapleton Overlapping fields are not allowed by SVD. You should use alternate register to specify another interpretation of register.

Also this split is not very useful. It does not work in normal mode (non-dithering) as lower [0:16] bits are used in this case.
And even in dithering mode  I do not see how it can help as you just need to do (`val << 4`). Maybe you can fix me.

astapleton: You're right, this file is a bit misguided, incorrect and not particularly useful. It looks like it has been refactored on master into 2 files, and neither is being used in the H5XX device files at the moment (their inclusion is commented out). I'm happy for them to just be removed. Still want me to do that?

PR #909, opened 2023-10-17
don't add novbussens field where it is not needed

burrbull:


adamgreig:
should `nonbussens` in the file name be `novbussens`?

adamgreig (approved):

PR #915, opened 2023-10-27
Add FDCAN2 peripheral to G491

### LINKED ISSUE #913: STM32G491 missing FDCAN2 peripheral
It appears to be missing in the svd from ST (i.e. it is present in the G484 svd, but not the G491)

Would submit a PR myself but I can't find any examples on how one would add entire peripheral with a patch rather than the individual registers.

liamkinne:
Resolves #913

### DIFF
--- a/devices/stm32g491.yaml
+++ b/devices/stm32g491.yaml
@@ -1,5 +1,20 @@
 _svd: ../svd/stm32g491.svd
 
+_derive:
+  FDCAN2:
+    _from: FDCAN
+    baseAddress: 0x40006800
+
+FDCAN2:
+  add:
+    _interrupts:
+      FDCAN2_intr0:
+        description: FDCAN2_intr0
+        value: 86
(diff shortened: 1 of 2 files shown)


adamgreig (approved):
thank you!

burrbull:
Python CI fails after this patch. Don't know why:
https://github.com/rust-embedded/svdtools/actions/runs/6796652591/job/18477192193?pr=178

devices/stm32g491.yaml
```diff
@@ -1,5 +1,20 @@
 _svd: ../svd/stm32g491.svd
 
+_derive:
+  FDCAN2:
+    _from: FDCAN
+    baseAddress: 0x40006800
+
+FDCAN2:
+  add:
```
burrbull, line 9: _add

liamkinne: Fixing with #917

PR #917, opened 2023-11-08
Fix key in patch

liamkinne:
A bug introduced in my last pr.


burrbull:
Could you rebase on master, please? Memory map looks strange.

adamgreig (approved):
thanks!

PR #918, opened 2023-11-11
update stm32g0 SVD ZIP to v1.4

RavuAlHemio:
This should fix, among other things, the missing peripherals on g0b0.

(This PR is a draft because I expect some conflicts with the patches in this repository.)


burrbull:
Please, don't use _replace_enum here.
Better remove all enums with
```
_clear_fields: ["*"]
```

RavuAlHemio:
> Please, don't use _replace_enum here.

Sure thing; may I ask you to be more specific? Are you referring to the patches under `peripherals/`, those directly under `devices/`, those under `devices/common_patches/`, or all three?

burrbull:
Replaced by #947

PR #919, opened 2023-11-12
mmaps_pr: also take SVDs from PR

RavuAlHemio:
The YAML files in the pull request might be based off of updated SVD files; ensure we are using the most up-to-date SVD files to generate the pull request's memory map files.

### DIFF
--- a/.github/workflows/mmaps_pr.yaml
+++ b/.github/workflows/mmaps_pr.yaml
@@ -55,10 +55,10 @@ jobs:
           BRANCH=pr-${{ github.event.number }}-$COMMIT
           echo "BRANCH=$BRANCH" >> $GITHUB_ENV
 
-          # Use the PR's YAML files to rebuild mmaps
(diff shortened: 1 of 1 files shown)


adamgreig:
Thanks for opening this PR! This is what we used to do but we had to change the behaviour because there's a subtle security problem: if you copy the `svd/` folder from the PR, you include the `extract.sh` script (which unzips and also duplicates SVDs as required), which is then executed by the Makefile. The CI runs with repo access permissions because it has to push commits and make comments. If you allow it to execute arbitrary PR code, someone could open a PR that immediately sends them all the repo access tokens.

So, overall we have to be really careful in any CI that uses `pull_request_target` to never execute anything supplied by the PR. In principle we could allow it to only copy zip files, but since most PRs that change the zip files also end up changing the `extract.sh` script, I'm not sure it's worth the risk; PRs updating the zip files are rare enough that I think it's better to just check those manually.

PR #925, opened 2023-11-19
Update stm32h735 RNG.

twantonie:
A register and several bits in the control register were missing from the SVD file. These missing registers were very similar to those in the l4+, so I edited those patches to make it work.


richardeoin (dismissed):
Checked against RM0486 Rev 3.

The enumeration coverage is now wider (includes `../peripherals/rng/rng_wl.yaml`), these also match the RM.

richardeoin:
@twantonie Would you be able to rebase this once more? It's conflicting with the base branch annoyingly

Then I will merge

richardeoin (approved):

PR #935, opened 2023-12-11
can F1

burrbull:


adamgreig (approved):

burrbull:
Not sure. Delete them instead?

burrbull:
изображение

adamgreig:
Right, CAN only exists on connectivity line (F105, F107), and F103, which means it's not ever present on F101/F102.

burrbull:
> Right, CAN only exists on connectivity line (F105, F107), and F103, which means it's not ever present on F101/F102.

Done.

adamgreig (approved):

PR #949, opened 2024-01-11
Build mmaps for comparison on PR branch.

Elemecca:
The GitHub Action which generates the mmap comparison link for pull requests currently applies the YAML patches from the PR branch using the SVDs and tooling from the master branch in order to avoid running the untrusted tooling code from the PR branch in a context where it has access to secrets. For more context see #919.

This replaces that workflow with a pair that implement the pattern described in [GitHub's blog post on how to combine an untrusted build with privileged actions](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/).

Unfortunately (as is typical of CI changes) I haven't been able to test new new workflows as that would require setting up a bunch of temporary repositories.

Since this switches from the `pull_request_target` event to `pull_request`, workflows may no longer run automatically for PRs submitted by first-time contributors. You should check [the setting for when to run workflows for PRs from forks](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#controlling-changes-from-forks-to-workflows-in-public-repositories).

The switch to the `pull_request` even also means that the job to generate the mmaps will run on a commit generated by GitHub that merges the PR head branch into master, and that it won't run at all if there are merge conflicts. I think that's probably a good thing, but if you want I think I can change it to run on the PR head commit.

I would also recommend [changing the default `GITHUB_TOKEN` permissions](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#configuring-the-default-github_token-permissions) to "Restricted". That will ensure that Actions jobs don't have write access to anything unless they request it specifically with the `permissons` setting in their YAML file. I've looked over the other workflows and I think all of the ones that need extra permissions already request them.


burrbull:
update changelog, please. Add newline in the end of mmaps_pr_generate.yaml

burrbull (approved):

burrbull:
@Elemecca @adamgreig  "mmaps_generate" works fine but

burrbull:
I've commented `name: mmaps` and now it just silently says "0 artifacts downloaded".

PR #958, opened 2024-02-18
Patch to fix SVD 1.6 of STM32H747 and STM32H757.

seisyuu-hantatsushi:
Because DMA_SxCR of H747/H757/H745/H755 has TRBUFF bit.


burrbull:
If those fields are identical it makes sense to add field array instead.

seisyuu-hantatsushi:
> If those fields are identical it makes sense to add field array instead.

Thank you for your advice.
I will try to rewrite fields.

burrbull (approved):

seisyuu-hantatsushi:
Thanks for merged my commit.

PR #974, opened 2024-04-05
STM32U5 improvements

RootCubed:
Improves support for the STM32U5 device family.

Changes made:
- Strip peripheral prefixes from register names
- ADC, DMA2D, EXTI, FMC, GPIO, I2C, OCTOSPI, PWR, RCC and UCPD peripheral support


RootCubed:
I think this is more or less ready for review now. Of course there are still many peripherals missing, but I think this a good amount for now.

A few points:
- The UCPD peripheral exists in some other STM32 devices, but I thought it would probably be better to keep the scope of this PR to just the STM32U5 family. Let me know if you'd prefer me adding them to those other devices as well.
- I made separate commits for each peripheral (apart from those where I just had to add the existing definitions). Would you prefer me squashing it all into one commit?
- For fields which can take any arbitrary value (e.g. OSR in ADC_CFGR2), I always explicitly wrote out the range in the yaml - is that what I should have done or is it unnecessary to do that?

burrbull:
Please, split it on several PRs. It is too big.

RootCubed:
OK, will do. If that's alright with you, I'll put the smaller changes (DMA2D + GPIO + I2C + FMC + EXTI) all in one PR (~200 additions) and the remaining ones (RCC, UCPD, OCTOSPI, PWR and ADC) in separate ones. Or would you prefer one PR for every peripheral?

burrbull:
Let's try. If it is not enough I'll say.

RootCubed:
First part is in #982, will wait with opening PRs for the remaining peripherals.

PR #978, opened 2024-05-01
Fix Typo that prevented Debug / defmt implementations

agrif:
When svd2rust.toml was added, there was a typo in the options to generate Debug and defmt implementations. As far as I can tell, this hasn't ever worked. The generated crates [optionally depend on defmt](https://github.com/stm32-rs/stm32-rs/blob/b8b881bb5145b0751fb3cb4896cdf69274a7cdc1/scripts/makecrates.py#L84) but none of the code ever uses it or the `defmt` feature flag.

This PR corrects the typos, so that the generated crates have extra Debug impls and actually use the `defmt` feature flag *at all*. [`--impl-debug` and `--impl-defmt` are the correct flags as documented by svd2rust.](https://docs.rs/svd2rust/latest/svd2rust/#the---impl-debug-flag)

If this is not the desired behavior, consider removing these lines and also the defmt dependency in the generated crates.


adamgreig (approved):
thanks, well spotted!

PR #983, opened 2024-07-04
STM32U5: Strip peripheral names from registers

RootCubed:


devices/common_patches/u5.yaml
```diff
@@ -0,0 +1,80 @@
...
+FDCAN?:
+  _strip:
+    - FDCAN?_
+
+TIM?:
+  _strip:
+    - TIM?_
+
+_include:
+  - usart/rename_CR1_ISR_enabled_disabled.yaml
```
burrbull, line 80: ?

RootCubed: My bad, didn't realize I still had that in there... and looking at that patch more closely, it doesn't quite work unchanged (RXFFIE and TXFEIE would end up missing from the patched svd). Will remove and deal with later.

burrbull (approved):

PR #984, opened 2024-07-04
STM32U5: Add RCC peripheral

RootCubed:
Originally part of #974.

This is probably the largest of the bunch, but mostly just because I included all the `DivXXX` enum values most of the time. I'm not sure if there's a general rule when to do it and when not (e.g. PLLxN I didn't do because it went up to 512), so let me know if there's any you would prefer also not having all written out like that.

### DIFF
--- a/peripherals/rcc/rcc_u5.yaml
+++ b/peripherals/rcc/rcc_u5.yaml
@@ -1,0 +1,716 @@
+# RCC peripheral
+# Applicable to STM32U5
+
+RCC:
+  CR:
+    PLL3RDY:
+      _read:
+        Unlocked: [0, "PLL3 unlocked"]
+        Locked: [1, "PLL3 locked"]
+    PLL3ON:
+      Disabled: [0, "PLL3 OFF"]
+      Enabled: [1, "PLL3 ON"]
+    PLL2RDY:
+      _read:
+        Unlocked: [0, "PLL2 unlocked"]
+        Locked: [1, "PLL2 locked"]
+    PLL2ON:
+      Disabled: [0, "PLL2 OFF"]
+      Enabled: [1, "PLL2 ON"]
+    PLL1RDY:
+      _read:
+        Unlocked: [0, "PLL1 unlocked"]
+        Locked: [1, "PLL1 locked"]
+    PLL1ON:
+      Disabled: [0, "PLL1 OFF"]
+      Enabled: [1, "PLL1 ON"]
+    HSEEXT:
+      Analog: [0, "external HSE clock analog mode"]
+      Digital: [1, "external HSE clock digital mode (through I/O Schmitt trigger)"]
+    CSSON:
+      Disabled: [0, "Clock security system OFF (clock detector OFF)"]
+      Enabled: [1, "Clock security system ON (Clock detector ON if the HSE oscillator is stable, OFF if not)"]
+    HSEBYP:
+      NotBypassed: [0, "HSE crystal oscillator not bypassed"]
+      Bypassed: [1, "HSE crystal oscillator bypassed with external clock"]
+    HSERDY:
+      _read:
+        NotReady: [0, "HSE oscillator not ready"]
+        Ready: [1, "HSE oscillator ready"]
+    HSEON:
+      Disabled: [0, "HSE oscillator off"]
+      Enabled: [1, "HSE oscillator on"]
+    SHSIRDY:
+      _read:
+        NotReady: [0, "SHSI oscillator not ready"]
+        Ready: [1, "SHSI oscillator ready"]
+    SHSION:
+      Disabled: [0, "SHSI oscillator off"]
+      Enabled: [1, "SHSI oscillator on"]
+    HSI48RDY:
+      _read:
+        NotReady: [0, "HSI48 oscillator not ready"]
+        Ready: [1, "HSI48 oscillator ready"]
+    HSI48ON:
+      Disabled: [0, "HSI48 oscillator off"]
+      Enabled: [1, "HSI48 oscillator on"]
+    HSIRDY:
+      _read:
+        NotReady: [0, "HSI16 oscillator not ready"]
+        Ready: [1, "HSI16 oscillator ready"]
+    HSIKERON:
+      NotForced: [0, "No effect on HSI16 oscillator"]
+      Forced: [1, "HSI16 oscillator forced on even in Stop mode"]
+    HSION:
+      Disabled: [0, "HSI16 oscillator off"]
+      Enabled: [1, "HSI16 oscillator on"]
+    MSIPLLFAST:
+      Normal: [0, "MSI PLL normal start-up"]
+      Fast: [1, "MSI PLL fast start-up"]
+    MSIPLLSEL:
+      MSIK: [0, "PLL mode applied to MSIK (MSI kernel) clock output"]
+      MSIS: [1, "PLL mode applied to MSIS (MSI system) clock output"]
+    MSIKRDY:
+      _read:
+        NotReady: [0, "MSIK (MSI kernel) oscillator not ready"]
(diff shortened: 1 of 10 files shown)


peripherals/rcc/rcc_u5.yaml
```diff
@@ -0,0 +1,721 @@
...
+      Enabled: [0, "AHB2_2 clock enabled, distributed to peripherals according to their dedicated clock enable control bits"]
+      Disabled: [1, "AHB2_2 clock disabled"]
+    AHB2DIS1:
+      Enabled: [0, "AHB2_1 clock enabled, distributed to peripherals according to their dedicated clock enable control bits"]
+      Disabled: [1, "AHB2_1 clock disabled"]
+    AHB1DIS:
+      Enabled: [0, "AHB1 clock enabled, distributed to peripherals according to their dedicated clock enable control bits"]
+      Disabled: [1, "AHB1 clock disabled"]
+    DPRE:
+      Div1: [0, "DCLK not divided"] # same for [0, 3]
```
burrbull, line 177:
```suggestion
      Div1: [-1, "DCLK not divided"] # same for [0, 3]
```
This makes `Div1` default value for reading.

peripherals/rcc/rcc_u5.yaml
```diff
@@ -0,0 +1,721 @@
...
+    SPI2SEL,I2C[124]SEL,UART[45]SEL,USART[23]SEL:
+      PCLK1: [0, "PCLK1 selected"]
+      SYSCLK: [1, "SYSCLK selected"]
+      HSI16: [2, "HSI16 selected"]
+      MSIK: [3, "MSIK selected"]
+    SPI1SEL,USART1SEL:
+      PCLK2: [0, "PCLK2 selected"]
+      SYSCLK: [1, "SYSCLK selected"]
+      HSI16: [2, "HSI16 selected"]
+      MSIK: [3, "MSIK selected"]
```
burrbull, line 525:
```suggestion
    SPI?SEL,I2C[124]SEL,UART[45]SEL,USART?SEL:
      PCLK: [0, "PCLKx selected"]
      SYSCLK: [1, "SYSCLK selected"]
      HSI16: [2, "HSI16 selected"]
      MSIK: [3, "MSIK selected"]
```
It is just recommendation, but I'd collect them to simplify writing a HAL.

peripherals/rcc/rcc_u5.yaml
```diff
@@ -0,0 +1,721 @@
...
+    OTGHSSEL:
+        HSE: [0, "HSE selected"]
+        PLL1P: [1, "PLL1 \"Q\" (pll1_q_ck) selected"]
+        HSE2: [2, "HSE/2 selected"]
+        PLL1P_Div2: [3, "PLL1 \"P\" divided by 2 (pll1_p_ck/2) selected"]
+    I2C[56]SEL,USART6SEL:
+      PCLK1: [0, "PCLK1 selected"]
+      SYSCLK: [1, "SYSCLK selected"]
+      HSI16: [2, "HSI16 selected"]
+      MSIK: [3, "MSIK selected"]
```
burrbull, line 536:
```suggestion
    I2C[56]SEL,USART6SEL:
      _derivedFrom: RCC.CCIPR1.SPI1SEL.SPI1SEL
```

RootCubed: I presume you meant `derivedFrom` without the underscore? But this syntax doesn't seem to work for me. I added `_name: ClkSelSpiUart` to the other one, but I am getting an error
```
    Caused by:
        0: Processing device `STM32U5A9`
        1: According to `RCC`
        2: Processing peripheral `RCC`
        3: According to `CCIPR2`
        4: Processing register `CCIPR2`
        5: Processing field matched to `I2C[56]SEL,USART6SEL`
        6: Adding read-write enumeratedValues
        7: Value is not a vector (array): String("RCC.CCIPR1.USART1SEL.ClkSelSpiUart")
```

burrbull: No I meant with underscore: https://github.com/rust-embedded/svdtools/blob/1426fdecb1af98020c62a3e58ec0964f59bd3c6e/src/patch/register.rs#L735 . But there no implementation for crossregister enum derive.
But you could derive whole field. It is even better:
```yaml
  CCIPR2:
    _modify:
      I2C?SEL,USART?SEL:
        derivedFrom: RCC.CCIPR1.USART1SEL
```

RootCubed: This does not throw any errors, but the patched .svd is missing the enumeratedValues for `I2C5SEL`, `I2C6SEL` and `USART6SEL`. I will push the other suggestions for now, let me know if you can get it working.

RootCubed:
I'll squash the commits once the changes are approved.

burrbull (dismissed):

peripherals/rcc/rcc_u5.yaml
burrbull: Works as expected for me:
```xml
            <field derivedFrom="RCC.CCIPR1.USART1SEL">
              <name>USART6SEL</name>
              <description>USART6 kernel clock source selection
These bits are used to select the USART6 kernel clock source.
The USART6 is functional in Stop 0 and Stop 1 modes only when the kernel clock is HSI16 or LSE.
Note: This bitfield is only available on some devices in the STM32U5 Series. Refer to the device datasheet for availability of its associated peripheral. If not present, consider this bitfield as reserved and keep it at reset value.</description>
              <bitOffset>16</bitOffset>
              <bitWidth>2</bitWidth>
              <access>read-write</access>
            </field>
```
It should not content new `enumeratedValues` section as it derives existent one.

RootCubed: Ah, I see now. I have added that change 👍

RootCubed:
Hmm, now the CI linting step fails because of the `derivedFrom`
```
svd/stm32u5a9.svd.patched:185387: element field: Schemas validity error : Element 'field', attribute 'derivedFrom': [facet 'pattern'] The value 'RCC.CCIPR1.USART1SEL' is not accepted by the pattern '((%s)|(%s)[_A-Za-z]{1}[_A-Za-z0-9]*)|([_A-Za-z]{1}[_A-Za-z0-9]*(\[%s\])?)|([_A-Za-z]{1}[_A-Za-z0-9]*(%s)?[_A-Za-z0-9]*)'.
```

burrbull:
> Hmm, now the CI linting step fails because of the `derivedFrom`
> 
> ```
> svd/stm32u5a9.svd.patched:185387: element field: Schemas validity error : Element 'field', attribute 'derivedFrom': [facet 'pattern'] The value 'RCC.CCIPR1.USART1SEL' is not accepted by the pattern '((%s)|(%s)[_A-Za-z]{1}[_A-Za-z0-9]*)|([_A-Za-z]{1}[_A-Za-z0-9]*(\[%s\])?)|([_A-Za-z]{1}[_A-Za-z0-9]*(%s)?[_A-Za-z0-9]*)'.
> ```

Revert. Let's merge working part first. We need discuss with @adamgreig what should we do with this check, as it blocks correct SVD.

burrbull (approved):

RootCubed:
@burrbull It seems the `[-1, ...]` syntax for enums broke the [gh-pages action](https://github.com/stm32-rs/stm32-rs/actions/runs/9799258180/job/27059165538). Is this a bug in svdtools?

Added a check in svdtools `html_cli.rs:264` to debug and got
```
Caused by:
    0: In peripheral RCC
    1: In register CFGR2
    2: Value is missing from EnumeratedValue { name: "Div1", description: Some("HCLK not divided"), value: None, is_default: Some(true) }
```

burrbull:
> Is this a bug in svdtools?

Some sort of.
Should be fixed by https://github.com/rust-embedded/svdtools/pull/234
Open svdtools dependecy bump PR please when it is merged.

PR #985, opened 2024-07-04
U5: Add OCTOSPI peripheral

RootCubed:
Originally part of #974.

Not many actual changes, but I extracted the common parts of `peripherals/octospi/l4.yaml` to `peripherals/octospi/common.yaml` so it shows quite a bit of noise.


burrbull:
Since you move most of code I propose to use slightly different way to include:
https://github.com/stm32-rs/stm32-rs/commit/f4c36240e320ae2dd2234a9a82227d7f8d4171e8

burrbull (approved):

PR #994, opened 2024-07-18
Add derivePathType to CMSIS-SVD schema

burrbull:


adamgreig:
I'm a little nervous to be even more different from the official CMSIS-SVD schema, is this necessary to validate some genuine SVD files?

burrbull:
> I'm a little nervous to be even more different from the official CMSIS-SVD schema, is this necessary to validate some genuine SVD files?

This blocks us from using full qualified `derivedFrom` pathes. Like `PER.REG.FIELD`, etc. It is valid SVD, but not supported by schema. I don't know why. Maybe it was not updated a long time because `svdconv` is recommended way for check now.

PR #996, opened 2024-08-01
Fix G4 interrupt numbers

### LINKED ISSUE #456: STM32G4 FDCAN Interrupts are swapped
FDCAN1_INTR0_IT and FDCAN1_INTR1_IT are swapped (seems to be an error from the ST), what would be the best way to fix this?

liamkinne:
Fixes #456

### DIFF
--- a/devices/common_patches/fdcan/fdcan_g4.yaml
+++ b/devices/common_patches/fdcan/fdcan_g4.yaml
@@ -305,7 +305,7 @@ _include:
 EXTI:
   _modify:
     _interrupts:
-      FDCAN1_INTR0_IT:
+      fdcan1_intr0_it:
         value: 21
-      FDCAN1_INTR1_IT:
+      fdcan1_intr1_it:
         value: 22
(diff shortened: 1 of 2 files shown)


burrbull (dismissed):

burrbull:
Should they are be uppercased?

devices/common_patches/fdcan/fdcan_g4.yaml
```diff
@@ -301,3 +301,11 @@ _include:
...
       CFIE:
         bitWidth: 3
+
+EXTI:
+  _modify:
+    _interrupts:
+      fdcan1_intr0_it:
+        value: 21
+      fdcan1_intr1_it:
+        value: 22
```
burrbull, line 311: Like this:
```suggestion
      fdcan1_intr0_it:
        name: FDCAN1_INTR0_IT
        value: 21
      fdcan1_intr1_it:
        name: FDCAN1_INTR1_IT
        value: 22
```

liamkinne: They aren't uppercase in the original .svd file. It doesn't match properly otherwise.

burrbull (approved):

PR #1004, opened 2024-08-06
Unify g49a and g4a1

usbalbin:
I just had a closer look at `devices/stm32g491.yaml` and  `devices/stm32g4a1.yaml` they seem to have diverged quite a bit. However looking in the [RM](https://www.st.com/resource/en/reference_manual/rm0440-stm32g4-series-advanced-armbased-32bit-mcus-stmicroelectronics.pdf#page=75) they are even in the same column under available peripherals so should be pretty much identical, I think, with AES being the only difference.

Also adds TIM20 to both devices


usbalbin:
I have just taken the config from G491 (except the first line with the svd file) and put into G4A1. Since from what I can tell G491 is the one which has gotten most attention.

burrbull (approved):

PR #1005, opened 2024-08-06
G471 Add OPAMP6

usbalbin:
Hopefully the last PR for now :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,7 @@
 * G4A1: Add FDCAN2 and rename USB peripheral
 * G4A1: Add BCDR and adjust to be closer to G491
 * G491 and G4A1: Add TIM20
+* G471: Add OPAMP6
 
 Family-specific:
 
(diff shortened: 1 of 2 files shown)


burrbull:
Is this normal that there is OPAMP6, but there are not OPAMP4/5?

usbalbin:
I will gether some more info just to be safe.

usbalbin:
There seems to be something weird here. I can not even find STM32G471 on ST's website except for errata and RM. I have also searched mouser, digikey.

I also find it on [ARM's website](https://www.keil.arm.com/devices/stmicroelectronics-stm32g471ceux/features/) but that is pretty much it.

What do we want to do here?

burrbull:
> What do we want to do here?

I can merge this PR unchanged.
But if there are all 6 registers are available I'd merge them in register array.

usbalbin:
What little info I can find says it only has OPAMP1, OPAMP2, OPAMP3 and OPAMP6. Thus a total of 4 op amps. This is also the same as for G491 and G4A1(however they can be found) they have the same set of op amps.

burrbull (approved):

usbalbin:
Thanks :)

PR #1011, opened 2024-09-18
stm32mp1: Fix RCC.RCC_BDCR.DIGBYP Access

### LINKED ISSUE #1010: stm32mp1: DIGBYP bit in RCC_BDCR is incorrectly set to read-only
Hi there!

The DIGBYP bit in the RCC_BDCR register for the stm32mp15{3,7} platforms is incorrectly set to read-only in the SVD files, as you can see from the following snippet:

```
<field>
	<name>DIGBYP</name>
	<description>DIGBYP</description>
	<bitOffset>3</bitOffset>
	<bitWidth>1</bitWidth>
	<access>read-only</access>
</field>
```

The datasheet for the parts states that this field is read/write and what its use is for software (see below images). I only noticed because the STM32CubeMP1 HAL code [clears this bit when setting up](https://github.com/STMicroelectronics/STM32CubeMP1/blob/master/Drivers/STM32MP1xx_HAL_Driver/Src/stm32mp1xx_hal_rcc.c#L921) the RCC external low speed oscillator (LSE).

image

image

I took a quick look through the git commit history and the `extract.sh` script, but I didn't really see a precedent for making corrections to the SVD files. Do you prefer for these changes to be made directly to the SVD files themselves and committed, or would you prefer for them to be added as patches that get applied to the SVD files when `extract.sh` is run?

AmateurECE:
Previously, this bitfield was created with read-only access, though it's supposed to be writable by software. Fixes #1010.

### DIFF
--- a/devices/stm32mp153.yaml
+++ b/devices/stm32mp153.yaml
@@ -22,6 +22,12 @@ TAMP:
   _strip:
     - "TAMP_"
 
+RCC:
+  RCC_BDCR:
+    _modify:
+      DIGBYP:
+        access: read-write
+
 _include:
  - common_patches/rtc/alarm.yaml
  - collect/rtc/tamp_bkpr.yaml
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+* MP15x: Fix DIGBYP bit access in RCC_BDCR
 * Fix yaml parsing errors
 * H7: add bit TRBUFF of DMA_SxCR of H747 cm4
 * Fix inconsistencies for HRTIM_TIMF - stm32g4x4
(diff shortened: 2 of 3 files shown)


devices/stm32mp153.yaml
```diff
@@ -22,6 +22,12 @@ TAMP:
   _strip:
     - "TAMP_"
 
+RCC:
+  RCC_BDCR:
```
burrbull, line 26: It makes sense to strip all `RCC_` prefixes from register names.
```suggestion
RCC:
  _strip: RCC_
  BDCR:
```

AmateurECE: I agree, but if possible, I'd like to make that change wholesale (for all registers and bitfields) in a future PR. I hesitate to make that change for this one register in this PR because that will leave us in an inconsistent state--currently, all registers for this platform are prefixed with the name of the peripheral. Interested to hear your thoughts!

PR #1012, opened 2024-10-02
[WIP] STM32C071 Support

mattthebaker:
stm32c071: update svd and various definitions.


mattthebaker:
I cleaned a few things up but only barely know what I'm doing.

There is a decision that has to be made regarding SPI. C071 has SPI1 and SPI2, so ST decided that SPI on C011 is now SPI1 in line with the updated reference manual. How should this be handled? Inconsistently the SVD for C031 still refers to it as SPI. I believe this is why mmaps fails, not entirely sure what that check is doing.

burrbull:
> not entirely sure what that check is doing

It creates `diff` friendly representation of registers. Very helpful for review. But it needs to run successful.
You can run it manually with `make mmaps` or `cargo make mmaps`.

If you think `SPI` should be `SPI1` you need to add `_modify` block in `stm32c031.yaml`:
```yaml
_modify:
  SPI:
    name: SPI1
```

mattthebaker:
Thanks! It runs clean on the branch but it looks like the mmaps CI action is running the new patches against the old zip which I think might always fail.

> If you think `SPI` should be `SPI1` you need to add `_modify` block in `stm32c031.yaml`:

Well it looks like ST is headed towards it being SPI1 in the docs, but thats going to break any code built on SPI naming. I guess I'm looking for community guidelines on that? I think the answer here is move to SPI1 for the long term by breaking things now.

burrbull:
> CI action is running the new patches against the old zip

You are right. A potential fix is #949. But I'm not good enough in CI to properly review it.

> I think the answer here is move to SPI1 for the long term by breaking things now

Feel free in such changes. Also check RCC reset/enable registers for conformity.

mattthebaker:
Okay, have learned to use svdtools a bit better and made quite a few clean ups and fixes along with reviewing mmaps to make sure all the redundant prefixes are gone.

Its in a pretty good state now but I need to check it against the hal for c031 and c011 -- I've been focused on playing with the new c071 only.

burrbull:
What state of this PR?

mattthebaker:
> What state of this PR?

I've been using it personally along with updates to the HAL, and rolled quite a number of small corrections into it from that effort.

I think its in a good place and could be a candidate for merging. 

The generated registers I feel are in a very good place, I reviewed them at mmap level and made edits as necessary updating the HAL.

I'm not sure how clean my fixups are in the yaml files or if I've met other project standards. Could use a review if possible.

I don't think it should need many more tweaks unless the SVD itself has yet undiscovered errors.

burrbull:
I've merged #949. Could you rebase yet one time.

mattthebaker:
Done, and MMAP check is passing now with that merge, excellent!

burrbull:
> Done, and MMAP check is passing now with that merge, excellent!

Check is passing, but now "compare mmaps" job fails.

burrbull (approved):

burrbull (approved):

PR #1014, opened 2024-10-25
stm32mp15x: Strip register prefixes

AmateurECE:
As promised in the discourse captured on #1011, strip the prefixes from the registers of nearly all the peripherals on these parts.

### DIFF
--- a/devices/common_patches/mpu_unprefix_registers.yaml
+++ b/devices/common_patches/mpu_unprefix_registers.yaml
@@ -1,0 +1,166 @@
+ADC:
+  _strip: "ADC_"
+
+ADC?:
+  _strip:
+    - "ADC_"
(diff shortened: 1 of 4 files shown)


devices/common_patches/mpu_unprefix_registers.yaml
```diff
@@ -0,0 +1,220 @@
+ADC:
+  _strip:
+    - "ADC_"
```
burrbull, line 3: As there is only single entry in the list you could write it shorter
```suggestion
ADC:
  _strip: ADC_
```

AmateurECE: That looks much better. Thanks!

burrbull (approved):

PR #1022, opened 2024-11-10
Hrtim #1021

usbalbin:
Hrtim #1021


usbalbin:
Here @burrbull

burrbull (approved):

burrbull:
@usbalbin Thanks.
What do you think. Is it make sense to remove timer block suffix from register names?
`CMP1xR (RM) => CMP1AR, CMP1BR (SVD now) => CMP1R`

usbalbin:
No problem

While it is a slight diviation from the reference manual, I don't think there is any ambiguity and the HAL implementations will be a lot cleaner having the same name for all timers.

So sure why not

usbalbin:
(Disclaimer, not sure if there is any prior art for other peripherals)

burrbull:
> What do you think. Is it make sense to remove timer block suffix from register names?

Make PR from [hrtim-strip](https://github.com/stm32-rs/stm32-rs/tree/hrtim-strip) branch, please.

PR #1024, opened 2024-11-13
Hrtim: derive identical registers

burrbull:
depends on svd2rust 0.35


burrbull:
@usbalbin This PR goes deeper and reexports most of `HRTIM_TIMA.REG`s as `HRTIM_TIMx.REG` instead of making copies.
So now those are the same Rust types.

usbalbin:
> @usbalbin This PR goes deeper and reexports most of `HRTIM_TIMA.REG`s as `HRTIM_TIMx.REG` instead of making copies.
> So now those are the same Rust types.

Wow, nice! :)

devices/collect/hrtim/v1.yaml
```diff
@@ -0,0 +1,25 @@
+"HRTIM_TIM[BCDEF]":
```
usbalbin, line 1: If this refers to the HRTIM found in the F334 then I believe that does not have timer F

burrbull: It is just rule for search. It applies rules only for founded peripherals. It warnings only if nothing has founded.

burrbull: You can see it in memory map comparison:
https://github.com/stm32-rs/stm32-rs-mmaps/compare/pr-1024-1734720a3f753ade57e0324215b1ba9a9cec7c38#diff-961b24096e627c8e5fbf2c34d8658c99a086687d00ecf7280b92055c1e37d96c

usbalbin: Okay that makes sense, thanks

burrbull:
Could you reopen this PR too?

PR #1028, opened 2024-11-16
Hrtim master

usbalbin:
@burrbull thanks a lot for this


burrbull:
Check "Memory map comparison" for correctness, please.

burrbull (approved):

usbalbin:
> Check "Memory map comparison" for correctness, please.

I will do that as soon as possible

burrbull:
Looks like some issues still present. Especially with `HRTIM_Common`:
https://stm32-rs.github.io/stm32-rs/STM32G484xx.html#HRTIM_Common:FLTINR2

usbalbin:
Also [ISR](https://stm32-rs.github.io/stm32-rs/STM32G484xx.html#HRTIM_Common:ISR) is supposed to be read only from what I can tell (27.5.58 in [RM0440](https://www.st.com/resource/en/reference_manual/rm0440-stm32g4-series-advanced-armbased-32bit-mcus-stmicroelectronics.pdf))

usbalbin:
...more reading

https://stm32-rs.github.io/stm32-rs/STM32G484xx.html#HRTIM_Common:BMCR:BMSTAT

```
27.5.64

Bit 31 BMSTAT: Burst mode status
This bit gives the current operating state.
0: Normal operation
1: Burst operation on-going. **Writing this bit to 0 causes a burst mode early termination**
```

So does this mean that this should default to `1` when writing unless the user has explicitly cleared it? Should this also not be possible to set to `1` by the user?

usbalbin:
Lots of errors in [BMTRGR](https://stm32-rs.github.io/stm32-rs/STM32G484xx.html#HRTIM_Common:BMTRGR) should be `TFCMP1` according to the RM under 27.5.65

* Bit 23 should be TFCMP1
* Bit 21 should be TFREP
* Bit 18 should be TFRST

usbalbin:
[BDTAUPR](https://stm32-rs.github.io/stm32-rs/STM32G484xx.html#HRTIM_Common:BDTAUPR) has a lot of TIMx-prefixes that could probably be removed. There is also the field [TIMx_DTxR](https://stm32-rs.github.io/stm32-rs/STM32G484xx.html#HRTIM_Common:BDTAUPR:TIMx_DTxR) where we might want to strip the last `x`?

Same thing for all registers BDTxUPR where x=A..F

usbalbin:
> Looks like some issues still present. Especially with `HRTIM_Common`: https://stm32-rs.github.io/stm32-rs/STM32G484xx.html#HRTIM_Common:FLTINR2

If you mean things like FLT6SRC being split over [FLT6SRC_0](https://stm32-rs.github.io/stm32-rs/STM32G484xx.html#HRTIM_Common:FLTINR2:FLT6SRC_0)(bit 0) and [FLT6SRC_1](https://stm32-rs.github.io/stm32-rs/STM32G484xx.html#HRTIM_Common:FLTINR2:FLT6SRC_1)(bit 1).

Same thing with FLT1SRC being split between two different registers [FLT[1]SRC](https://stm32-rs.github.io/stm32-rs/STM32G484xx.html#HRTIM_Common:FLTINR1:FLT[1]SRC) (bit 0) in FLTINR1 and [FLT1SRC_1](https://stm32-rs.github.io/stm32-rs/STM32G484xx.html#HRTIM_Common:FLTINR2:FLT1SRC_1) (bit 1) in FLTINR2. From what I can tell from reading RM0440, that is just how weird the hardware is.

When comparing with rm0364 for F334, that processor seems to only have one src bit so probly some attempt at backwards compat?

PR #1029, opened 2024-11-17
Hrtim common

usbalbin:
@burrbull


usbalbin:
Sorry I do not know to read this from the memory map but I believe HRTIM_TIMA.SET[1-4]R is supposed to be different from the ones in the other timers. Since they all have different neighbors. I just want to make sure that is still the case here

usbalbin:
Is it possible to get that nice [web render thing](https://stm32-rs.github.io/stm32-rs/STM32G484xx.html#HRTIM_Common:FLTINR2) directly from a PR, for easier comparison with the register tables in the RMs?

burrbull:
> Is it possible to get that nice [web render thing](https://stm32-rs.github.io/stm32-rs/STM32G484xx.html#HRTIM_Common:FLTINR2) directly from a PR, for easier comparison with the register tables in the RMs?

You can do it locally with:
```
make html
```
or
```
CRATES=stm32g4 make html
```

burrbull:
> Sorry I do not know to read this from the memory map but I believe HRTIM_TIMA.SET[1-4]R is supposed to be different from the ones in the other timers. Since they all have different neighbors. I just want to make sure that is still the case here
> 
> I believe it is the same thing with RSTR and possibly more event subscription related places

Accordingly to RM (if it does not lie) there are `SET[12]R`, `RST[12]R` which are identical and `RSTR` which is different on G4 and identical on F3 and F7.

usbalbin:
> Accordingly to RM (if it does not lie) there are `SET[12]R`, `RST[12]R` [...]

Sorry my bad.

What I was thinking of was the [TIMEVNT[1-9]](https://stm32-rs.github.io/stm32-rs/STM32G484xx.html#HRTIM_TIMA:RST1R:TIMEVNT[9]) bits. They are described in Table 219 of RM0440. It would help from a users perspective if those names were direcly in the register fields descriptions so one does not have to look it up through the table. However I understand if that is out of scope for stm32-rs.

Example:
#####  HRTIM_TIMA
* TIMEVNT[1] - CMPB1 Event from HRTIM_TIMB's CMP1
* TIMEVNT[2] - CMPB2
* TIMEVNT[3] - CMPC2
etc...

#####  HRTIM_TIMB
* TIMEVNT[1] - CMPA1
* TIMEVNT[2] - CMPA2
* TIMEVNT[3] - CMPC3
etc...

Again I understand if that is out of scope for stm32-rs.

usbalbin:
> > Is it possible to get that nice [web render thing](https://stm32-rs.github.io/stm32-rs/STM32G484xx.html#HRTIM_Common:FLTINR2) directly from a PR, for easier comparison with the register tables in the RMs?
> 
> You can do it locally with:
> 
> ```
> make html
> ```
> 
> or
> 
> ```
> CRATES=stm32g4 make html
> ```

Thanks, will try

burrbull:
> What I was thinking of was the [TIMEVNT[1-9]](https://stm32-rs.github.io/stm32-rs/STM32G484xx.html#HRTIM_TIMA:RST1R:TIMEVNT%5B9%5D) bits. They are described in Table 219 of RM0440.

Makes sense. But I'd leave this for another PR.

burrbull (approved):

burrbull:
@usbalbin Look at https://github.com/stm32-rs/stm32-rs/tree/hrtim-set
Is it what you expect?

usbalbin:
Sorry, must have looked in the wrong RM. Bit 31 and 0 in [RSTR](https://stm32-rs.github.io/stm32-rs/STM32G484xx.html#HRTIM_TIMA:RSTR) for timer A-E are TIM**F**CMP2 and TIM**F**CMP1 wich checks out.

However for timer F RSTR should be identical to E's with the exception of bit 31 and 0 which should instead be TIM**E**CMP2 and TIM**E**CMP1 (see 27.5.38 in [RM0440](https://www.st.com/resource/en/reference_manual/rm0440-stm32g4-series-advanced-armbased-32bit-mcus-stmicroelectronics.pdf))

PR #1033, opened 2024-11-20
H5: Add RCC register patches and field definitions for the for H56x/H573 processors

astapleton:
This adds RCC register patches and field definitions for all processors covered by RM0481 (this includes H523/533 processors, but these haven't yet been added as products, yet).

### DIFF
--- a/devices/stm32h562.yaml
+++ b/devices/stm32h562.yaml
@@ -28,7 +28,9 @@ _include:
   - ../peripherals/gpio/gpio_with_brr.yaml
   - ../peripherals/gpio/gpio_with_hslvr.yaml
 
+  - common_patches/rcc/h56x_h57x.yaml
   - ../peripherals/rcc/rcc_h5.yaml
+  - ../peripherals/rcc/rcc_h56x_h57x.yaml
 
   - common_patches/rtc/alarm.yaml
   - common_patches/rtc/rtc_cr.yaml
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+* H5: Add RCC defitions for H56x/H573
 * MP15x: Strip prefixes from peripherals
 * MP15x: Fix DIGBYP bit access in RCC_BDCR
 * Fix yaml parsing errors
--- a/devices/common_patches/rcc/h56x_h57x.yaml
+++ b/devices/common_patches/rcc/h56x_h57x.yaml
@@ -1,0 +1,85 @@
+RCC:
+  PLL3CFGR:
+    _modify:
+      DIVM3:
+        name: PLL3M
+  CCIPR2:
+    _modify:
+      USART12SEL:
+        name: UART12SEL
+  CCIPR4:
+    _add:
+      I3C2SEL:
+        description: I3C2 kernel clock source selection
+        bitOffset: 26
+        bitWidth: 2
+        access: read-write
+  AHB2ENR:
+    _modify:
+      DAC12EN:
+        name: DAC1EN
+        description: DAC clock enable
+      ADC12EN:
+        name: ADCEN
+        description: ADC1 and 2 peripherals clock enable
+  AHB2LPENR:
+    _modify:
+      DAC12LPEN:
+        name: DAC1LPEN
+        description: DAC clock enable during sleep mode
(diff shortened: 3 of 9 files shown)


burrbull:
Thanks for your work.
But I'd want to give some new recommendations about `enumeratedValues`. See [derive-rcc draft branch](https://github.com/stm32-rs/stm32-rs/compare/master...derive-rcc) for references

1. If there is several identical enums in the register, use a **single** rule to add them. This forces to create one Rust `enum` and reuse it instead of several copies.
2. It makes sense to group them by peripheral groups and add `_name` field to fix `enum` name. In other case `field` name (with minimal bitOffset) will be used for `enum` name.
3. (not required but recommended). Use `PCLK` without index. So you can use (just import from the PAC) single enum for peripherals on different busses). It much easier to differentiate it on HAL level instead of PAC level with something like [RccBus trait](https://github.com/stm32-rs/stm32f4xx-hal/blob/f3e5246f9ba54de574ab01dcc05a253e41dda43a/src/rcc/mod.rs#L47).
For example:
```yaml
    USART*SEL:
      _name: UartSel
      PCLK: [0, "PCLKx selected as clock source (rcc_pclk_x)"]
      PLL2_Q: [1, "PLL2 Q clock selected as clock source (pll2_q_ck)"]
      HSI_KER: [3, "HSI kernel clock selected as clock source (hsi_ker_ck)"]
      CSI_KER: [4, "CSI kernel clock selected as clock source (csi_ker_ck)"]
      LSE: [5, "LSE clock selected as clock source (lse_ck)"]
```
4. (also just a recommendation). Now fields also can be derived (reexported) across registers:
```yaml
RCC:
  CCIPR3:
    _modify:
      LPUART1SEL:
        derivedFrom: RCC.CCIPR1.USART1SEL
```

astapleton:
Thanks for the suggestions. 

I agree with points 1, 3, 4 and will implement those changes were appropriate.

However, I'm not sure I agree with point 2 with respect to renaming USART to UART. The reference manual actually distinguishes them in multiple places: the USART is capable of synchronous operation like SPI, and can in fact be used as a SPI port on these parts, while the UART cannot, it is only asynchronous. The enable bits and kernel clock selections all distinguish between USART and UART, and it may actually be confusing to a consumer looking for USART10, and finding UART10 in the PAC, and wondering which source of information is correct about it's capabilities. 

I'm also happy to make the change to the H503 definitions to be consistent with these suggestions too, unless you'd rather I didn't touch those now?

burrbull:
> However, I'm not sure I agree with point 2 with respect to renaming USART to UART.

I was saying not about this. I just recommend to specify `_name` manually instead of rely on automatic naming when there is several fields under one rule.

astapleton:
> Is that a bug? I can work around it by inserting an underscore like so:
> 
> ```
>     SPI[456]SEL:
>       _name: Spi456_Sel
> ```
> 
> That produces an enum name of `SPI456_SEL` so it's consistent with the UART naming. But that seems like an unexpected inconsistency.

The underscore is also inconsistent with the rest of the enum naming for fields that are not renamed via that mechanism, so it seems like the presence of the underscore is a bug?

astapleton:
Specifying everything in uppercase works around the underscore bug:
```
USART*SEL,UART*SEL:
      _name: USARTSEL
```

astapleton:
@burrbull I've incorporated all your suggestions (with the uppercase workaround) into the latest commit

burrbull (approved):

devices/stm32h562.yaml
```diff
@@ -28,7 +28,9 @@ _include:
   - ../peripherals/gpio/gpio_with_brr.yaml
   - ../peripherals/gpio/gpio_with_hslvr.yaml
 
+  - common_patches/rcc/h56x_h57x.yaml
   - ../peripherals/rcc/rcc_h5.yaml
```
burrbull, line 32: `rcc_h5.yaml` is already included in `peripherals/rcc/rcc_h56x_h57x.yaml`. Remove this line, please. Same for h56x/h57x

astapleton: Do you want me to revert and reopen this PR or submit a new PR?

burrbull: New PR, of cause.

PR #1035, opened 2024-11-22
Make TIM4 16-bit on stm32g0b1

rblaze:
Currently it is incorrectly implemented as 32-bit timer.


burrbull:
If `TIM4` is identical to `TIM3` I think it is much more sense to derive it from `TIM3`:
```yaml
_derive:
  TIM4: TIM3
```

rblaze:
Good point, done.

PR #1039, opened 2024-12-04
G0: describe USB CHEP?R register

rblaze:
Add enums for UTYPE, STATRX, and STATTX fields in CHEP_R.

Unfortunately, this also generates `disabled()/stall()/nak()/valid()` methods in STAT* writers, which make no sense: the field is "1-to-toggle" and needs to be XORed with an existing value to set a new one. I wonder if there is a way to disable writer helpers for "toggle" fields.


burrbull:
> Unfortunately, this also generates `disabled()/stall()/nak()/valid()` methods in STAT* writers,

This is very strange. Looks like (`svd2rust`?) bug.

burrbull:
Oh. I understood. The issue is `svd2rust` does not support properly non-1-bit fields with `modifiedWriteValues` yet.

burrbull (approved):

PR #1088, opened 2024-12-30
Update field definitions in USB ISTR, USB CHEPR and EXTICR

rblaze:
Marking USB bits to avoid accidental flips.
EXTI only had human-readable enums in CR1 until this change.

### DIFF
--- a/devices/fields/exti/exti_g0.yaml
+++ b/devices/fields/exti/exti_g0.yaml
@@ -3,7 +3,7 @@ _include:
   - derive.yaml
 
 EXTICR1:
-  "EXTI*":
+  EXTI0:
     PA: [0, "Select PAx as the source input for the EXTIx external interrupt"]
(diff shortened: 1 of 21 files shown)


devices/fields/exti/exti_g0.yaml
```diff
@@ -2,7 +2,7 @@ _include:
   - common.yaml
   - derive.yaml
 
-EXTICR1:
+EXTICR*:
```
burrbull, line 5: Revert this change, please.

burrbull: Those fields already derived in 3 line.

burrbull: Sorry. There really was an error. Thanks for finding it. Fixed.

burrbull (dismissed):

burrbull (approved):

PR #1107, opened 2025-01-06
H7 add missing series

burrbull:


burrbull:
@usbalbin Looks at this, please. And try to merge if OK.

usbalbin:
> @usbalbin Looks at this, please. And try to merge if OK.

Sure, anything in perticular or just in general?

burrbull:
In general.

scripts/makecrates.py
```diff
@@ -28,7 +28,7 @@
     "stm32f4": ["atomics", "critical-section", "defmt", "rt", "stm32f401", "stm32f407", "stm32f413", "stm32f469"],
     "stm32f7": ["atomics", "critical-section", "defmt", "rt", "stm32f733", "stm32f779"],
     "stm32h5": ["atomics", "critical-section", "defmt", "rt", "stm32h503", "stm32h523", "stm32h562", "stm32h573"],
```
usbalbin, line 30: Completely unrelated to this PR but I believe the "stm32h563" feature was removed in #1124 https://github.com/stm32-rs/stm32-rs/pull/1124/commits/6cf40035f18ce99355903ef3672a0fd5b103566b#r1924396832

devices/stm32h7b3.yaml
```diff
@@ -155,6 +155,14 @@ RTC:
     - fields/rtc/v3.yaml
     - collect/rtc/alarm.yaml
 
+
+SAI1:
```
usbalbin, line 159: Still trying to wrap my head around how to review something like this...

Why is this only added for the h7b3 and not h7a3 and h7b0? Is the bug only in the svd for the h7b3? From what I can see they all have those peripherals, right?

scripts/makecrates.py
burrbull: those are not "features" which isbacceptable. those are features which will be rendered on docs.rs

devices/stm32h7b3.yaml
burrbull: Interrupts is better to look in the end of "mmap": https://github.com/stm32-rs/stm32-rs-mmaps/blob/86e7310017ab9a8919d985c3580e75ff6e81ca2a/stm32h7a3.mmap

> Still trying to wrap my head around how to review something like this...

Do not overwork yourself. I've asked you first of all to test your write permissions in the repo.

scripts/makecrates.py
usbalbin: Okay, thanks

devices/stm32h7b3.yaml
usbalbin: Thanks :)

usbalbin (approved):
I still dont understand a lot of this and can not really tell if there is a problem. However for what it's worth, I can not find anyting obviously wrong :)

usbalbin:
image

I take it this means I now have write permissions. Thanks :)

Do you want me to merge this now?

burrbull:
Let's go.

PR #1117, opened 2025-01-17
H5: Cluster GPDMA channels

astapleton:
Adds cluster definitions for H5 GPDMA channels. Differentiate 0-5 from 6-7 based on 2D addressing features present in 6-7.


burrbull:
I'm trying to get away from using global `_include`s.

Please rebase up on https://github.com/stm32-rs/stm32-rs/tree/h5-gpdma-channels
I've moved some files, added clusters for U5 and field arrays for common registers.

Issue #957, opened 2024-02-04
STM32G4 SPI4 has swapped reset values for CR1 and CR2


techmccat:
I notieced while looking at the g474 svd files because svd2rust had created different modules for spi1 and spi4, despite the manual not saying anything about different registers.

spi2 and spi3 are affected too since they re-export spi4.
Fixing this difference could also allow all the spi peripheral to share RegisterBlock definition (thus implementing Deref for the same type)

Register definitions are on page 1787 of the g4 manual [found here](https://www.st.com/resource/en/reference_manual/rm0440-stm32g4-series-advanced-armbased-32bit-mcus-stmicroelectronics.pdf)
Here's the diff of the peripheral definitions I manually copy-pasted from the svd
```diff
2c2
< <name>SPI1</name>
---
> <name>SPI4</name>
5c5
< <baseAddress>0x40013000</baseAddress>
---
> <baseAddress>0x40013C00</baseAddress>
12,14c12,14
< <name>SPI1</name>
< <description>SPI1</description>
< <value>35</value>
---
> <name>SPI4</name>
> <description>SPI4</description>
> <value>84</value>
24c24
< <resetValue>0x00000000</resetValue>
---
> <resetValue>0x00000700</resetValue>
119c119
< <resetValue>0x00000700</resetValue>
---
> <resetValue>0x00000000</resetValue>
```

pietgeursen:
I just hit this issue too using stm32-hal2. It means I can only use SPI1 rather than all 4.

PR #1127, opened 2025-01-23
Fix resetValue of g4 SPI4 CR1 and CR2

pietgeursen:
Fixes issue described in #957 

This is my first time delving into svd file so not sure if I'm doing this correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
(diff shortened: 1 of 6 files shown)


burrbull:
As those peripherals are identical it is better to "derive" SPI4 from SPI1. Try add this in root files instead:
```yaml
_derive:
  SPI4: SPI1
```

pietgeursen:
Root files?

pietgeursen:
> Root files?

Nevermind, figured it out :)

burrbull (approved):

PR #1133, opened 2025-01-24
TIM refactor, part 3: fields

burrbull:
cc @usbalbin


devices/collect/tim/ccr.yaml
```diff
@@ -20,14 +28,16 @@ CCER:
...
-    CC?S:
-      description: Capture/Compare %s selection
     IC?PSC:
       description: Input capture %s prescaler
     IC?F:
       description: Input capture %s filter
+"?~CCMR2_Input*":
+  _derive:
+    IC%sF: CCMR1_Input.IC%sF
+    IC%sPSC: CCMR1_Input.IC%sPSC
```
usbalbin, line 40: What does the `?~` mean here?

Also, will the `%s` be replaced by whatever the `*` matched?

burrbull: `?~` means skip rule if name does not exists.
Without this it will cause "not found error".

`%s` is array suffix placeholder. See SVD specification. https://arm-software.github.io/CMSIS_5/SVD/html/elem_special.html#dimElementGroup_gr

devices/fields/tim/bdtr_common.yaml
```diff
@@ -0,0 +1,41 @@
...
+    ActiveLow: [0, Break input BRKx is active low]
+    ActiveHigh: [1, Break input BRKx is active high]
+  "BKE,BK2E":
+    Disabled: [0, Break function x disabled]
+    Enabled: [1, Break function x enabled]
+  OSSR:
+    Disabled: [0, "When inactive, OC/OCN outputs are disabled"]
+    IdleLevel: [1, "When inactive, OC/OCN outputs are enabled with their inactive level"]
+  OSSI:
+    Disabled: [0, "When inactive, OC/OCN outputs are disabled"]
```
usbalbin, line 22: What about something like `HiZ` here? I'm fine either way

devices/fields/tim/ccpc.yaml
```diff
@@ -0,0 +1,11 @@
+CR2:
+  CCPC:
+    NotPreloaded: [0, "CCxE, CCxNE and OCxM bits are not preloaded"]
+    Preloaded: [1, "CCxE, CCxNE and OCxM bits are preloaded"]
+  CCUS:
+    Bit: [0, "When capture/compare control bits are preloaded (CCPC=1), they are updated by setting the COMG bit only"]
+    BitOrEdge:
```
usbalbin, line 7: What about something like `Sw` and `SwOrEdge`?

devices/collect/tim/ccr.yaml
usbalbin: Ok, thanks

devices/fields/tim/v1/tim1.yaml
```diff
@@ -0,0 +1,15 @@
...
+  - ../generic.yaml
+  - ../tim_ckd.yaml
+  - ../tim_advanced.yaml
+  - ../ccr.yaml
+  - ../ccxnp.yaml
+  - occe.yaml
+  - ccm.yaml
+
+RCR:
+  REP: [0, 0xFF]
```
usbalbin, line 15: Is this the valid range of values? If so, atleast for RM0440 and RM0444 this seems to be 16 bits.

usbalbin: However I have not checked if those are tim v1...

usbalbin (dismissed):
LGTM from what I can tell

devices/fields/tim/v1/tim1.yaml
burrbull: This is `v1`.
To make more clear how I've splitted them see
изображение

devices/fields/tim/bdtr_common.yaml
burrbull: high impedance?

burrbull:
update approve, please

devices/fields/tim/bdtr_common.yaml
usbalbin: Yes, atleast that is mentioned in RM0440 and RM0444. RM0440:

> 0:When inactive, OC/OCN outputs are disabled (the timer releases the output control which is taken over by the GPIO logic, which forces a Hi-Z state).

But I dont know what is more useful to the user

usbalbin (approved):

usbalbin:
I take it this is the kind things that lets the user avoid the unsafe `w.field().bits()` method and also makes a lot of things clearer? :)

PR #1144, opened 2025-01-29
GPIO registers/fields derive

burrbull:


burrbull:
cc @usbalbin 
Peripherals can't be derived due to different register resetValues. For the same reason we can't derive only part of registers and derive fields for other.

burrbull:
> Will this then lead to a lot of the registers and/or register blocks sharing the same types? Or atleast those with identical reset values etc?

Register structures only which are identical. Enums and field structures all should be shared. The exception are WB/WL serie GPIOC/GPIOH registers which are little crazy.

PR #1121, opened 2025-01-19
H5 & U5: Cluster GPDMA channels

z-av:
Replaces #1117

PR #1149, opened 2025-01-30
H5: gpdma: Restore PAM_1, PAM_2 enums

astapleton:
This fixes a regression in #1121 which removed the PAM_1 and PAM_2 definitions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+* H5: gpdma: Restore PAM_1, PAM_2 enums
 * Refactor timers, add enums
 * STM32H5xx: Add H533 (#1129)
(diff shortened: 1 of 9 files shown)


burrbull:
As I've already said there is prohibited to use overlapping fields but allowed overlapping registers in SVD specification.
You can make sure of it by running `svdconv` tool.

Even if it is allowed by `svd2rust` I'd prefer to avoid adding new bugs.

astapleton:
> As I've already said there is prohibited to use overlapping fields but allowed overlapping registers in SVD specification. You can make sure of it by running `svdconv` tool.
> 
> Even if it is allowed by `svd2rust` I'd prefer to avoid adding new bugs.

Ah ok. We can close this then. What's the reasoning there, if I may know?

burrbull:
> What's the reasoning there, if I may know?

I don't know.

Maybe to prevent a real overlap bugs which are very often.

PR #1155, opened 2025-02-02
USBSEL 0x2: PLL2Q for H503, PLL3Q for rest of H5

usbalbin:
From what I can tell only [H503](https://stm32-rs.github.io/stm32-rs/STM32H503.html#RCC:CCIPR4:USBSEL) should have PLL2Q as option for USBSEL with value 0x2. The rest of the timers should be [PLL3Q](https://stm32-rs.github.io/stm32-rs/STM32H533.html#RCC:CCIPR4:USBSEL)


usbalbin:
cc @burrbull

burrbull (approved):

PR #1157, opened 2025-02-02
F2, F4, F7: Add definitions for OPTCRx register of FLASH

### LINKED ISSUE #1154: How to handle field value as flags instead of enum ?
I'm trying to document the `OPTCR` and `OPTCR1` registers of the FLASH peripheral for STM32F4 devices, but I don't know how to handle the `nWRP` field.

The FLASH memory of STM32F4 devices is split into several sectors, 12 or 24 sectors depending on device and memory configuration. The `nWRP` field is 12 or 24 bits long (to match the number of sectors of the FLASH memory), with each bit corresponding to the write protection status of the a sector.

This means the value stored in `nWRP` shouldn't be treated as an enumeration but rather as a flags.

Here's an extract of ST documentation:

Image

Image

I tried (without success) to create a patch for `OPTCR` and `OPTCR1` registers using `_split` which looks like this

```yaml
OPTCR:
  nWRP:
    _split: 
      nWRPx:
        name: nWRP%s
        description: Value of the write-protection option bytes after reset for sector %s
```

But the `svdtools` command fails with the following message

```bash
svdtools patch devices/stm32f405.yaml
[2025-02-02T10:50:12Z ERROR svdtools::cli] by svdtools (0.4.3)
    
    Caused by:
        0: Processing device `STM32F405`
        1: According to `FLASH`
        2: Processing peripheral `FLASH`
        3: According to `OPTCR`
        4: Processing register `OPTCR`
        5: Processing field matched to `nWRP`
        6: Adding read-write enumeratedValues
        7: `EnumeratedValues error: EnumeratedValues is empty
        8: EnumeratedValues is empty
make: *** [Makefile:50: svd/stm32f405.svd.patched] Error 1
```

Is this the correct way to do this ? Is it even possible ?

vinchatl:
This PR adds definitions for OPTCR, OPTCR1 and OPTCR2 registers for all F2, F4 and F7 devices.

## Patch structure

Most of the fields are implemented the same way across the target devices, but there are still differences between models. Taking the whole range of devices, there are at least 7 OPTCR variants, 2 OPTCR1 variants and 1 OPTCR2 variant.

To share as much code as possible while enabling a lean composition in the device yaml, the added patch files are organised like this:

- common fields are described in `devices/fields/flash/flash_optcr.yaml`
- specific fields are described in their specific yaml file `devices/fields/flash/flash_optcr_*field*.yaml`

Register `OPTCR1` doesn't really have a default variant so it is just split between `devices/fields/flash/flash_optcr1_boot.yaml` or `devices/fields/flash/flash_optcr1_nwrp.yaml`.

Register `OPTCR2` only has 1 variant, written in `devices/fields/flash/flash_optcr2.yaml`.

## Bit fields management

There are 2 peculiar fields in these registers, `nWRP` and `PCROP`, which are not enums but bit fields (or flags).

To enable a dedicated read/write access to each bit of the fields, there are split in N fields using the _split_ command in `devices/patches/flash/optcr_*.yaml`, then collected as an array in `devices/collect/flash/optcr*_*.yaml`.

### nWRP

This technic is especially usefull for the `nWRP` field, as it doesn't have the same size across F2, F4 and F7 devices (it can be 5, 8, 12, 15 bits long). Using the `*` wildcard character ensures the _split_ and _collect_ methods covers all these sizes.

Using `?` doesn't work in this case, as it only counts for a single character, thus limiting the array index to 9.

### PCROP

On the other hand, the `PCROP` register uses `?` as its index is limited to 8, and the pattern `PCROP*` would conflict with another field named `PCROP_MOD` in the same register.

## Miscellaneous

During testing I noticed some issues with SVDs for 401, 411, 412, 413 and 730 devices. A patch has been created here while the issue is handled upstream by ST developpers.

I also noticed inconsistencies in the `nWRP` bit size between SVD and reference manuals, I didn't create a patch as I didn't feel it was a major issue, but I'll report this upstream anyway.

Closes #1154.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -138,6 +138,7 @@
 * G0: mark flags in USB CHEPR as W0C/W1T
 * G0: add value enums for EXTICR[2-4]
 * H5, U5: Add cluster definitions for GPDMA channels
+* F2, F4, F7: Add definitions for OPTCR, OPTCR1 and OPTCR2 registers of FLASH peripheral
 
 Family-specific:
 
--- a/devices/collect/flash/optcr1_nwrp.yaml
+++ b/devices/collect/flash/optcr1_nwrp.yaml
@@ -1,0 +1,3 @@
+OPTCR1:
+  _array:
+    nWRP*: {}
--- a/devices/collect/flash/optcr2_pcrop.yaml
+++ b/devices/collect/flash/optcr2_pcrop.yaml
@@ -1,0 +1,3 @@
+OPTCR2:
+  _array:
+    PCROP?: {}
--- a/devices/collect/flash/optcr_nwrp.yaml
+++ b/devices/collect/flash/optcr_nwrp.yaml
@@ -1,0 +1,3 @@
+OPTCR:
+  _array:
+    nWRP*: {}
--- a/devices/fields/flash/flash_optcr.yaml
+++ b/devices/fields/flash/flash_optcr.yaml
@@ -1,0 +1,32 @@
+# Description of OPTCR register with nWRP only.
+OPTCR:
+  nWRP*:
+    Active: [0, Write protection active on sector %s]
+    Inactive: [1, Write protection inactive on sector %s]
+  RDP:
+    Level0: [0xAA, Read protection not active]
+    Level2: [0xCC, Chip read protection active]
+    Level1: [-1, Read protection of memories active]
+  nRST_STDBY:
+    Reset: [0, Reset generated when entering Standby mode]
+    NoReset: [1, No reset generated]
+  nRST_STOP:
+    Reset: [0, Reset generated when entering Stop mode]
+    NoReset: [1, No reset generated]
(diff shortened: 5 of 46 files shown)


burrbull (approved):

burrbull:
Good work. Thanks.

vinchatl:
Hi @burrbull, github-actions bot just published a memory map comparison for U5 mcus now that the PR is closed, while it hasn't published anything on F2 F4 F7 devices when I first opened the PR, is it drunk ?

burrbull:
This is due to bug in mmaps CI.
It was splitted on "Generate PR mmaps" and "Compare PR mmaps" to allow to pass when new yamls are added.
But now they often do not synchronize and second one can't get correct `workflow_id`.
I think that the issue is due to permissions of PR creator, but I do not know how to fix it.
cc @adamgreig @Emilgardis @Elemecca

PR #1173, opened 2025-02-19
TIM 16bit v2

burrbull:


usbalbin (approved):
So i take it this sets the bitwidth for a lot of the timer registers? I have tried to look it through as best I can. LGTM for what it's worth :)

PR #1184, opened 2025-03-31
TIM derives

burrbull:


burrbull:
cc @usbalbin

burrbull:
This is only first easiest step I've planned.
Also I temporary skipped CCR%s due to bug in svd2rust.

PR #1186, opened 2025-04-02
TIM: TS SMS MMS

burrbull:
Unify `TS`/`SMS`/`MMS` field names across MCs and timers. 

Similar to `OCxM` field.

cc @usbalbin


devices/patches/tim/ts.yaml
```diff
@@ -0,0 +1,6 @@
+SMCR:
+  _modify:
+    "TS,TS1,TS_2_0":
+      name: TS
+    "TS2,TS3_4,TS_4_3,TS_[12],TS4_3":
+      name: TS2
```
burrbull, line 6: Not sure how to name this part of `TS`.
As it width is 2-bit (3th & 4th) and compared to 3th bit of OCxM, SMS and MMS.

usbalbin: Sorry for the very late response, the last several weeks have been quite hectic at work..

Hm, good question.. There is TS_H/TS_L but I think TS2/TS works too.

Would it make sense/be possible to be more consistent with SMS being SMS_3 and SMS and same thing with MMS. (I assume the `3` is for bit 3?). While TS is TS2. Perhaps TS3 or TS_3 for "start-bit 3" to be more consistent with SMS and MMS? Either that or perhaps have SMS2 and MMS2? Not sure if there are other situations like this that have good solutions to look at?

PR #1199, opened 2025-04-23
SPI v3/v4

burrbull:
cc @usbalbin


devices/fields/spi/spi_v3.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - spi_v3_base.yaml
+  - spi_v3_i2s.yaml
+
+CFG1:
+  UDRDET:
+    StartOfFrame: [0, Underrun is detected at begin of data frame]
+    EndOfFrame: [1, Underrun is detected at end of last data frame]
+    StartOfSlaveSelect: [2, Underrun is detected at begin of active SS signal]
+  UDRCFG:
+    RepeatTransmitted: [2, Slave repeats last transmitted data frame]
```
usbalbin, line 11: Should there also be the variants for `0` and `1`?

[RM0433](https://www.st.com/resource/en/reference_manual/rm0433-stm32h742-stm32h743753-and-stm32h750-value-line-advanced-armbased-32bit-mcus-stmicroelectronics.pdf) page 2221:
> Bits 10:9 UDRCFG[1:0]: behavior of slave transmitter at underrun condition
>   00: slave sends a constant pattern defined by the user at SPI_UDRDR register
>   01: slave repeats lastly received data frame from master
>   10: slave repeats
>   11: reserved

devices/fields/spi/spi_v3.yaml
```diff
@@ -0,0 +1,21 @@
...
+CR2:
+  TSER: [0, 65535]
+IER:
+  TSERFIE:
+    Masked: [0, TSER loaded interrupt masked]
+    NotMasked: [1, TSER loaded interrupt not masked]
```
usbalbin, line 17: What about enabled/disabled instead of masked/notmasked?
```suggestion
    Disabled: [0, TSERF interrupt disabled]
    Enabled: [1, TSERF interrupt enabled]
```
From [RM0433](https://www.st.com/resource/en/reference_manual/rm0433-stm32h742-stm32h743753-and-stm32h750-value-line-advanced-armbased-32bit-mcus-stmicroelectronics.pdf) page 2224:

> Bit 10 TSERFIE: additional number of transactions reload interrupt enable
>   0: TSERF interrupt disabled
>   1: TSERF interrupt enabled

devices/fields/spi/spi_v3_i2s.yaml
```diff
@@ -0,0 +1,45 @@
+I2SCFGR:
+  MCKOE:
+    Disabled: [0, Master clock output disabled]
+    Enabled: [1, Master clock output enabled]
+  ODD:
+    Even: [0, Real divider value is I2SDIV*2]
+    Odd: [1, Real divider value is I2SDIV*2 + 1]
+  I2SDIV: [0, 255]
```
usbalbin, line 8: RM433 page 2233 mentions
> Bits 23:16 I2SDIV[7:0]: I2S linear prescaler
>I2SDIV can take any values except the value 1, when ODD is also equal to 1.

Is this something that changes anything?

burrbull: When range is defined svd2rust adds safe `fn set` in addition to `unsafe fn bits`.

But in this case it depends on other field state. So we should delete this.
I'm not sure.

devices/fields/spi/spi_v3.yaml
burrbull: Good catch. Thanks

devices/fields/spi/spi_v3.yaml
burrbull: Different enums for each field are not needed here. Replaced with:
```yaml
  "*IE":
    Disabled: [0, Interrupt disabled]
    Enabled: [1, Interrupt enabled]
```

devices/fields/spi/spi_v3.yaml
burrbull: Reverted. Already present in `spi_v3_base.yaml`.

devices/fields/spi/spi_v3.yaml
usbalbin: nice

devices/fields/spi/spi_v3_i2s.yaml
usbalbin: Ok, good no know. Thanks

usbalbin (approved):
Nice LGTM! :)

PR #1200, opened 2025-04-24
FMC coverage & derive

burrbull:


burrbull:
cc @usbalbin

devices/fields/fsmc/sram.yaml
```diff
@@ -32,6 +35,7 @@ BCR?:
   FACCEN:
     Disabled: [0, Corresponding NOR Flash memory access is disabled]
     Enabled: [1, Corresponding NOR Flash memory access is enabled]
+  # TODO: fix
```
usbalbin, line 38: Why the todo?

usbalbin (approved):
Not really familiar with the FMC, but LGTM as far as I can tell

devices/fields/fsmc/sram.yaml
burrbull: There is also can be Bits32 in MWID.
Need to investigate more. In another PR maybe.

PR #1202, opened 2025-04-27
USB v2

burrbull:


burrbull:
cc @usbalbin

devices/fields/usb/v2_with_LPM.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+LPMCSR:
+  BESL: [0, 0b1111]
+  REMWAKE:
+  LPMACK:
+    Nyet: [0, the valid LPM Token will be NYET]
+    Ack: [1, the valid LPM Token will be ACK]
+  LPMEN:
+    Disabled: [0, enable the LPM support within the USB device]
+    Enabled: [1, no LPM transactions are handled]
```
usbalbin, line 14: I have not looked through the other RM's but RM0440 under USB_LPMCSR states:

>Bit 0 LPMEN: LPM support enable
>This bit is set by the software to enable the LPM support within the USB device. If this bit is
at ‘**0 no LPM transactions are handled**

devices/fields/usb/v1.yaml
```diff
@@ -137,6 +137,9 @@ DADDR:
 BTABLE:
   BTABLE: [0, 0b1111111111111]
 EP*R:
+  _modify:
+    STAT_[RT]X:
+      access: read-write
```
usbalbin, line 142: Just out of curiosity, this bit is marked as `t` for `toggle` in [RM0440](https://www.st.com/resource/en/reference_manual/rm0440-stm32g4-series-advanced-armbased-32bit-mcus-stmicroelectronics.pdf). Is that something we care about?

> Bits 13:12 STAT_RX [1:0]: Status bits, for reception transfers
> [...] These bits can be toggled by software to initialize
their value. When the application software writes ‘0, the value remains unchanged, while
writing ‘1 makes the bit value toggle.

burrbull: There is `_W1T` in line 153.

devices/fields/usb/v2_with_LPM.yaml
burrbull: Thanks.

devices/fields/usb/v1.yaml
usbalbin: Oh, ok. Interesting. Thanks :)

usbalbin (approved):
LGTM

PR #1211, opened 2025-05-01
Add stm32n6

jaskij:
PR to add the STM32N6 family. Points of concern:

- The STM32N6x5 line does not seem to have a publicly available reference manual
- I have no clue how to modify `CRATE_DOC_FEATURES`, so I went monkey see monkey do
- I could not run `scripts/matchperipherals.py` (I'll open an issue on this one)
- `make` failed

Now, on to why `make` failed: whatever is ran inside, seems to expect that the `*.svd.patched` file already exists - which it does not, since I'm adding new MCUs. I *could* just wing it and create those files manually or something, but I'm not sure that's a good idea.

```
svdtools patch devices/stm32n645.yaml
[2025-05-01T14:26:22Z ERROR svdtools::cli] by svdtools (0.4.6)

    Caused by:
        No such file or directory (os error 2)
make: *** [Makefile:50: svd/stm32n645.svd.patched] Error 1
```


burrbull (approved):

burrbull:
> Now, on to why `make` failed

You just made a typo in svd file path.

I've fixed the PR and added minimal patches. So you can start compare N6 peripherals with similar families and try to apply existing patches when possible.

burrbull:
@jaskij existing enumerates values were deleted from SVDs.
But they are kept in [enums branch](https://github.com/stm32-rs/stm32-rs/tree/enums/enums). So you can use them as reference if you want to re-add enums with nice names. 
For example
https://raw.githubusercontent.com/stm32-rs/stm32-rs/refs/heads/enums/enums/stm32n657.yaml

PR #1218, opened 2025-05-04
DFSDM

burrbull:
@usbalbin


devices/fields/dfsdm/f7.yaml
```diff
@@ -0,0 +1,316 @@
...
+  AWHT: [0, 0xFFFFFF]
+  BKAWH: [0, 0xF]
+
+DFSDM[0123]_AWLTR:
+  AWLT: [0, 0xFFFFFF]
+  BKAWL: [0, 0xF]
+
+DFSDM[0123]_AWSR:
+  AWHTF: [0, 0xF]
+  AWLTF: [0, 0xF]
```
usbalbin, line 301: This is 8bits from what I can tell from [rm0433](https://www.st.com/resource/en/reference_manual/rm0433-stm32h742-stm32h743753-and-stm32h750-value-line-advanced-armbased-32bit-mcus-stmicroelectronics.pdf) 30.8.11

devices/fields/dfsdm/f7.yaml
```diff
@@ -0,0 +1,316 @@
...
+
+DFSDM[0123]_AWCFR:
+  CLRAWHTF: [0, 0xF]
+  CLRAWLTF: [0, 0xF]
```
usbalbin, line 305: Same here

devices/fields/dfsdm/f7.yaml
burrbull: Now when I've look at this closer I think these fields need to split and collect in array.

usbalbin (approved):
Nice! LGTM as far as I can tell :)

PR #1219, opened 2025-05-04
H7 rcc - Remove hs from usb1otghs

usbalbin:


devices/patches/rcc/h7_usbotghs.yaml
```diff
@@ -0,0 +1,10 @@
+AHB1LPENR,C1_AHB1LPENR:
+  _modify:
+    USB1OTGLPEN,USB1OTGHSLPEN: #TODO: Help
+      name: USB1OTGLPEN
+    USB1OTGULPILPEN,USB1ULPILPEN,USB1OTGHSULPILPEN: #TODO: Help
+      name: USB1OTGULPILPEN
```
usbalbin, line 6: @burrbull some of the devices have some, but not all, of the fields correct. Is there a better way to do this?

burrbull:
```yaml
    "?~USB1OTGHSLPEN":
      name: USB1OTGLPEN
    "?~USB1ULPILPEN,USB1OTGHSULPILPEN":
      name: USB1OTGULPILPEN
```
Starting with `?~` means "if present".
If I understood your question right.

usbalbin:
> squash commits and add to changelog, please.

All into one single commit? Or one for renaming fields and one for deleting fields(without the mistakes)?

Also do you think this is correct? From what I can tell the 18th bit `USB2OTGHSULPIEN` should not exist, hence i removed it

burrbull:
The change is not a big. So single commit is fine.

>  Also do you think this is correct? From what I can tell the 18th bit USB2OTGHSULPIEN should not exist, hence i removed it

Looks correct.

burrbull (approved):

PR #1224, opened 2025-05-06
H7 hi mem flash

usbalbin:


devices/patches/rtc/h7_hi_mem.yaml
```diff
@@ -0,0 +1,3 @@
+_modify:
+  ICSR:
+    name: ISR
```
usbalbin, line 3: @burrbull 
What am I doing wrong here?

Without this fix:
```
$ cargo c --features stm32h7b3,rt,defmt
[...]
error[E0599]: no method named `isr` found for struct `Periph<stm32h7::stm32h7b3::rtc::RegisterB
lock, 1476411392>` in the current scope
   --> src/rtc.rs:765:26
    |
765 |                 self.reg.isr().modify(|_, w| w.wutf().clear_bit());
    |                          ^^^
    |
help: there is a method `icsr` with a similar name
    |
765 -                 self.reg.isr().modify(|_, w| w.wutf().clear_bit());
765 +                 self.reg.icsr().modify(|_, w| w.wutf().clear_bit());
[...]
```

usbalbin: For reference I can see [ICSR here](https://stm32-rs.github.io/stm32-rs/STM32H7A3.html#RTC:ICSR)

usbalbin: Or perhaps this should not even be done in the first place. There does seem to be quite a bit of differences between the RTC in RM0455 as compared to RM0433. In RM0433 we have the battery backed ram in RTC_BKP[0-31]R but I can not seem to find that in RM0455 even though it is mentioned as a feature from what I can tell.

burrbull: You should not do this on PAC level. H7A/B have newer RTC.
ISR and ICSR work different.

devices/patches/flash/h7_hi_mem.yaml
```diff
@@ -0,0 +1,4 @@
+CR[12]:
+  _modify:
+    SSN[12]:
+      name: SNB
```
usbalbin, line 4: How about this? Do you think this should be done? From what I can tell, the SSN in [RM0455](https://www.st.com/resource/en/reference_manual/rm0455-stm32h7a37b3-and-stm32h7b0-value-line-advanced-armbased-32bit-mcus-stmicroelectronics.pdf) 4.9.4 seems to do the same thing as SNB from [rm0433](https://www.st.com/resource/en/reference_manual/rm0433-stm32h742-stm32h743753-and-stm32h750-value-line-advanced-armbased-32bit-mcus-stmicroelectronics.pdf) 4.9.4, however at a different bit position and width

rm0455
> Bits 12:6 SSN1: Bank 1 sector erase selection number
> These bits are used to select the target sector for an erase operation (they are unused
> otherwise). SSN1 can be programmed only when LOCK1 is cleared to 0.
> 0x00: Sector 0 of user flash bank 1 selected
> 0x01: Sector 1 of user flash bank 1 selected
> ..

rm0433
> Bits 10:8 SNB1: Bank 1 sector erase selection number
> These bits are used to select the target sector for a sector erase operation (they are unused
> otherwise). SNB1 can be programmed only when LOCK1 is cleared to 0.
> 000: sector 0 of bank 1
> 001: sector 1 of bank 1
> ...

usbalbin: Let me know if you want me to revert this too :)

burrbull: Not sure. I'm fine with both variants.

burrbull:
@usbalbin Is it a draft yet?

burrbull (approved):

PR #1231, opened 2025-05-13
Add STM32WBA

burrbull:


Makefile.toml
```diff
@@ -9,7 +9,11 @@ PATCHED_SVDS = set ""
 FORMATTED_SVDS = set ""
 RUST_DIRS = set ""
 for crate in ${crates}
-    crate_yamls = glob_array devices/${crate}*.yaml
+    if eq ${crate} "stm32wb"
+        crate_yamls = glob_array devices/stm32wb[1-5]*.yaml
+    else
+        crate_yamls = glob_array devices/${crate}*.yaml
+    end
```
burrbull, line 16: Fixed for `cargo make`. But I don't know how to fix for `make`.
https://github.com/stm32-rs/stm32-rs/blob/f11dc3deff80dc9d701f51c71667d4d69ab5f345/Makefile#L16-L17

cc @adamgreig

burrbull:
cc @usbalbin

usbalbin (approved):
From a quick look this looks good. I have also looked at the output of `make html` for the ADC4 of the WBA50 and compared it to the RM and it seems good. However there are a lot of things I have not looked too close at

Issue #1227, opened 2025-05-08
STM32L0x2 RCC CIER is read-only


vasilNnikolov:
The RCC_CIER register is read-only in the `stm32l0x{0,1,2,3}.svd` files, checked in svd archive 1.4 from ST. I'll try to fix it tomorrow.

PR #1237, opened 2025-05-21
fix issue #1227

vasilNnikolov:
fixes https://github.com/stm32-rs/stm32-rs/issues/1227


burrbull:
Update changelog, please.

And try to avoid of "merge commits". Use `rebase` for PRs instead.

PR #1244, opened 2025-06-02
Fix RCC SMENR reset values, fix interrupts

god-is-a-crab:
My program wasn't running correctly because the `RCC AHB1SMENR` register doesn't have the correct reset value. I dumped the reset values for all the SMENR registers and edited the patch for them.

I also fixed some of the other interrupts I previously noticed were wrong:
- RCC (missing interrupt)
- LPTIM2 (has LPTIM1 instead of LPTIM2)
- LPUART1 (has LPTIM2 instead of LPUART1)
- SPI1 (missing interrupt)
- SPI2 (has SPI1 instead of SPI2


burrbull:
Squash commits, please.

god-is-a-crab:
@burrbull Hi, can you review again please

PR #1246, opened 2025-06-09
STM32WB0

burrbull:


burrbull:
cc @usbalbin

devices/stm32f768.yaml
```diff
@@ -0,0 +1,488 @@
+_svd: ../svd/stm32f768.svd
```
usbalbin, line 1: Should this file be in this PR?

burrbull: No, it should not. Removed.

usbalbin (approved):
LGTM

PR #1259, opened 2025-07-10
fix groupNames

burrbull:


devices/stm32n645.yaml
```diff
@@ -2,6 +2,16 @@ _svd: ../svd/stm32n645.svd
 
 _clear_fields: "*"
 
+_modify:
+  ADC12:
+    groupName: ADC
```
usbalbin, line 7: Out of curiosity, What does that do?

burrbull: It is not used for code generation. Although `svd2rust` has an option to generate cargo features based on groupNames.
But it useful for analyze.
As for me I use this tag in https://github.com/burrbull/svd-yaml-compare/ tool which helps me to find differences across MC families.

burrbull: for example all SPIx should have same groupName so we can move them in one module/crate, under feature gate.
Or we have `HRTIM_Common`, `HRTIM_Master`, `HRTIM_TIMx`, but all them are part of `HRTIM` peripheral. So it makes sense to somehow collect them together.

usbalbin (approved):

PR #1279, opened 2025-12-10
[STM32WB09] Add Missing Deepstop SRAM Retention Fields

physgun:
These fields are missing in the SVD, but present in RM0505. I know the RM is correct since the WB09 has the same total RAM as the WB07, which has the missing fields in its SVD and in stm32-rs-nightlies.

I've ignored a few fields that are present in the SVD but not in the RMs - since I don't know which one can be trusted without testing on hardware.


burrbull:
Is this bug related to other chips? Could you check it?

burrbull (approved):

PR #1282, opened 2025-12-17
Patches for stm32c0xx SVD files

Ramla-I:
Updated reset values for registers in the RCC and DBG peripherals, and bit widths in the FLASH peripheral. The new values are taken from the RM0490 reference manual.

### DIFF
--- a/devices/patches/flash/c0_bit_width.yaml
+++ b/devices/patches/flash/c0_bit_width.yaml
@@ -1,0 +1,32 @@
+WRP1AR:
+  _modify:
+    WRP1A_STRT:
+      bitWidth: 7
+    WRP1A_END:
+      bitWidth: 7
+WRP1BR:
+  _modify:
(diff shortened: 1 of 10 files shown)


devices/patches/flash/c0_bit_width.yaml
```diff
@@ -0,0 +1,32 @@
...
+    WRP1A_STRT:
+      bitWidth: 7
+    WRP1A_END:
+      bitWidth: 7
+WRP1BR:
+  _modify:
+    WRP1B_STRT:
+      bitWidth: 7
+    WRP1B_END:
+      bitWidth: 7
```
burrbull:
```suggestion
WRP1[AB]R:
  _modify:
    WRP1[AB]_STRT:
      bitWidth: 7
    WRP1[AB]_END:
      bitWidth: 7
```

burrbull (approved):

PR #1287, opened 2026-04-12
Add TIM12 peripheral for STM32H523

astapleton:
The TIM12 peripheral is entirely missing from the STM32H523 SVD, but is included on the chip per the datasheet. I've generated the TIM12 entry from the STM32H533 SVD using svdtools, and included the patch in stm32h523.yaml. I've confirmed that the generated rust code is the same as that for the STM32H533 crate.

### DIFF
--- a/devices/stm32h523.yaml
+++ b/devices/stm32h523.yaml
@@ -1,5 +1,10 @@
 _svd: ../svd/stm32h523.svd
 
+# TIM12 is omitted from stm32h523.svd
+_copy:
+    TIM12:
+        from: ../svd/stm32h533.svd:TIM12
+
 _clear_fields: "*"
 
 _derive:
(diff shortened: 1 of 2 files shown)


burrbull:
If it is just a copy of 533 you could use _copy rule instead

astapleton:
> If it is just a copy of 533 you could use _copy rule instead

Ah, I didn't see that. That would have been much easier. I'll do that.

devices/stm32h523.yaml
```diff
@@ -315,6 +315,21 @@ TIM6:
   _include:
     - fields/tim/v3/tim6.yaml
 
+# TIM12 is omitted from stm32h523.svd
+_copy:
+    TIM12:
+        from: ../svd/stm32h533.svd:TIM12
```
burrbull: Move this to start of file, please.
And squash commits.

burrbull (approved):

